PR #7, opened 2018-12-01
Support for BlackBox updates for Treadle

chick:
> Biggest issue on this push is that the examples are pulled from licensed files in the rocket-chip repo.  Since rocket-chip is in the same organization as this repo is this a problem?

- clean-up imports in several files to reflect actual use
- introduce Testers2OptionsManager to stop extend/with madness
- add AsyncResetReg.v verilog black box for fidelity to rocket usage
- add Async reset test case based on Rocket example
- update build.sbt to track current version of treadle


src/test/resources/vsrc/AsyncResetReg.v
```diff
@@ -0,0 +1,79 @@
+// See LICENSE.SiFive for license details.
```
ducky64: This license file doesn't actually exist.

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
...
+import org.scalatest._
+import treadle.executable.{PositiveEdge, Transition}
+import treadle.{ScalaBlackBox, ScalaBlackBoxFactory}
+
+object AsyncResetReg {
+  val bitWidth: Int = 3
+}
+
+class AsyncResetReg(resetValue: Int = 0)
+        extends BlackBox(Map("RESET_VALUE" -> IntParam(resetValue))) with HasBlackBoxResource {
```
ducky64: We should probably be using ExtModule instead, which has more consistent semantics.

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
...
+      case "d"     => nextValue = value
+      case "en"    => enable = value > BigInt(0)
+      case "rst"   =>
+        if(value > BigInt(0)) {
+          nextValue = resetValue
+          currentValue = nextValue
+        }
+      case _ =>
+    }
+    println(s"next $nextValue cur $currentValue, en $enable")
```
ducky64: Is this println meant to be visible in production?

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
+// See LICENSE for license details.
```
ducky64: How much SiFive code does this incorporate? Does this need to be LICENSE.SiFive?

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
...
+import chisel3.core.IntParam
+import chisel3.tester._
+import chisel3.util.{Cat, HasBlackBoxResource}
+import firrtl.ir.{Param, Type}
+import org.scalatest._
+import treadle.executable.{PositiveEdge, Transition}
+import treadle.{ScalaBlackBox, ScalaBlackBoxFactory}
+
+object AsyncResetReg {
+  val bitWidth: Int = 3
```
ducky64: wat? Would it ever not be one?

src/main/scala/chisel3/tester/Testers2OptionsManager.scala
```diff
@@ -0,0 +1,17 @@
+// See LICENSE for license details.
```
ducky64: Is this used enough to justify creating an OptionsManager class? Is it sufficiently general? For example, what happens if we add Verilator or VCS support?

My current preference is to provide options as a "escape hatch" only, and zero-configuration (or maybe parsing of parameters passed to ScalaTest in infrastructure) should handle most cases. I know that Treadle needs a registry of BlackBoxFactories, but I think creating a `new ExecutionOptionsManager with HasTreadleSuite with ...` at the use site is fine for now.

Also from an API standpoint I don't want to give the impression that spamming OptionsManager is encouraged.

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -162,9 +160,8 @@ class TreadleBackend[T <: MultiIOModule](dut: T,
 }
 
 object TreadleExecutive {
-  import chisel3.internal.firrtl.Circuit
   import chisel3.experimental.BaseModule
-
+  import chisel3.internal.firrtl.Circuit
```
ducky64: wat? Where does get used?

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
...
+class AsyncResetRegTest extends FreeSpec with ChiselScalatestTester {
+
+  "it should work normally with respect to clock" in {
+
+    val manager = new Testers2OptionsManager {
+      treadleOptions = treadleOptions.copy(
+        blackBoxFactories = Seq(new AsyncResetBlackBoxFactory)
+      )
+    }
+    test(new UsesAsyncResetReg, manager) { c =>
```
ducky64: Overall this seems like a way heavyweight, probably overkill, test. Can you not just instantiate one AsyncResetReg, which is defined as a single-bit register, and directly inline (into `test(...)`) the DUT shim Module? The addition of AsyncResetVec seems unnecessary if the goal is just to determine whether async reset works. We'd hope that whatever underlies the Vec construction would be tested in base Chisel regressions.

I'd also like to get away from the spaghetti-with-meatballs code style that defined the older testers (eg inline Modules and Bundles where there's no re-use - there's no point in making the code reader jump around unnecessarily).

Potentially also should be a TODO somewhere to replace the BlackBox implementation with proper Chisel async reset support when that happens.

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+      // testers default reset will set initial value to 3
+      c.io.in.poke(7.U)
+      c.io.out.expect(3.U)
+      c.clock.step()
+      c.io.out.expect(7.U)
+
+      // reset immediately is visible with after reset is poked again
+      c.reset.poke(true.B)
+      c.io.out.expect(3.U)
```
ducky64: I'd also test these conditions:
- Register doesn't update when reset is asserted and its clock steps
- Register updates on clock after reset is deasserted (yes, this is technically handled by testers2 infrastructure, but still...)
- Register maintains reset value after reset deassert but before clock step
- With the single-bit AsyncResetReg, maybe instantiate two of them, one resetting to 0 and the other to 1

ducky64:
Also, does this just work with current Treadle or is it dependent on a particular branch or open PR?

src/test/resources/vsrc/AsyncResetReg.v
chick: Reference removed

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: Uses ExtModule now

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: printf removed

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: Re-wrote with ExtModule based on SiFive, but refactor names and interface, so hopefully very little direct SiFive stuff now

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: Simple way to define constants for tests that might change in future. Removed for now

src/main/scala/chisel3/tester/Testers2OptionsManager.scala
chick: It is definitely worth creating the class, just to be DRY and also to manage the rat's nest of traits involved. I don't think this should be a sticking point as this will all be re-done more rationally
with the Options As Annotations project.

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: dangling import, removed

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: Broke test into two, one for single bit, one for multi-bit, makes test simpler and clearer.
Multi-bit code uses SingleBit components to keep things DRY

src/test/scala/chisel3/tests/AsyncResetRegTest.scala
chick: I think I got all these please confirm

chick:
@ducky64 Ready for re-review

src/test/scala/chisel3/tests/AsyncResetSingleBitRegTest.scala
```diff
@@ -0,0 +1,190 @@
...
+  val enable: Bool = Input(Bool())
+}
+
+/**
+  * This is a black box example that only works with treadle as it does not
+  * define the necessary verilog for verilator/VCS
+  * @param resetValue reset value for this 1 bit register
+  */
+class AsyncResetReg(resetValue: Int = 0) extends ExtModule(Map("RESET_VALUE" -> IntParam(resetValue))) {
+  val io: SingleBitRegIO = IO(new SingleBitRegIO)
```
ducky64: Unless you're using SingleBitRegIO elsewhere (which I don't see), perhaps just inline the bundle def?

src/test/scala/chisel3/tests/AsyncResetSingleBitRegTest.scala
```diff
@@ -0,0 +1,190 @@
...
+  "register is one, after tester startup's default reset" in {
+    test(new UsesSingleBitAsyncResetReg(1), manager) { dut =>
+      dut.io.out.expect(1.U)
+    }
+  }
+  "reset a register works after register has been altered" in {
+    test(new UsesSingleBitAsyncResetReg(1), manager) { dut =>
+
+      // The register starts at zero after default reset and step in tester startup
+      dut.clock.step()
```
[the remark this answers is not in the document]
chick: I think the comment is wrong and I've just changed the test and gotten rid of the step

src/test/scala/chisel3/tests/AsyncResetSingleBitRegTest.scala
chick: I agree it's equivalent but what's to be gained by removing it now. And if structural types go away, all the code may go this way anyway

chick:
The `  import chisel3.internal.firrtl.Circuit` is necessary for the `getTopModule` right below it. Maybe there's an alias for it, but I don't see it. Furthermore it was in the code originally, I am not sure why it appears in the diff

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: This just got moved one line down, it is not new, I'll move it back but you must have put it there originally

src/main/scala/chisel3/tester/Testers2OptionsManager.scala
chick: - I made this class after spending way to much time trying to the `with HasXXXOptions` right in order to get the options manager passed down to treadle so I could see the options. 
- I don't want to go through gymnastics of the raw API here. It's not worth the time
- It's not DRY to write all those `with`s
- OptionsAsAnnotations will explicitly remove the current options managers so this will definitely go away

ducky64:
Pushed changes discussed earlier today:
- Removed Testers2OptionManager in favor of fixing options propagation properly. It will now match on the relevant subtypes of ExecutionOptionsManager and update the one passed into the the tester.
- Removed the multiple bit test.

Does this look fine? @chick

chick:
Yes, I think that it looks fine.

PR #9, opened 2018-12-27
Preserve annotations from initial FIRRTL pass

nhynes:
This PR fixes an issue where new/updated annotations from the [initial compiler pass](https://github.com/ucb-bar/chisel-testers2/blob/master/src/main/scala/chisel3/tester/TreadleBackend.scala#L233) are not passed to the [following lowering pass](https://github.com/freechipsproject/treadle/blob/master/src/main/scala/treadle/executable/ExecutionEngine.scala#L430).

Related: https://github.com/freechipsproject/chisel3/issues/968

### DIFF
--- a/src/main/scala/chisel3/tester/TreadleBackend.scala
+++ b/src/main/scala/chisel3/tester/TreadleBackend.scala
@@ -235,6 +235,9 @@ object TreadleExecutive {
         firrtlExecutionResult match {
           case success: FirrtlExecutionSuccess =>
             val dut = getTopModule(circuit).asInstanceOf[T]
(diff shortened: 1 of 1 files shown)


nhynes:
> this is something that will be fixed by annotations as options?

Yep, I can successfully load memories from file now. I think that the root issue is that the emitted FIR string doesn't actually contain any of the annotations--they're stored separately in the `CircuitState`, so unless _that_ were passed into treadle wholesale, then they have to be added through options.

jackkoenig:
I believe that Annotations as options will fix this because it is generalizing everything (including `CircuitState`) into Annotations which will then be what is passed to everything (including treadle)

Issue #10, opened 2018-12-28
[Proposal] Add optional timeout to (en|de)queue


nhynes:
Setting a maximum number of cycles would prevent infinite loops when testing buggy circuits.

ducky64:
I'm not sure implicit conversions will work, since wouldn't a new object be created at each conversion point and internal state (like elapsed and max cycles) lost?

I think timeout by idle cycles makes more sense than a simple max cycles, since I could see a case for long(ish) tests, but less of a case for long periods where the testdriver doesn't write into the circuit.

I do like the idea of calling setMaxCycles on the clock object itself, this is probably cleaner than adding a slew of flags at testdriver invocation. It might also make sense for the max cycles to be mutable, since there may be segments of test where things are expected to take longer. Though this also brings up, does it make sense for timeout cycles to be scoped?

As for what cause a potentially unbounded test, I could see anything in the general class of waiting for event. For some designs (especially system-level designs?) though, this may be primarily Decoupled-based.

What do you mean by real-time constraints?

nhynes:
> timeout by idle cycles

I like that better actually. I'm not quite seeing anything in treadle that would expose whether the circuit state changed, but I also didn't look too deeply.

> real-time constraints

Like if I have a specification that states that `output.valid` is asserted `x` cycles after `input.valid`. The [Xilinx DMA IP Spec Table 2-2](https://www.xilinx.com/support/documentation/ip_documentation/axi_dma/v7_1/pg021_axi_dma.pdf#G5.425382) is an example.

https://github.com/ucb-bar/chisel-testers2/pull/11 shows an example of how a simple global per-clock limit would work.

ducky64:
I think a timeout would be a different mechanism than LTL constraints (which I think are the basis for some SVA properties). LTL would seem to be the right mechanism for expressing something like "make sure X happens Y time after Z", and generally allows constraints to apply across time. I think @jackkoenig worked on LTL-style assertions in base Chisel with CACL, though I'm not sure about the current status of that project. It might also be possible to implement LTL-style assertions with threads in testers2, though those would need to be written into every single test as opposed to being a circuit property associated with the RTL.

My thoughts on timeouts vs. LTL constraints are that LTL constraints encode correctness properties, whereas timeouts are supposed to prevent the need for the user to kill sbt because the test deadlocks. One might fit some of the needs of the other, but I don't think the use cases map over generally.

[...]

PR #11, opened 2018-12-30
Global per-clock step limit

nhynes:
See the RFC in #10 

This isn't quite as good as a max idle time, but it gets the job done. This PR is mostly a proof-of-concept (i.e. this should be closed once the RFC has been finalized).


ducky64:
Does #13 resolve this use case?

nhynes:
For sure, thanks!

PR #22, opened 2019-02-27
Provide support for a way to use testers2 in jupyter notebook

chick:
- Removes requirement that tests be run within a scalatest context
- New Entry point ChiselTester.apply creates and runs a test
- Create ChiselTesterHelper for shared code existing and new ways to run


src/main/scala/chisel3/tester/ChiselTester.scala
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+/**
+  * This is a simple tester that does not require that it be within the scope of a scalatest
+  * in order to run. This form is suitable for running in the Jupyter notebook.
+  *
+  * @todo When Phases, Stages is implemented add ability to change testing options.
+  */
+object ChiselTester {
```
ducky64: I think I'd prefer it for users to just import a specific package (eg `chisel.tester.main` - but a better name) and write `test(dutGen) { testFn }`. The test would have no name - the name would either be generated from the DUT name, or just 'anonymous'. In the future, we can consider extensions like `test(dutGen).withName(...) { testFn }`, but I don't see a pressing need for that in the immediate case.

This primarily means renaming this object and putting it in a subpackage that isn't visible if you just import testers (since it could conflict with another test environment interface). ChiselTester as a class can stay as-is, since users shouldn't interact with that directly. It might help if it had a more useful name, eg BareTester, that differentiates it from the other test environment interfaces.

src/main/scala/chisel3/tester/ChiselTesterHelper.scala
```diff
@@ -0,0 +1,62 @@
+// See LICENSE for license details.
+
+package chisel3.tester
+
+import org.scalatest.exceptions.TestFailedException
+
+import scala.collection.mutable
+
+trait ChiselTesterHelper { self: TestEnvInterface =>
```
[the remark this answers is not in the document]
chick: Not sure which way this means. ChiselTesterMethod provides concrete implementations of most of the abstract methods in TestEnvInterface.
I found that all the implementations I need for jupyter style was same as your scalatest based.
Shall I just make these concrete ones be the TestEnvInterface default implementation?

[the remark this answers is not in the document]
ducky64: Yeah, basically if they're helper functions that are probably applicable across a range of test environments, I don't think there's a point to breaking them out into a separate class.

src/main/scala/chisel3/tester/ChiselTester.scala
chick: I agree this sounds cleaner, but the hidden consequence is tests overwriting each other possibly picking up bad artifacts or instead making more and more new directories for every test. This maybe ok here, but it should be pondered. In the case of MyBinder the latter is probably fine as the sessions go away. What's you call on the default way of generating a target dir name

ducky64: So I don't expect people to use this in production testing, because if you're running lots of tests you should probably be using ScalaTest or some similar powerful test framework. This is mainly for people who either don't want to set up the ScalaTest infrastructure (so probably pretty small-scale testing) or where ScalaTest isn't available (bootcamp).

I've usually considered temporary directories for testing as generated on a best-effort basis without guarantees. I'd actually prefer no artifacts were generated unless specifically requested by the user (for example, in debug mode), since disk IO takes time and the vast majority of the time won't be used. Possibly unavoidable with Verilator, but possible in treadle.

Under that model, where temporaries are needed for compilation, I think inferring from the DUT name is fine, or even just using the same directory that gets cleared and written for each test.

src/main/scala/chisel3/tester/BackendInterface.scala
```diff
@@ -131,15 +131,64 @@ trait BackendInstance[T <: MultiIOModule] extends BackendInterface {
...
+      val trace = new Throwable
+      val expectStackDepth = trace.getStackTrace.indexWhere(ste =>
+        ste.getClassName == "chisel3.tester.package$testableData" && ste.getMethodName == "expect")
+      require(expectStackDepth != -1,
+        s"Failed to find expect in stack trace:\r\n${trace.getStackTrace.mkString("\r\n")}")
+
+      val trimmedTrace = trace.getStackTrace.drop(expectStackDepth + 2)
+      val detailedTrace = topFileName.map(getExpectDetailedTrace(trimmedTrace.toSeq, _)).getOrElse("")
+
+      batchedFailures += new TestFailedException(
```
ducky64: You should throw a testers2-specific exception (eg ExpectMismatchException) instead of reusing Scalatest infrastructure for a trait that's not supposed to be Scalatest specific

src/main/scala/chisel3/tester/ChiselTester.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  private var count = 0
```
ducky64: What's this for?

src/main/scala/chisel3/tester/ChiselTester.scala
ducky64: Once again, can we rename this to something more descriptive than ChiselTester that specifically calls out the distinguishing characteristics?

src/main/scala/chisel3/tester/BackendInterface.scala
chick: Ok, making it into FailedExpectException

src/main/scala/chisel3/tester/ChiselTester.scala
chick: Intended to strenghten guarantee of unique name.

chick:
@ducky64 Name suggestion? Do you want specific like `SimpleJupyterTester`.
The name is only going to appear in import statement.

src/main/scala/chisel3/tester/ChiselTester.scala
ducky64: I'd prefer this was eliminated entirely. It's shared global state that (though synchronized) may still interact in unpredictable ways. It also makes test names not reproducible (though since you're throwing the system time in there, that's out the window anyways).

If you didn't eliminate this, you should at least make the def private, and possibly put it into another object that encapsulates this. This seems like a purely internal function, not meant to be user-invoked or accessible.

ducky64:
Naming: I'd say something like `RawTester` or `BareTester`. It's not specifically tied to Jupyter (though the bootcamp is going to be one of the main use cases). The specific differentiation is that there's no overarching test infrastructure this connects into.

chick:
Ready for review cycle 3

chick:
I would prefer to conver the tests to check for FailedExpectException which seems more explicit.

ducky64:
Yeah that sounds like a good structural improvement, if you mean what I think you mean.

chick:
I mean change `assertThrows[exceptions.TestFailedException]` to `assertThrows[FailedExpectException]`

ducky64:
I disagree with that. It's fine from an assertThrows standpoint for tests that check for failures, but it's really nice to see that the test itself failed when you run ScalaTest from sbt instead of the kind-of-bloated-and-useless trace it gives you when a test throws an exception.

ducky64:
Yeah, the current implementation doesn't look like great separation of concerns.

Does using the [TestFailedException constructor](http://doc.scalatest.org/1.7/org/scalatest/TestFailedException.html) that has the message, cause, and failedCodeStackDepth arguments propagate the original stack trace correctly in the catch, modify, and re-throw pattern

chick:
Safe to merge

chick:
@ducky64 I'll leave to you any modification that handles conversion from FailedExpectException to TestFailedException

Issue #14, opened 2019-01-29
Support RawModule / multi-clock modules


edwardcwang:
Example of multi-clock circuits: https://github.com/edwardcwang/chisel-multiclock-demo

Ideas
* testers1 semantics (poke-clock and wait-on-clock semantics)
* clock wrapper generator

ducky64:
I have a few ideas, but let's recap the current testers2 architecture:
- Testers2 currently only works on modules with an implicit clock. This clock is driven by the testdriver, with the smallest time step being one cycle of the implicit clock.
- Internal (derived in RTL) clocks are handled by the simulator, so you can do things like digital clock dividers which will work as expected. See [ClockCrossingTest](src/test/scala/chisel3/tests/ClockCrossingTest.scala) and [ClockDividerTest](src/test/scala/chisel3/tests/ClockDividerTest.scala) (which also steps on a derived clock)
- Clocks can also be poked, which will also work as expected since the details are handled by the simulator. See [AsyncClockTest](src/test/scala/chisel3/tests/AsyncClockTest.scala).
- The `clock.step(...)` operation blocks until a rising edge is detected on the target clock, by polling once per time step (currently one cycle of the implicit clock). This works fine for derived clocks (eg digital clock dividers, since the implicit clock is an integer multiple), but not for poked clocks. This is the case in [AsyncClockTest](src/test/scala/chisel3/tests/AsyncClockTest.scala): the DUT clock is manipulated by pokes and is NOT the implicit clock, but the implicit clock is used to pass time (which should be equivalent to a no-op in this particular design).
- - We can (but currently don't) do sanity checks on this by disallowing stepping on Input clocks and Output clocks that are combinationally dependent on any input.

I think the current two ideas all have trade-offs and depends on use cases.

## Clock Generator Wrapper
Turns out this might not be feasible given MultiIOModules, because I don't think it's possible to generate a Module that has the same top-level-IO as another Module (the one it wraps).

The idea would have been to encapsulate the non-implicit-clock-module DUT in an implicit-clock-module wrapper, pass through the non-clock IOs, and generate digital clock dividers for the clock IOs from the wrapper's implicit clock. The derived clocks could be specified simply as a map of the Clock object to the period multiple of the implicit clock.

More thought would be required about optionally specifying phase and default reset behavior.

[...]

PR #31, opened 2019-04-17
Add experimental with[Option] to run, high/low() on clock

### LINKED ISSUE #29: Expose TesterOptions / getTestOptions to programmatically set options
Sometimes it is very useful to be able to set options (notably VCD dumping) programmatically as opposed to from the command line. Right now there isn't an easy way to programmatically set it since `TesterOptions` is private. It can be in an experimental or some other not-fully-supported API.

ducky64:
Resolves #29 

Perhaps unsurprisingly, this also went deep down the refactor rabbit hole.

- Refactors `run(dutGen => T)` to return a TestBuilder, which can be chain applied to run a test.
- Adds an experimental implicit conversion on the ChiselScalatestTester TestBuilder to allow `.withExecOptions`, `.withTesterOptions`. See the AsyncResetRegTest for an usage example.
- Removes the `run(dutGen, execOptions)` methods, so you have to import experimental to specify options.
  - Note: TestBuilder is not provided for RawTester (so there's no way to specify execution options / tester options there). But the primary use for that is probably simple examples, where you probably don't need this power anyways. If this is wrong (as in you have a legit use case), I can add a TestBuilder object to RawTester.
- Expose `TesterOptions` in experimental.
- Add `Clock.high()`, `Clock.low()` in experimental, which does a unchecked clock poke (as in, if you use this with threads, you need to know how Testers2 orders threads). Proper clock poking is under discussion in #14, but this acts as temporary solution for those who need to poke clocks now.

### DIFF
--- a/src/main/scala/chisel3/tester/experimental/package.scala
+++ b/src/main/scala/chisel3/tester/experimental/package.scala
@@ -29,17 +29,11 @@ package object experimental {
     }
   }
 
-//  implicit class ChiselScalatestTesterOptions(x: ChiselScalatestTester) {
-//    // Experimental API, individual options must be specified with named args
-//    def test[T <: MultiIOModule](dutGen: => T, dummy: Int = 0,
-//        execOptions: Option[ExecutionOptionsManager] = None, testOptions: Option[TesterOptions] = None
-//        )(testFn: T => Unit) {
-//      x.runTest(defaults.createDefaultTester(() => dutGen,
-//        testOptions.getOrElse(x.getTestOptions), execOptions))(testFn)
-//    }
-//  }
-
-  implicit class uncheckedPokeableClock(signal: Clock) {
+  /** This provides a quick and dirty way to poke clocks. Inter-thread dependency is NOT checked,
+    * so it is up to you to understand the thread ordering semantics to use this correctly.
+    * Note that thread ordering IS deterministic, so you will NOT get a nondeterministic test.
+    */
+  implicit class UncheckedPokeableClock(signal: Clock) {
     def high(): Unit = {
       if (DataMirror.directionOf(signal) != ActualDirection.Input) {
         throw new UnpokeableException("Cannot only poke inputs")
--- a/src/main/scala/chisel3/tester/experimental/package.scala
+++ b/src/main/scala/chisel3/tester/experimental/package.scala
@@ -3,8 +3,40 @@
 package chisel3.tester
 
 import chisel3._
+import chisel3.experimental.MultiIOModule
+import firrtl.ExecutionOptionsManager
 
+/** Your warranty is now void.
+  *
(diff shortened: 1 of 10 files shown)


edwardcwang:
Generally looks good, looking forwards to trying out the features.

Re. `TestBuilder is not provided for RawTester (so there's no way to specify execution options / tester options there)` -> I may be using RawTester at some point (and having this API/flexibility is a really cool aspect of testers2), so how hard is it to provide this for RawTester as well for completeness' sake?

ducky64:
It's not difficult, though it either involves duplicating code (TestBuilder) into RawTester or more refactoring to make that class separate. If "at some point" isn't soon, I'd prefer to push this as-is and implement the complete feature later, especially since I expect there will be iteration.

edwardcwang:
OK that sounds good to me. I will poke you when I need that feature more urgently then.

src/main/scala/chisel3/tester/experimental/package.scala
```diff
@@ -3,9 +3,49 @@
...
+  * work as expected, or may have unforeseen side effects, or may be powerful yet dangerous.
+  *
+  * You have been warned.
+  */
 package object experimental {
-  implicit class uncheckedPokeableClock(x: Clock) {
+  type TesterOptions = chisel3.tester.internal.TesterOptions
+  val TesterOptions = chisel3.tester.internal.TesterOptions  // expose this internal object, whose "API" is unstable
 
+  implicit class ChiselScalatestOptionBuilder[T <: MultiIOModule](x: ChiselScalatestTester#TestBuilder[T]) {
```
edwardcwang: FYI, this forces us to `import chisel3.tester.experimental._` which may be good to avoid?

ducky64: It's now in its own sub package object, so you would instead either
```scala
import chisel3.tester.experimental.TestOptionBuilder._
```

```scala
import chisel3.tester.experimental.UncheckedClockPoke._
```

I don't think there's a way to do implicit conversions without a wildcard import.

ucbjrl: I've seen cases where the implicits are grouped in an `Implicits` object and then wildcard imported:
```scala
object foo {
  object Implicits {
   ...
  }
}

import ...foo.Implicits._
```

ducky64:
@ucbjrl yeah, though the thing is that we don't want people importing all the experimentals. Ideally I'd like to have:
```scala
import chisel3.tester.experimental.TestOptionBuilder
```
which can also be done in a single line multi-import:
```scala
import chisel3.tester.experimental.{blah, blah, TestOptionBuilder}
```

instead of what is currently implemented
```scala
import chisel3.tester.experimental.TestOptionBuilder._
```

PR #33, opened 2019-04-19
Add support for clock peeking

edwardcwang:
This is super useful since clock poking has been implemented


edwardcwang:
Testcase and experimental added

PR #37, opened 2019-05-14
Add BundleLiteralsSpec

chick:
Three simple tests showing syntax and usage
Third tests showing peeking currently expects an exception


src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
```diff
@@ -36,35 +25,44 @@ class BundleLiteralSpec extends FlatSpec with ChiselScalatestTester {
       io.aOut := io.in.a
       io.bOut := io.in.b
     }) { c =>
-      c.io.in.poke(c.io.in.Lit(0.U, 1.U))
+      c.io.in.poke(c.io.in.cloneType.Lit(_.a -> 0.U, _.b -> 1.U))
```
ducky64: `cloneType` is a crappy, confusing, and informally-deprecated API. Please stop using it, or encouraging people to use it, or throwing it in new code. We can't actually mark it deprecated, because some legacy code makes heavy use of it, but new code should not use it.
Use `chiselTypeOf` instead.

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
```diff
@@ -36,35 +25,44 @@ class BundleLiteralSpec extends FlatSpec with ChiselScalatestTester {
...
-        c.out.expect(c.in.Lit(0.U, 2.U))
+        c.in.poke(c.in.cloneType.Lit(_.a -> 0.U, _.b -> 1.U))
+        c.out.expect(c.in.cloneType.Lit(_.a -> 0.U, _.b -> 2.U))
       }
     }
   }
 
-  ignore should "peek Bundle literals" in {
-    // TODO to be written
+  // peek on bundle not supported yet, this test will fail and should
```
ducky64: Why'd you get rid of the ignore? I think the ignore is cleaner and expresses intent better than an exception intercept. The test name is also very confusing.

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
```diff
@@ -36,35 +25,44 @@ class BundleLiteralSpec extends FlatSpec with ChiselScalatestTester {
...
+  // be altered when BundleLiteral peeking works
+  it should "not peek Bundle literals at this point in time" in {
+    intercept[LiteralTypeException] {
+      test(new PassthroughModule(new DoubleElements)) { c =>
+        c.in.poke(c.in.cloneType.Lit(_.a -> 0.U, _.b -> 1.U))
+        val output = c.out.peek()
+        output.a === 0.U should be(true.B)
```
ducky64: Uh, I really don't think that will work, since we don't do const prop in the frontend - it will just end up as a wire, and that's assuming you don't get some error for trying to invoke HCL operations outside a Builder context.

That being said, I'm not sure how we check for literal equality.

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
chick: Agreed, I meant to ask you about this, but forgot about it. Fixed

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
chick: I think getting rid of the ignore is a better way of making sure we update the test when peeking is of Bundled is enabled. I agree the test name is bad and is [changed.](url)

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
chick: @ducky64 Can you think of a better example of a use case for peeking a Bundle?

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
ducky64: I'd prefer expressing intent over some confusing system that results in automated reminders. Because when you're building a new feature, you should already be thinking of writing tests for it, and the ignored should act as a flag that gets cleaned up (though perhaps eventually). And the tests are intended to essentially codify behavior, and a LiteralTypeException is not the end goal here.

I guess ideally there would be an ignore flag that checks for a test failure and suppresses it, but errors on a non-failure. Don't know if that exists in ScalaTest.

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
chick: Test is now an ignore

src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
chick: Test is now an ignore

PR #42, opened 2019-06-04
Lot's of changes to incorporate Stage/Phases API.

chick:
- Simplifies the code a fair amount. All tests pass
- Add new withOptionsModifier for TestOptionBuilder
- deprecated withExecOptions, withTesterOptions
- Fix up test work directory identification and creation
- Add scaladoc for test()
- Get rid of testerOptions from createDefaultTester
- move sanitizeFileName to be re-usable
- Fix RawTester to allow annotations
- Re-work test assembly code in TreadleExecutive
- Deprecate TesterOptions


src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
```diff
@@ -13,6 +18,24 @@ trait TestEnvInterface {
...
+    * if there is not a TargetDirAnnotation already present
+    *
+    * @param defaultDir     a default directory
+    * @param annotationSeq  annotations to add it to, unless one is already there
+    * @return
+    */
+  def addDefaultTargetDir(defaultDir: String, annotationSeq: AnnotationSeq): AnnotationSeq = {
+    if(annotationSeq.exists { case _: TargetDirAnnotation => true ; case _ => false }) {
+      annotationSeq
+    }
```
ducky64: what is this `else` style and why are we using it

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
```diff
@@ -13,6 +18,24 @@ trait TestEnvInterface {
...
+  /**
+    * Will add a TargetDirAnnotation with defaultDir with "test_run_dir" path prefix to the annotations
+    * if there is not a TargetDirAnnotation already present
+    *
+    * @param defaultDir     a default directory
+    * @param annotationSeq  annotations to add it to, unless one is already there
+    * @return
+    */
+  def addDefaultTargetDir(defaultDir: String, annotationSeq: AnnotationSeq): AnnotationSeq = {
+    if(annotationSeq.exists { case _: TargetDirAnnotation => true ; case _ => false }) {
```
ducky64: shouldn't there be a space after the if and the parens?

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
```diff
@@ -13,6 +18,24 @@ trait TestEnvInterface {
...
 
+  /**
+    * Will add a TargetDirAnnotation with defaultDir with "test_run_dir" path prefix to the annotations
+    * if there is not a TargetDirAnnotation already present
+    *
+    * @param defaultDir     a default directory
+    * @param annotationSeq  annotations to add it to, unless one is already there
+    * @return
+    */
+  def addDefaultTargetDir(defaultDir: String, annotationSeq: AnnotationSeq): AnnotationSeq = {
```
ducky64: I'd prefer this to fail noisily. It should be required the annotation isn't in the AnnotationSeq already. This seems to be asking for nonintuitive behavior.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
ducky64: I think this may be a case where isInstanceOf produces cleaner code than the case matching.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
```diff
@@ -13,6 +18,24 @@ trait TestEnvInterface {
...
+    if(annotationSeq.exists { case _: TargetDirAnnotation => true ; case _ => false }) {
+      annotationSeq
+    }
+    else {
+      val target = TargetDirAnnotation("test_run_dir" + File.separator + defaultDir)
```
ducky64: consider just using the target dir, instead of prepending test_run_dir, especially since you're putting this in a base class where it may be used in other contexts.

src/main/scala/chisel3/tester/internal/Testers2.scala
```diff
@@ -2,16 +2,26 @@
...
-import scala.util.DynamicVariable
 import chisel3.experimental.MultiIOModule
+import firrtl.AnnotationSeq
+import firrtl.stage.phases.DriverCompatibility.TopNameAnnotation
+import treadle.WriteVcdAnnotation
 
 import scala.util.DynamicVariable
 
 // Internal common set of options to be understood by all backends
+@deprecated("Use annotation based options instead. See: .withAnnotations", "chisel-testers2 20190604")
```
ducky64: May also want to note that the write vcd annotation probably won't work with a theoretical Verilator backend since it seems to be a treadle specific thing.

src/main/scala/chisel3/tester/package.scala
```diff
@@ -164,4 +164,13 @@ package object tester {
   }
 
   implicit def decoupledToDriver[T <: Data](x: DecoupledIO[T]) = new DecoupledDriver(x)
+
+  /**
+    * Simple file name sanitizer
+    * @param name file name to be sanitized
+    * @return
+    */
+  def sanitizeFileName(name: String): String = {
```
ducky64: Should this be part of the testers2 API? I think not, this seems like an internal function.

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
```diff
@@ -2,24 +2,28 @@
...
-  class TestBuilder[T <: MultiIOModule](val dutGen: () => T,
-      val execOptions: Option[ExecutionOptionsManager], val testOptions: Option[TesterOptions]) {
-    protected def getTestOptions: TesterOptions = {
-      val test = scalaTestContext.value.get
-      TesterOptions(test.name, test.configMap.contains("writeVcd"))
+  class TestBuilder[T <: MultiIOModule](val dutGen: () => T, annotationSeq: AnnotationSeq) {
+    def getTestName: String = {
+      scalaTestContext.value match {
+        case Some(x) =>
+          sanitizeFileName(x.name)
```
ducky64: what prompted this?

src/test/scala/chisel3/tests/OptionsPassingTest.scala
```diff
@@ -0,0 +1,66 @@
...
+import chisel3.tester._
+import org.scalatest._
+import treadle.{VerboseAnnotation, WriteVcdAnnotation}
+import chisel3.tester.experimental.TestOptionBuilder._
+
+class OptionsPassingTest extends FlatSpec with ChiselScalatestTester with Matchers {
+  behavior of "Testers2"
+
+  it should "write vcd output when passing in a WriteVcdAnnotation" in {
+    // TODO: this actually relies on total thread ordering
```
ducky64: Um, this seems besides the point of the test. Why are we testing shifter behavior in a test that should be about dumping VCDs? And you don't even care about the particular stimulus, you're only looking at a particular file being generated? And does this clean up the relevant file each time, or are regressions going to be flakey?

src/test/scala/chisel3/tests/OptionsPassingTest.scala
```diff
@@ -0,0 +1,66 @@
...
+    val outputStream = new ByteArrayOutputStream()
+    Console.withOut(new PrintStream(outputStream)) {
+      test(new ShifterModule(UInt(8.W), 4)).withAnnotations(Seq(VerboseAnnotation)) { c =>
+        c.clock.step(1)
+      }
+    }
+
+    val output = outputStream.toString
+
+    output.contains("Symbol table:") should be (true)
```
ducky64: Do we have anything we're particularly looking for in verbose mode or is this just going to be a [change detector](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) test?
Especially `clock/prev`, is there any reason for that particular signal and formatting of the signal name? Or is this going to be brittle to however verbose mode is interpreted by Treadle?

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -237,68 +245,48 @@ object TreadleExecutive {
...
-      optionsManager.treadleOptions = optionsManager.treadleOptions.copy(writeVCD = true)
-    }
+      val chiseledAnnotations = (new ChiselStage).run(
+        annotationSeq ++ Seq(generatorAnnotation, NoRunFirrtlCompilerAnnotation)
+//        annotationSeq ++ Seq(generatorAnnotation)
+      )
 
-    // Force a cleanup: long SBT runs tend to fail with memory issues
-    System.gc()
+      val circuit = generatorAnnotation.elaborate.circuit
```
ducky64: Is this dependent on the post-ChiselStage annotation? In other words, is this annotation mutable?

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -237,68 +245,48 @@ object TreadleExecutive {
...
+      val dut = getTopModule(circuit).asInstanceOf[T]
+
+      val portNames = DataMirror.modulePorts(dut).flatMap { case (name, data) =>
+        getDataNames(name, data).toList
+      }.toMap
+
+      val treadledAnnotations = TreadleTesterPhase.transform(chiseledAnnotations)
+
+      val treadleTester = treadledAnnotations.collectFirst { case TreadleTesterAnnotation(t) => t }.getOrElse(
+        throw new Exception(s"TreadleTesterPhase did not build a treadle tester")
```
ducky64: Any reason this handles the get failure with a custom Exception while the next one (circuitState) just does a get?

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
```diff
@@ -63,7 +67,33 @@ trait ChiselScalatestTester extends Assertions with TestSuiteMixin with TestEnvI
     }
   }
 
+  /**
+    * Constructs a unit test harness for the Chisel Module generated by dutGen.
+    * General use looks like
+    * {{{
+    *   test(() => new PlusOne) { c =>
```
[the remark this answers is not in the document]
chick: Fixing signatures, I don't think it's worth mentioning the `new Module { ??? }` syntax

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
```diff
@@ -63,7 +67,33 @@ trait ChiselScalatestTester extends Assertions with TestSuiteMixin with TestEnvI
...
+    * If you need to add options to this unit test you can tack on a .withAnnotations modifier
+    * You must add `import chisel3.tester.experimental.TestOptionBuilder._` to use .withAnnotations
+    *
+    * For example:
+    * {{{
+    *   test(() => new TestModule).withAnnotations(Seq(WriteVcdAnnotation) { c =>
+    *     // body of the unit test
+    *   }
+    * }}}
+    * @see src/test/scala/chisel3.tests/OptionsBackwardCompatibilityTest for examples
```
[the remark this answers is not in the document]
chick: Yep, I was looking at package view in IntelliJ when I added the comment

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
```diff
@@ -2,24 +2,28 @@
...
-      val execOptions: Option[ExecutionOptionsManager], val testOptions: Option[TesterOptions]) {
-    protected def getTestOptions: TesterOptions = {
-      val test = scalaTestContext.value.get
-      TesterOptions(test.name, test.configMap.contains("writeVcd"))
+  class TestBuilder[T <: MultiIOModule](val dutGen: () => T, annotationSeq: AnnotationSeq) {
+    def getTestName: String = {
+      scalaTestContext.value match {
+        case Some(x) =>
+          sanitizeFileName(x.name)
+        case _ => "a"
```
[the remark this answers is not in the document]
chick: will use `.get`. I tend to think of a naked `.get` as a bit sketchy, the exception thrown by that is unlikely to be of any use to an end user, but it is probably not that likely to happen here.

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
```diff
@@ -2,24 +2,28 @@
...
+      }
     }
 
     def apply(testFn: T => Unit): Unit = {
-      runTest(defaults.createDefaultTester(dutGen, testOptions.getOrElse(getTestOptions), execOptions))(testFn)
+      val newAnnos = addDefaultTargetDir(getTestName, annotationSeq)
+
```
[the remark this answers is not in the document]
chick: Removed, I personally find two dense lines like that less readable but I don't have strong feelings about it.

src/test/scala/chisel3/tests/OptionsPassingTest.scala
chick: Um, I needed something to run that would produce a VCD a more advanced test might check that the stimulus is reflected in the vcd, but I wasn't interested in that, I mostly wanted to see that the file was created. The vcd file is 749 bytes and the test will re-use this directory, so I don't think this will be a regression problem. But I will delete to avoid confusion with successive runs.

src/test/scala/chisel3/tests/OptionsPassingTest.scala
chick: There is not a hard contract for what verbose mode is so it would be hard to make this an implementation test. Currently there is only treadle in the backend, when we introduce another backend, we will have to decide if there will be tests that depend on the backend selected.
I chose those two strings as unlikely to be found by accident. The `VerboseAnnotation` is treadle specific but it also something that I personally need to make sure I can pass into treadle when debugging. This test was made to confirm that process. I'd be happy to discuss how to make this all more rigorous

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: This annotation is not mutable.

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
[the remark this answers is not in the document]
seldridge: This use of `"a"` may have come from me originally (see: `AddImplicitOutputFile` phase that adds an output file annotation derived from a circuit, falling back to `"a"` if no such circuit exists). This was a legacy carry-over of the idea of "viewing" an arbitrary, possibly garbage `AnnotationSeq` as some type where you had to provide a default for a non-optional value. But, the "view" idea predates Stage/Phase. Once you move to Stage/Phase, then you have an explicit ordering of Phases (enforced/canonicalized by the dependency API) and can rely on guarantees on what is in the annotation sequence.

So, if you can guarantee that this will always succeed, a naked `.get` should be fine. If you can't guarantee that, then it might hint that there's a phase-based organization that could untangle things.

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: The error handling could be improved. In treadle there are a number of ways that it can fail to produce a tester, but if it does produce a tester there should not be a case where a circuit state was not generated. I have made the message more explicit in that it will now show the annotations involved. This will help me, at least, debug this mostly unlikely case.

ducky64: Unless it's common enough that you think a user might encounter it frequently, I'm fine with it dumping a stack trace, which will have enough information for someone to open an issue.

src/main/scala/chisel3/tester/TreadleBackend.scala
ducky64: Can you move this above the ChiselStage invocation so that it's clear that it doesn't depend on ChiselStage running?

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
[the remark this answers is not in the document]
ducky64: Unless there's a known or expected failure mode, I'd prefer things fail noisily so it won't result in unexpected (and therefore difficult-to-debug) behavior.

We might also want to have a better discussion of robust error handling, but my current philosophy is that the stack trace of a .get() is going to provide enough information for a developer to debug it. And users that encounter a random stack trace probably are just going to open an issue and post the stack trace, I don't expect them to be able to do anything with a random failure in the middle of the compilation process.

src/main/scala/chisel3/tester/TreadleBackend.scala
ducky64: Also, would it make sense to just have the TreadleTesterPhase throw an exception? If there's a fatal error in there, it seems like it would make sense to terminate there and provide a stack trace from inside treadle?

src/main/scala/chisel3/tester/ChiselScalatestTester.scala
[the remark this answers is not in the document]
seldridge: Yeah, I think a hard error here is fine. This error is entirely unexpected and a stack trace is better than the alternative.

Towards error handling (not a solution to accumulate errors), we could just use what @jackkoenig has done with [FIRRTL error handling](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/stage/FirrtlStage.scala#L27) everywhere to push users to file issues on unexpected exceptions.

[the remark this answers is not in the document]
chick: Made it into a naked `.get`

src/main/scala/chisel3/tester/RawTester.scala
```diff
@@ -24,38 +21,40 @@ private class RawTester(testName: String) extends Assertions with TestEnvInterfa
...
-  *
-  * @todo When Phases, Stages is implemented add ability to change testing options.
   */
 object RawTester {
 
   /**
     * Run one test
+    * General use looks like
+    * {{{
+    *   test(() => new PlusOne) { c =>
```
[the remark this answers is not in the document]
chick: Done

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -17,9 +17,12 @@ class TreadleBackend[T <: MultiIOModule](dut: T,
     val dataNames: Map[Data, String], val combinationalPaths: Map[Data, Set[Data]],
     tester: TreadleTester)
     extends BackendInstance[T] with ThreadedBackend {
+
+  type CycleTracker = mutable.Map[Clock, Int]
```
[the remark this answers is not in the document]
chick: returned to original form. I had changed this because it was the only obstacle remaining for this file to be style check clean. I try to keep my files that way as it helps IntelliJ catch stuff for me, but this is your code. I had introduced the type declaration trying to minimize the boilerplate to get to clean, I wasn't entirely satisfied by it.

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -225,7 +228,12 @@ object TreadleExecutive {
   })
 
   // TODO: better name
-  def combinationalPathsToData(dut: BaseModule, paths: Seq[CombinationalPath], dataNames: Map[Data, String]) = {
+  def combinationalPathsToData(
+    dut: BaseModule,
```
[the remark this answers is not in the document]
chick: returned to original form. This too was a style violation of line length, perhaps we should up what we consider to be the reasonable max

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -225,7 +228,12 @@ object TreadleExecutive {
...
+    paths: Seq[CombinationalPath],
+    dataNames: Map[Data, String]
+  ): Map[Data, Set[Data]] = {
+
```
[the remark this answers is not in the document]
chick: removed, when using multi-line method declaration I prefer a blank line for readability in my code

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -237,68 +245,48 @@ object TreadleExecutive {
...
+      val generatorAnnotation = chisel3.stage.ChiselGeneratorAnnotation(dutGen)
 
-    // Tester options take priority over exec options
-    val testName = testOptions.name.replaceAll(" ", "_").replaceAll("\\W+", "")  // sanitize filename
-    optionsManager.commonOptions = optionsManager.commonOptions.copy(
-        targetDirName = s"test_run_dir/$testName")
-    if (testOptions.writeVcd) {
-      optionsManager.treadleOptions = optionsManager.treadleOptions.copy(writeVCD = true)
-    }
+      val chiseledAnnotations = (new ChiselStage).run(
```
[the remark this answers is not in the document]
chick: Yes, these annotations have been created by the chisel stage

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -237,68 +245,48 @@ object TreadleExecutive {
...
+        annotationSeq ++ Seq(generatorAnnotation, NoRunFirrtlCompilerAnnotation)
+//        annotationSeq ++ Seq(generatorAnnotation)
```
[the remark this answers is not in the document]
chick: gone

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: moved

src/main/scala/chisel3/tester/TreadleBackend.scala
```diff
@@ -237,68 +245,48 @@ object TreadleExecutive {
...
-        }
-      case _ =>
-        throw new Exception("Problem with compilation")
+      val dut = getTopModule(circuit).asInstanceOf[T]
+
+      val portNames = DataMirror.modulePorts(dut).flatMap { case (name, data) =>
+        getDataNames(name, data).toList
+      }.toMap
+
+      val treadledAnnotations = TreadleTesterPhase.transform(chiseledAnnotations)
```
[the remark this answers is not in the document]
chick: Yes it is a verb, treadled is past tense.

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: I have changed this to show the annotations that were sent in. I am not entirely sure what roles treadle phase will take on. So I am hesitant to make it an exception internally. I think this will make debugging easiest.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
chick: I disagree, for some tests the user may want to specify a specific working directory. The way to do that (at least for now) is to add their own `TargetDirAnnotation`

src/main/scala/chisel3/tester/TreadleBackend.scala
seldridge: My understanding of this is that if you give treadle an empty annotation seq, it will do nothing to it. Hence, this PR needs to check if an expected output annotation is present. This contrasts with FIRRTL, where there is an explicit `Checks` phase that validates the input annotation seq and will, e.g., throw an exception internally if the input annotation seq doesn't make sense. You may consider that pattern for treadle with a checks phase.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
chick: Added, I have always used the no space convention. Doesn't trigger a warning from IntelliJ style checker. Convention seems mixed
```
CHICKMBP2:chick brew pr-masters > grep -R 'if(' */src | sort -u | wc
    1768   11106  201011
CHICKMBP2:chick brew pr-masters > grep -R 'if (' */src | sort -u | wc
    2297   17887  253366
```

chick: Uses `isInstanceOf` now

src/main/scala/chisel3/tester/TreadleBackend.scala
ducky64: I think it's also useful to look at it from a user intent point of view: what do you expect the user to expect when TreadleTesterPhase is called? If it's expected to do one thing, and it can't do that one thing, then that should probably be a hard error. If the invocation is super generalized and might do totally different things (including nothing - by design and "contract"), then this strategy probably won't work, but you might also want to rethink the interface there.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
chick: reverted to your vernacular. I have been using my else style in scala since 2008, I have been using the equivalent in C since 1983. It seems to work ok :-)

src/test/scala/chisel3/tests/OptionsPassingTest.scala
ducky64: Yeah, sounds like there isn't a great solution here (because there's no well-defined interface, but we still want this to be tested), so we can leave it as-is until this becomes a maintenance burden.

src/main/scala/chisel3/tester/internal/TestEnvInterface.scala
chick: Not sure what you are suggesting here. Are you saying working directories should end up somewhere other than where they are now, or somehow generate the current ones in a different manner?

ducky64: No, the end result should be the same, it's just a matter of where the test_run_dir gets tacked on. I'd argue that should be at whatever invokes this function, since the test environment might want to define something completely different, but still reuse this annotation creator.

src/main/scala/chisel3/tester/internal/Testers2.scala
chick: Added this as //TODO

src/main/scala/chisel3/tester/package.scala
chick: Good question. I have added `//TODO` and API `@note` to build a safe useful api for user access to the directory being used if that is deemed merited. But I'd rather not solve that right now.

chick:
@ducky64 Just pushed an update that brings github code up to the current conversation and review questions state

src/main/scala/chisel3/tester/TreadleBackend.scala
chick: I'd like to leave it the way it is for the moment. TreadlePhase is used by the TreadleRepl which can be spun up without specifying firrtl source. Old testers needs to be converted as well and it has it has its own pattern, perhaps that will help drive where to put the error and what it should be.

src/main/scala/chisel3/tester/TreadleBackend.scala
ducky64: Maybe I'm just using GitHub wrong but I don't see the change. Did you push your code?

src/main/scala/chisel3/tester/package.scala
ducky64: Can you move this into either internal or experimental? I strongly think this doesn't belong in the tester package.

chick:
moved sanitizeFileName to experimental
use anonymous Module in write vcd test

src/test/scala/chisel3/tests/OptionsPassingTest.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    val vcdFileName = "test_run_dir" +
+            File.separator +
+            sanitizeFileName(s"Testers2 should write vcd output when passing in a WriteVcdAnnotation") +
+            File.separator +
+            sanitizeFileName("OptionsPassingTestanon2") +
+            ".vcd"
+
+    val vcdFile = new File(vcdFileName)
+    println(s"vcd file name $vcdFileName ${vcdFile.getAbsolutePath}")
```
ducky64: automated tests shouldn't have `print`s

src/test/scala/chisel3/tests/OptionsPassingTest.scala
```diff
@@ -0,0 +1,56 @@
...
+    vcdFile.exists() should be (true)
+    vcdFile.delete()
+  }
+
+  it should "allow turning on verbose mode" in {
+    val outputStream = new ByteArrayOutputStream()
+    Console.withOut(new PrintStream(outputStream)) {
+      test(new ShifterModule(UInt(8.W), 4)).withAnnotations(Seq(VerboseAnnotation)) { c =>
```
ducky64: nit: ShifterModule is also besides the point

PR #44, opened 2019-06-27
Fix 2.11 test failure

chick:
2.11 scalatest api returns different test context name
than 2.12, this fix checks to make sure a vcd file was created by using less restrictive name checking.


chick:
@ducky64 @ucbjrl Vcd file is deleted after its presence is detected.

PR #48, opened 2019-07-17
Verilator pass 2

chick:
Second round of building support for verilator
- Add c++ headers and related as resources for building verilator simulation
- add experimental.backends.verilator for verilator backend
- A few other glue bits for new backend
- A separate test package for verilator backend for now
>Note: verilator does not work correctly with the clock.high and clock.low methods. The
tests above fail on those usages with an exception. Working to fix this in next pass


src/main/scala/chisel3/tester/internal/Utils.scala
```diff
@@ -0,0 +1,29 @@
...
+object Utils {
+
+  /** Converts an unsigned BigInt of width X to it's signed value
+    * Basically if msb is set convert it to a negative number
+    *
+    * @param unsigned an unsigned value that should be converted to signed
+    * @param width    the width of the target (this helps to determine if sign (MSB) is set
+    * @return         a signed BigInt
+    */
+  def unsignedBigIntToSigned(unsigned: BigInt, width: Int): BigInt = {
```
ducky64: I'm not sure if this is useful as a general util as opposed to a specific communication protocol detail for just the Verilator backend.

src/main/scala/chisel3/tester/internal/Testers2.scala
```diff
@@ -25,6 +29,47 @@ case class TesterOptions(
   }
 }
 
+trait TestOption extends Unserializable { this: Annotation => }
+trait TestOptionObject extends NoTargetAnnotation with HasShellOptions with TestOption
+
+case object WriteVcdAnnotation extends TestOptionObject {
```
ducky64: Does this get used anywhere - I see an import treadle.WriteVcdAnnotation remaining elsewhere in the code? Also, why isn't this in Treadle?

.scalafmt.config
```diff
@@ -0,0 +1,15 @@
+style = defaultWithAlign
```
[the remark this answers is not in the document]
chick: removed.

src/main/resources/testers2/verilator/sim_api.h
```diff
@@ -0,0 +1,353 @@
+#ifndef __SIM_API_H
```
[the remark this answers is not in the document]
chick: moved into experimental directory in parallel with code structure

src/test/scala/chisel3/experimental/tests/BasicTest.scala
```diff
@@ -0,0 +1,118 @@
+package chisel3.experimental.tests
+
+import chisel3._
+import chisel3.tester._
+import chisel3.tests.{PassthroughModule, StaticModule}
+import chisel3.tester.experimental.TestOptionBuilder._
+import chisel3.tester.internal.VerilatorBackendAnnotation
+import org.scalatest._
+
+class BasicTest extends FlatSpec with ChiselScalatestTester with Matchers {
```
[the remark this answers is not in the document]
chick: Renamed for the moment to VerilatorBasicTest.
As we discussed after the meeting I would like to change or get rid of any copied tests by creating some mechanism that specifies to run an existing tests with a different backend. I'd like to wait for us to discuss that more before I both fixing this particular cargo cult programming

src/main/scala/chisel3/tester/internal/Utils.scala
chick: Moved into verilator utils

src/main/scala/chisel3/tester/experimental/backends/verilator/CommandEditor.scala
```diff
@@ -0,0 +1,100 @@
+// See LICENSE for license details.
```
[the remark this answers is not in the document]
chick: This was code specifically written for verilator based on user feedback from Steve Burns. It might belong somewhere more generally in the long run, but I don't think it is correct to call it *legacy*

src/main/scala/chisel3/tester/internal/Testers2.scala
chick: I am not entirely happy with this; here is my reasoning for keeping it. Treadle has and uses a WriteVcd annotation but verilator uses a SuppressVcd. I think there will be a few examples like this where similar concepts are implemented differently by different backends and in those cases testers2 may have an annotation for the high-level construct that gets mapped to the backend specific version.  Your thoughts?

src/test/scala/chisel3/experimental/tests/ClockPokeTest.scala
```diff
@@ -0,0 +1,106 @@
...
+import chisel3.tester._
+import chisel3.tester.experimental.TestOptionBuilder._
+import chisel3.tester.experimental.UncheckedClockPoke._
+import chisel3.tester.internal.{TreadleBackendAnnotation, VerilatorBackendAnnotation}
+import chisel3.util._
+import org.scalatest._
+import treadle.executable.ClockInfo
+import treadle.{ClockInfoAnnotation, WriteVcdAnnotation}
+
+class ClockPokeTest extends FlatSpec with ChiselScalatestTester {
```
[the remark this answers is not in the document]
chick: Renamed this as well.

chick:
@ducky64 Ready for re-review

src/main/scala/chisel3/tester/internal/Testers2.scala
ducky64: I think since we have control over the backends and how they interpret annotations, we should be able to standardize it. Perhaps there should be a collection of standard simulator annotations that backends should understand?

I'm fine with legacy backends (like pre-refactor Verilator) being spotty but we should keep the modern backends (treadle, post-refactor Verilator) compliant with whatever standard annotations we set.

seldridge: I'd probably like to think about this, but I expect we could do a common `Stage` here and then subclass off that. I agree that unifying things as much as possible would be great for a future non-legacy version.

chick: I have changed verilator now to behave the same treadle. The default for both is no VCD is produced. Adding a testers2 WriteVcdAnnotation causes VCD output to be turned on for both.
The internal mechanisms are different between the backends and TreadleBackend converts the testers2 WriteVcdAnnotation into a treadle WriteVcdAnnotation. As new backends are added, the <NewBackend> and <NewBackendExecutive> classes will mediate when necessary between any global testers2 general options and their own internal representations. I'm happy to discuss this more but I feel ok about the current implementation until we have more backends

ducky64:
I'd prefer to separate out experimental and legacy - though both have the expectation of "this code may go away in the future" they have different meanings.

Also, (ignoring the above temporarily) why is it `experimental.backends.verilator` instead of `experimental/backends/verilator`? Is this some kind of new recommended style, because I haven't seen it anywhere else before?

chick:
@ducky64
The latest commits seem to fix the clock.high clock.low problem (Thanks to @grebe)
I also fixed up the resource path.
I think this is ready to merge.
I need to do some additional work with logging etc, but I will do that stuff in a follow-up PR

ducky64:
It seems like a lot of comments weren't addressed. Perhaps you forgot to push?
- Would prefer most of code be moved to `legacy` instead of `experimental`.
- Orthogonally to the above, why is there a folder called `experimental.backends.verilator` instead of `experimental/backends/verilator`

Also consider noting somewhere that WriteVcdAnnotation may be moved to FIRRTL in the future so it can be shared across backends.

The changes to make the clock poke work looks fine. I don't quite understand it (consider documenting the nuance between `dut->_eval_settle(dut->__VlSymsp);` and `dut->eval();`) but it's probably fine for legacy code.

What's up with the GCD test? `GcdRegression` doesn't seem like a particularly informative name. Can you reduce this down to a minimalistic test case? If the goal is to test a few variations of GCD, I'd prefer it be done in the [way it was presented during the CCC18 intensive](https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala). I'm not a fan of spamming tests for the sake of spamming tests, I'd prefer tests as examples and tests as documentation where possible - and the style issues make this very questionable.

I'm fine with removing the test and adding the GCD test from CCC18 in a separate PR, unless this is actually testing some edge case and GCD is just incidental. In which case, I think this should be reduced to a minimalistic test case.

chick:
@ducky64 Sorry, a push failed for some reason. It has now succeeded.
- The resource file now have the a more normal looking directory structure
  - The problem was caused by the way that IntelliJ constructed the resource for me.
- The GcdRegression was an accidental inclusion
  - But I am interested in doing some performance regressions against testers
  - particularly to make sure the c++ changes do not overly impact simulation times.
- I have added a note that WriteVcdAnnotation may be moved to firrtl
- I have no idea what files should be legacy vs experimental.
  - All the files brought in for verilator have been moved into the experimental folder
    - previously I had added some outside the experimental scope
  - verilator is not some legacy case, it is a common production usage of testers.
  - experimental should be more than sufficient to warn users that the API might change

ducky64:
side question (no code changes needed), did you have an explanation for the `dut->_eval_settle(dut->__VlSymsp);` and `dut->eval();` change? Might be relevant as we refactor this backend eventually.

chick:
I don't know exactly why the `eval()` works and the `eval_settle` does not. Paul and I were looking through usages of those functions in the code and finding references. It appeared through experimentation that `eval_settle` does not evaluate registers to see if their driving clock has changed, but the ordinary eval does. I have run some regressions and I don't see a huge performance hit with small circuits like GCD

adoerflinger:
I'm playing around a little bit with chisel-testers2 and wanted to test different backends. There seems to be a problem with generating VCDs, in particular for the verilator backend mentioned in this issue:
<pre><code>test(new MyConfig).withAnnotations(Seq(WriteVcdAnnotation))</code></pre> produces a vcd file.
<pre><code>test(new MyConfig).withAnnotations(Seq(VerilatorBackendAnnotation, WriteVcdAnnotation))</code></pre> does not produce a vcd file.
Providing the parameter <pre><code>-DWriteVcd=1</code></pre>  when executing the test (as described in the Readme) does not produce a vcd at all (not for treadle backend and not for verilator backend).
Am I missing something? Happy to hear your response!

ducky64:
The `-DWriteVcd=1` not working is a known bug, it was (accidentally?) dropped in an unrelated PR. Will try to get that fixed.

Will need to do more debugging on why Verilator VCD isn't working,

ducky64:
Ok, looks like there's actually two `WriteVcdAnnotation`: `treadle.WriteVcdAnnotation` and `chiseltest.internal.WriteVcdAnnotation`. The former is treadle-only but the latter also works for Verilator.

adoerflinger:
Thanks a lot for the quick reply, that solved my problem! Verilator VCD is produced when using `chiseltest.internal.WriteVcdAnnotation`.

PR #54, opened 2019-08-14
Fixes getting circuit from ChiselCircuitAnnotation

chick:
Recent changes to ChiselCircuitAnnotation caused
compile failure. This fixes failure in testers2


chick:
No, that was a mistake

PR #55, opened 2019-08-15
This adds a test to show support for new AsyncReset

chick:
AsyncReset is new type in chisel3 and firrtl.
This PR adds test that demonstrates support for new native AsyncReset
Test defaults to Treadle, but can be manually directed to verilator


src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+import chisel3.tester.internal.VerilatorBackendAnnotation
+import firrtl.AnnotationSeq
+import org.scalatest._
+
+/**
+  * circuit that illustrates usage of async register
+  * @param resetValue value on reset
+  */
+class NativeAsyncResetRegModule(resetValue: Int) extends MultiIOModule {
+  //noinspection TypeAnnotation
```
ducky64: What's this for?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+
```
ducky64: I don't think we need an empty newline here?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+import chisel3.tester.experimental.TestOptionBuilder._
+import chisel3.tester.internal.VerilatorBackendAnnotation
+import firrtl.AnnotationSeq
+import org.scalatest._
+
+/**
+  * circuit that illustrates usage of async register
+  * @param resetValue value on reset
+  */
+class NativeAsyncResetRegModule(resetValue: Int) extends MultiIOModule {
```
ducky64: I don't think it makes sense to mark this as Native, that should be implicit and default. If we have another AsyncReset test that eg uses blackboxes, that should be marked as special.

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+  //noinspection TypeAnnotation
+
+  val io = IO(new Bundle {
+    val in = Input(UInt(8.W))
+    val out = Output(UInt(8.W))
+  })
+  val asyncReset = reset.asAsyncReset
```
ducky64: Can you inline this into withClockAndReset?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  val reg = withClockAndReset(clock, asyncReset) {
+    RegInit(resetValue.U(8.W))
```
ducky64: Consider doing the combinational assign to io.out in this block, then you don't need to pass a return to withClockAndReset. The register update probably also makes more sense inside the withClockAndReset scope.

Also, can you replace withClockAndReset with just withReset?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  reg := io.in
+  io.out := reg
+}
+
+class NativeAsyncResetFeedbackModule() extends MultiIOModule {
```
ducky64: Same comments as above: what's the noinspect TypeAnnotation, prefer not needing the Native prefix, inlining the asyncReset, and moving the assign / updates to inside the reset scope.

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+  reg0 := reg1
+  reg1 := reg0
+
+  io.out(0) := reg0
+  io.out(1) := reg1
+}
+
+class AsyncResetNativeTest extends FreeSpec with ChiselScalatestTester {
+
+  val annotations: AnnotationSeq = {
```
ducky64: um wut

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      dut.io.out.expect(11.U)
+
+      dut.clock.step()
+      dut.io.out.expect(3.U)
+    }
+  }
+  "register is one, after tester startup's default reset" in {
+    test(new NativeAsyncResetRegModule(1)).withAnnotations(annotations) { dut =>
+
+      dut.reset.poke(true.B)
```
ducky64: This is redundant?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      dut.io.out.expect(1.U)
+    }
+  }
+  "reset a register works after register has been altered" in {
+    test(new NativeAsyncResetRegModule(1)).withAnnotations(annotations) { dut =>
+      // The register starts at 1 after default reset in tester startup
+      dut.io.out.expect(1.U)
+
+      // register is still zero, after it has been poked to 0
```
ducky64: Uh, what?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      dut.io.in.poke(1.U)
+      dut.io.out.expect(1.U)
+
+      // after clock is stepped, now poked input from above appears as output
```
ducky64: Well the register doesn't change value?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+  "register is one, after tester startup's default reset" in {
+    test(new NativeAsyncResetRegModule(1)).withAnnotations(annotations) { dut =>
+
+      dut.reset.poke(true.B)
+      dut.io.out.expect(1.U)
+    }
+  }
+  "reset a register works after register has been altered" in {
```
ducky64: How is this different from the first test case? what is this testing?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      // after de-assert of reset, register retains value
+      dut.reset.poke(false.B)
+      dut.io.out.expect(1.U)
+
+      // after step zero from io.in is now seen in register
+      dut.clock.step()
+      dut.io.out.expect(0.U)
+    }
+  }
+  "de-assert reset behaviour" in {
```
ducky64: I don't think there's a good argument for writing tests for the sake of writing tests? I think I'd prefer lots of smaller unit tests that test independent functionality, like reset propagating immediately, register propagating on clocks if reset not asserted, reset blocking propagation even on clock, ...

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
```diff
@@ -0,0 +1,171 @@
...
+      dut.io.out(1).expect(0.U)
+
+      dut.clock.step()
+
+      dut.io.out(0).expect(0.U)
+      dut.io.out(1).expect(1.U)
+
+      dut.reset.poke(true.B)
+
+      dut.io.out(0).expect(0.U)
```
ducky64: Maybe you should poke reset on a cycle where you expect change from the registers?

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: IntelliJ artifact, removed

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: done

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: done

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: done for all similar cases

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: gone

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: I've re-worked the tests to be more specific

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: I've re-worked the tests to be more specific

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: done

src/test/scala/chisel3/tests/AsyncResetNativeTest.scala
chick: done

src/test/scala/chisel3/tests/AsyncResetTest.scala
```diff
@@ -0,0 +1,145 @@
...
+      dut.io.in.poke(7.U)
+
+      dut.io.out.expect(11.U)
+      dut.clock.step()
+      dut.io.out.expect(11.U)
+
+      dut.reset.poke(false.B)
+      dut.io.in.poke(7.U)
+      dut.clock.step()
+      dut.io.in.poke(7.U)
```
ducky64: Did you mean expect?

src/test/scala/chisel3/tests/AsyncResetTest.scala
```diff
@@ -0,0 +1,145 @@
...
+  "reset value maintained while reset asserted, goes to input value when de-asserted" in {
+    test(new AsyncResetRegModule(11)).withAnnotations(annotations) { dut =>
+      dut.reset.poke(true.B)
+      dut.io.in.poke(7.U)
+
+      dut.io.out.expect(11.U)
+      dut.clock.step()
+      dut.io.out.expect(11.U)
+
+      dut.reset.poke(false.B)
```
ducky64: Consider throwing an expect(11.U) after the poke?

src/test/scala/chisel3/tests/AsyncResetTest.scala
```diff
@@ -0,0 +1,145 @@
...
+
+      dut.clock.step()
+      dut.io.out(0).expect(1.U)
+      dut.io.out(1).expect(0.U)
+
+      dut.clock.step()
+      dut.io.out(0).expect(0.U)
+      dut.io.out(1).expect(1.U)
+
+      dut.reset.poke(true.B)
```
ducky64: Do you want to do the reset on a cycle where you will see a change in the DUT IOs?

src/test/scala/chisel3/tests/AsyncResetTest.scala
chick: Yes, good catch

src/test/scala/chisel3/tests/AsyncResetTest.scala
chick: Added this

src/test/scala/chisel3/tests/AsyncResetTest.scala
chick: at line 145 I added a step that should toggle them, but because of poke of reset they are not toggled.

chick:
@ducky64 Fixes pushed

src/test/scala/chisel3/tests/AsyncResetTest.scala
ducky64: I meant more of why have the block in 121 since that means you don't see any difference in output after 125 because the reset value is the same as the current toggled value. But whatever.

PR #56, opened 2019-08-16
Better README, with overview and getting started

ducky64:


README.md
```diff
@@ -1,17 +1,119 @@
 # chisel-testers2
-See [freechipsproject/chisel3#725](https://github.com/freechipsproject/chisel3/issues/725).
-Discussion should happen on the issue above.
+This is **experimental software** that is currently under development, and interfaces may be subject to change (see [stability](#stability) for details).
```
chick: I think it might be better to just say alpha here (or maybe beta) those are more traditional terms for new software

README.md
```diff
@@ -1,17 +1,119 @@
...
+Testers2 integrates with the [ScalaTest](scalatest.org) framework, which provides a framework for detection and execution of unit tests.
+
+Assuming a typical Chisel project, create a new file in `src/test/scala/`, for example, `BasicTest.scala`.
+
+In this file:
+1.  Add the necessary imports:
+    ```scala
+    import org.scalatest._
+    import chisel3._
+    import chisel3.tester._
```
chick: I think you should have here (or separately) `import chisel3.experimental.BundleLiterals._`
similar arguments could be made for `import chisel3.tester.experimental.TestOptionBuilder._`
Both are reasonably likely to be necessary and finding them is not obvious, including IntelliJ's otherwise
robust auto-importing tools

README.md
ducky64: changed

README.md
ducky64: added experimental imports in the bundle literals section, and a Verilator section

ducky64:
For naming, I'd prefer to sit on it until at least chisel-dev.

ducky64:
How does this look? Anything else before this gets merged?

PR #64, opened 2019-09-16
TreadleBackend elaborates chisel module twice

chick:
- Refactor TreadleExecutive to not do that
- Saves time, looks nicer

### DIFF
--- a/src/main/scala/chisel3/tester/backends/treadle/TreadleExecutive.scala
+++ b/src/main/scala/chisel3/tester/backends/treadle/TreadleExecutive.scala
@@ -24,30 +24,37 @@ object TreadleExecutive extends BackendExecutive {
     // Force a cleanup: long SBT runs tend to fail with memory issues
     System.gc()
 
-    val annotationSeq = (new OptionsAdapter).transform(testersAnnotationSeq)
-
     val generatorAnnotation = chisel3.stage.ChiselGeneratorAnnotation(dutGen)
 
-    val circuit = generatorAnnotation.elaborate.collect { case x: ChiselCircuitAnnotation => x }.head.circuit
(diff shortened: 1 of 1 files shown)


ducky64:
Is the running twice because of the separate elaborate and the ChiselStage calls, now merged into the single Elaborate phase transform call?

Also not sure I'm onboard with using generatorAnnotations as a very mutable var... is that critical to the control flow or otherwise provides significant benefits to something?

chick:
@ducky64 
- Yes, The running twice was a direct result of running elaborate and ChiselStage.
- I feel like this is a rare case where using var makes sense
  - the method uses a series of operations that consume and emit annotations
  - not using var requires unique names for each intermediate seq of annotations
    - previous versions of this ended up with several confusing names that did not aid understanding
  - using unique names makes it very easy to accidentally use the wrong seq
    - This is particularly likely when adding functionality to this progression of operations.

seldridge:
Mutating a `var` `AnnotationSeq` begs for a `foldLeft` of some number of phases repeatedly applied to the initial `AnnotationSeq`.

This doesn't work so well when you need to get at an annotation at the end that will be deleted. Here, that's `ChiselCircuitAnnotation`. One pattern I've used for this is to use a custom [`Translator`](https://www.chisel-lang.org/api/firrtl/SNAPSHOT/firrtl/options/Translator.html). Here, that would split the input `AnnotationSeq` into two `(AnnotationSeq, AnnotationSeq)` or a `(ChiselCircuitAnnotation, AnnotationSeq)`. One path is fed to FIRRTL, the other path is "held out" and merged back into the output of FIRRTL.

As an example of this, the `firrtl.options.phases.DeletedWrapper` works this way to track what was deleted.

PR #77, opened 2019-10-22
use latest utest, fix issue utest#211

sequencer:
Sorry I forget 2.11 is still supported by chisel3.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ resolvers ++= Seq(
 
 libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "3.0.5",
-  "com.lihaoyi" %% "utest" % "0.7.1"
+  "com.lihaoyi" %% "utest" % "latest.integration"
 )
 
 testFrameworks += new TestFramework("utest.runner.Framework")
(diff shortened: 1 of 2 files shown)


ucbjrl:
Don't you also need to revert the text of the test to something compatible with 0.6.8?
```
[info] Compiling 28 Scala sources to .../chisel-testers2/target/scala-2.11/test-classes ...
[error] .../chisel-testers2/src/test/scala/chisel3/tests/UtestTesterTest.scala:11:5: not found: value test
[error]     test("This tester does not rely on scalatest to run") {
[error]     ^
[error] one error found
[error] (Test / compileIncremental) Compilation failed
[error] Total time: 247 s (04:07), completed Oct 22, 2019 3:05:58 PM
```

ucbjrl:
I believe this should be compatible with both 0.6.8 and 0.7.1:
```scala
object UtestTesterTest extends ChiselUtestTester {
  val tests: Tests = Tests {
    "This tester does not rely on scalatest to run" - {
      testCircuit(new StaticModule(42.U)) { c =>
        c.out.expect(42.U)
      }
    }
  }
}
```

sequencer:
Yes, I need to back port this too.
I just found a deprecated problem in:
```
https://github.com/lihaoyi/utest/blob/4827e4433781259573b6f222fac939dadcb8ae94/utest/src/utest/package.scala#L42
```
I'm afraid this will API break in `0.7.1` and `0.6.8`, this could be a temporary fix.
I'll try to PR or convince Haoyi to keep some back port compatibility.

PR #79, opened 2019-10-30
Fix toolchain so treadle supports boringutils

chick:
- Add the low firrtl compiler to TreadleExecutive
  - Send hint to treadle so it knows compiled form
- Fix backward compatibility with Driver
  - Check compiler type to help figure out the backend
  - Backward compatibility only supported for treadle and verilator
- Add boringutils test
- Backward compatibilty requires compile workaround
  - Can we just eliminate Driver compatibility mode here???
- Fixed CopyVpiFiles
  - looks like this might fail if early file was there but a late one wasn't
  - this code came from chisel-testers
- Fix top level test names
  - behavior of "Testers2 with Vcs"
  - behavior of "Testers2 with Verilator"
- Shortened VCS to just see if a simple circuit will build and simulate.
  - in long run we still need mechanism to run an arbitrary test with arbitrary backend

### DIFF
--- a/src/main/scala/chisel3/tester/experimental/TestOptionBuilder.scala
+++ b/src/main/scala/chisel3/tester/experimental/TestOptionBuilder.scala
@@ -5,9 +5,11 @@ package chisel3.tester.experimental
 import chisel3._
 import chisel3.tester._
 import chisel3.tester.internal.{TreadleBackendAnnotation, VerilatorBackendAnnotation}
-import firrtl.stage.CompilerAnnotation
-import firrtl.{AnnotationSeq, ExecutionOptionsManager, LowFirrtlCompiler, NoneCompiler, VerilogCompiler}
 import treadle.HasTreadleSuite
+import firrtl.stage.CompilerAnnotation
+import firrtl.{
+  AnnotationSeq, ExecutionOptionsManager, LowFirrtlCompiler, MinimumVerilogCompiler,
+  NoneCompiler, SystemVerilogCompiler, VerilogCompiler}
 
 package object TestOptionBuilder {
   implicit class ChiselScalatestOptionBuilder[T <: MultiIOModule](x: ChiselScalatestTester#TestBuilder[T]) {
@@ -18,9 +20,11 @@ package object TestOptionBuilder {
     @deprecated("Use withAnnotations instead")
     def withExecOptions(manager: ExecutionOptionsManager with HasTreadleSuite): ChiselScalatestTester#TestBuilder[T] = {
       val annos = manager.toAnnotationSeq.map {
-        case CompilerAnnotation(compiler) if compiler.isInstanceOf[LowFirrtlCompiler] => TreadleBackendAnnotation
-        case CompilerAnnotation(compiler) if compiler.isInstanceOf[NoneCompiler]      => TreadleBackendAnnotation
-        case CompilerAnnotation(_)                                                    => VerilatorBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[LowFirrtlCompiler]      => TreadleBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[NoneCompiler]           => TreadleBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[VerilogCompiler]        => VerilatorBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[MinimumVerilogCompiler] => VerilatorBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[SystemVerilogCompiler]  => VerilatorBackendAnnotation
         case anno => anno
       }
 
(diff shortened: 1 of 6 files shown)


chick:
Requires merging of [treadle PR 145](https://github.com/freechipsproject/treadle/pull/145)

src/main/scala/chisel3/tester/experimental/TestOptionBuilder.scala
```diff
@@ -15,7 +17,14 @@ package object TestOptionBuilder {
 
     @deprecated("Use withAnnotations instead")
     def withExecOptions(manager: ExecutionOptionsManager with HasTreadleSuite): ChiselScalatestTester#TestBuilder[T] = {
-      new x.outer.TestBuilder[T](x.dutGen, manager.toAnnotationSeq)
+      val annos = manager.toAnnotationSeq.map {
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[LowFirrtlCompiler] => TreadleBackendAnnotation
+        case CompilerAnnotation(compiler) if compiler.isInstanceOf[NoneCompiler]      => TreadleBackendAnnotation
+        case CompilerAnnotation(_)                                                    => VerilatorBackendAnnotation
```
ducky64: Why is this the default? Wouldn't it make more sense to error out on an unknown backend instead of having this perhaps surprising behavior?

src/test/scala/chisel3/tests/BoreTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    val y = IO(Output(UInt(6.W)))
+    val constant = Module(new Constant)
+    val expect = Module(new Expect)
+    y := expect.y
+
+    util.experimental.BoringUtils.bore(constant.x, Seq(expect.y))
+  }
+
+  it should "honor Wiring Transform for BoringUtils" in {
+//    val annos = Seq(VerilatorBackendAnnotation)
```
ducky64: Dead code

src/test/scala/chisel3/tests/BoreTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    val annos = Seq()
+    test(new Top).withAnnotations(annos) { c =>
```
ducky64: Wouldn't it make more sense to inline the Seq()?

src/test/scala/chisel3/tests/MixedVec.scala
```diff
@@ -0,0 +1,33 @@
...
+import chisel3.tester._
+import org.scalatest.FreeSpec
+
+
+class VecIO extends Bundle {
+  val x = UInt(5.W)
+}
+
+class UsesVec extends MultiIOModule {
+  val in   = IO(Input(Vec(4, new VecIO)))
```
ducky64: Can you test the Vec functionality orthogonally from the Bundle functionality (which is probably already tested)? Unless you're specifically trying to test the interaction of both?

src/test/scala/chisel3/tests/MixedVec.scala
```diff
@@ -0,0 +1,33 @@
...
+}
+
+class UsesVecSpec extends FreeSpec with ChiselScalatestTester {
+  "run" in {
+    test(new UsesVec) { c =>
+      c.in(0).x.poke(5.U)
+      c.in(1).x.poke(5.U)
+      c.in(2).x.poke(4.U)
+      c.addr.poke(2.U)
+      c.clock.step()
```
ducky64: Don't need this since the output is not registered?

src/test/scala/chisel3/tests/MixedVec.scala
```diff
@@ -0,0 +1,33 @@
...
+      c.out.expect(4.U)
```
ducky64: Want to test the other addresses too?

src/main/scala/chisel3/tester/experimental/TestOptionBuilder.scala
chick: Made the list specific for all verilog and treadle.

src/test/scala/chisel3/tests/BoreTest.scala
chick: Fixed

src/test/scala/chisel3/tests/BoreTest.scala
chick: Removed the withAnnotations call

src/test/scala/chisel3/tests/MixedVec.scala
chick: Done

src/test/scala/chisel3/tests/MixedVec.scala
chick: step removed

src/test/scala/chisel3/tests/MixedVec.scala
chick: Removed bundle

chick: Renamed this to be VecPokeExpectTest

PR #80, opened 2019-10-30
Interval support

chick:
- Adds support for Interval to peek, poke and expect
- Adds test that demonstrates circuit using Intervals instead of UInts


chick:
@ducky64 Changed interval test to be specific to Intervals and include more testing of binary points

chick:
Depends on [Chisel PR #1274](https://github.com/freechipsproject/chisel3/pull/1274)

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    val inputBinaryPoint = inputRange.binaryPoint.get
+    val outputBinaryPoint = outputRange.binaryPoint.get
+
+    test(new HasIntervalWithBinaryPoint(inputRange, outputRange)).withAnnotations(Seq(TreadleBackendAnnotation)) { c =>
+      logger.debug(f"${"in"}%10s   " + f"${"Clip"}%10s   " + f"${"Squeeze"}%10s   " + f"${"Wrap"}%10s")
+
+      for (i <- inputRange.getPossibleValues) {
```
ducky64: Not a huge fan of tests where the expected value is a chain of complex logic from the inputs, because there's a higher chance of the test being wrong from bugs in the logic. I also don't think the generator nature in this specific case buys you much: you aren't exhaustively testing anything, so it might make more sense to write a few testvectors that address the common cases (zero, one, two, another power of two, another non-power-of-two).

Furthermore, this doesn't need to test Intervals as a feature, only that the testbench can interface with them. I think I would have written something similar to the FixedPoint test in [ElementTest.scala](https://github.com/ucb-bar/chisel-testers2/blob/master/src/test/scala/chisel3/tests/ElementTest.scala)

src/test/scala/chisel3/tests/IntervalGcdTest.scala
```diff
@@ -0,0 +1,59 @@
+// See README.md for license details.
```
ducky64: Still not convinced this gives you much over the basic Interval interface test. Not a huge fan of more tests for the sake of more tests or the general scattershot approach of large, complex tests.

src/main/scala/chisel3/tester/package.scala
```diff
@@ -174,4 +183,11 @@ package object tester {
   implicit def decoupledToDriver[T <: Data](x: ReadyValidIO[T]) = new DecoupledDriver(x)
 
   implicit def validToDriver[T <: Data](x: ValidIO[T]) = new ValidDriver(x)
+
+  def intervalToBigDecimal(i: Interval): BigDecimal = {
```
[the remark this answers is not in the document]
chick: Yes, it should, but doing so can have API implications that need to be discussed and it's not a small change if done right because if should be done symmetrically for all types, FixedPoint has a potential bug related to this. See Chisel Issue #1272.
But in the meantime it would be nice to have Interval working in here.

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
```diff
@@ -0,0 +1,96 @@
+// See README.md for license details.
```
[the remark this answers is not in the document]
chick: I don't have the slightest idea how it got here. I will move it.

src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,6 +39,10 @@ package object tester {
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
       }
+      case (x: Interval, value: Interval) =>
+        require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
+        pokeBits(x, value.litValue)
+
```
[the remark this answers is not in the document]
chick: removed

src/main/scala/chisel3/tester/package.scala
```diff
@@ -85,6 +91,9 @@ package object tester {
         require(x.binaryPoint == value.binaryPoint, s"binary point mismatch between value $value from IO $x")
         Context().backend.expectBits(x, value.litValue, message, stale)
       }
+      case (x: Interval, value: Interval) =>
+        Context().backend.expectBits(x, value.litValue, message, stale)
+
```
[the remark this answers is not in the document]
chick: removed

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
```diff
@@ -0,0 +1,96 @@
...
+    val out1Wrap = Output(Interval(outputRange))
+  })
+
+  io.out1 := io.input
+  io.out1Clip := io.input.clip(io.out1Clip)
+  io.out1Squeeze := io.input.squeeze(io.out1Squeeze)
+  io.out1Wrap := io.input.wrap(io.out1Wrap)
+}
+
+class IntervalTest extends FreeSpec with ChiselScalatestTester with LazyLogging with Matchers {
```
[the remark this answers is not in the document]
chick: LazyLogging provides the framework for logger.debug

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
```diff
@@ -0,0 +1,96 @@
...
+    val inputBinaryPoint = inputRange.binaryPoint.get
+    val outputBinaryPoint = outputRange.binaryPoint.get
+
+    test(new HasIntervalWithBinaryPoint(inputRange, outputRange)).withAnnotations(Seq(TreadleBackendAnnotation)) { c =>
+      logger.debug(f"${"in"}%10s   " + f"${"Clip"}%10s   " + f"${"Squeeze"}%10s   " + f"${"Wrap"}%10s")
+
+      for (i <- inputRange.getPossibleValues) {
+        c.io.input.poke(i.I(binaryPoint = inputBinaryPoint.BP))
+
+        logger.debug(
```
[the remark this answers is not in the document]
chick: Having worked on this stuff multiple times now, it can be really tricky to figure out what's wrong with things particularly with edge cases. When I've written the same debugging code multiple times. I sometimes want to keep it so I don't have to re-invent it when something changes in the future. I believe logging here provides a modestly intrusive way of doing this. There are potential changes to the firrtl as we further debug intervals that make it exceedingly like that I will be debugging this code once again

src/main/scala/chisel3/tester/package.scala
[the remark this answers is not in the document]
ducky64: So what's the long-term plan for this? Once the chisel3 proper fix comes out, this goes away (unless you have an argument for two ways to do the same thing?). Which means unless it's API compatible with the planned chisel3 feature (which might cause compile-time aliasing issues?) it's going to break everyone's code.

If you want to use it as an experimental or internal feature, sure, but I don't want to give the illusion of long term support for a construct that is almost certainly disappearing in the near future. Either because we implemented it in chisel3 and it's now redundant, or for whatever reason we decided to not do it in chisel3 and those same reasons probably apply here too.

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
chick: Each of these tests turned up unexpected bugs in both the Interval implementation and in the tools that let one iterate over the values which. If I could reasonably anticipate what the edge cases are I could write a bunch of individual test, but I don't think that's easy to do. Having software models for things can be a problem but it's reasonable approach to testing that your code does what you think it does. Firrtl fixes might reveal errors in the model but I'd like to know when the model is flawed.

src/test/scala/chisel3/tests/IntervalGcdTest.scala
chick: The tests revealed things that I am not sure I would have found with a more spartan approach. I also would not underestimate the value of giving example usage of things in the tests suites. It might be better to create other venues for that but it is a lot of work. Whereas here the work is done in direct relation to functional testing.

src/main/scala/chisel3/tester/legacy/backends/vcs/IntervalTest.scala
ducky64: So does this test belong here, or in chisel3 core, or in firrtl, or in treadle? The tests here aren't meant to be the chisel3 regression suite - and having such tests here means they aren't run as part of the chisel3 pre-merge test suite.

It might be useful in that it is testing something, but I don't think most of what this is testing is in this repo?

src/test/scala/chisel3/tests/IntervalGcdTest.scala
ducky64: This might be a chisel cookbook example or something, but I'm not sure it belongs here. This is arguably more of how to build circuits in chisel3 w/ intervals than how to write tests in chiseltest. I'm sure there's a place for it somewhere (maybe "100 variations on GCD" or something) but I'm not sure that place is here...

chick: Yeah, this was a mistake. I think I have it deleted it now. Previous push failed somehow

chick:
Latest commit depends on [Chisel PR #1275](https://github.com/freechipsproject/chisel3/pull/1275)

src/main/scala/chiseltest/package.scala
```diff
@@ -167,4 +174,11 @@ package object chiseltest {
   implicit def decoupledToDriver[T <: Data](x: ReadyValidIO[T]) = new DecoupledDriver(x)
 
   implicit def validToDriver[T <: Data](x: ValidIO[T]) = new ValidDriver(x)
+
+  def intervalToBigDecimal(i: Interval): BigDecimal = {
```
ducky64: I still don't think this belongs in the base API. Perhaps as experimental or internal, but there should be no expectation of continued support here since this is going into chisel base?

src/test/scala/chiseltest/tests/ElementTest.scala
```diff
@@ -125,4 +125,42 @@ class ElementTest extends FlatSpec with ChiselScalatestTester {
...
+
+      def checkOutcome(input: Interval, clipped: Interval, wrapped: Interval): Unit = {
+        c.io.input.poke(input)
+
+        c.io.out1.expect(input)
+        c.io.out1Clip.expect(clipped)
+        c.io.out1Wrap.expect(wrapped)
+      }
+
+      val in1 = (-6.0).I(inputRange)
```
ducky64: Whats this line for

src/test/scala/chiseltest/tests/ElementTest.scala
```diff
@@ -125,4 +125,42 @@ class ElementTest extends FlatSpec with ChiselScalatestTester {
...
+      val io = IO(new Bundle {
+        val input = Input(Interval(inputRange))
+
+        val out1 = Output(Interval(inputRange))
+        val out1Clip = Output(Interval(outputRange))
+        val out1Squeeze = Output(Interval(outputRange))
+        val out1Wrap = Output(Interval(outputRange))
+      })
+      io.out1 := io.input
+      io.out1Clip := io.input.clip(io.out1Clip)
```
ducky64: This looks slightly suspect. I assume what this means is io.out1Clip := io.input clipped to the interval defined by io.out1Clip, where clip ignores the hardware value of the argument? In that case, would it be cleaner to make that drop-hardware-value explicit with a chiselTypeOf operation, or is that value still somehow necessary?

ducky64: Resolved in offline discussion: appears the hardware node is needed by FIRRTL to get the type (in FIRRTL) of the intervals. So not a problem to be solved here, though might be worth revisiting in chisel.

ucbjrl:
Unfortunately, this triggered the following under Jenkins:
```
[info] ElementTest:
[info] Testers2 with Element types
[info] - should work with UInt
[info] - should work with SInt
[info] - should work with Bool
[info] - should work with FixedPoint
[info] - should work with Interval *** FAILED ***
[info]   chisel3.internal.ChiselException: Error literal interval value -24 is not contained in specified range range"[-6,6].2"
[info]   at chisel3.experimental.Interval$.Lit(Bits.scala:2173)
[info]   at chisel3.experimental.Interval$Implicits$fromBigDecimalToLiteralInterval.I(Bits.scala:2309)
[info]   at chiseltest.tests.ElementTest.$anonfun$new$15(ElementTest.scala:157)
[info]   at chiseltest.tests.ElementTest.$anonfun$new$15$adapted(ElementTest.scala:147)
[info]   at chiseltest.backends.treadle.TreadleBackend.$anonfun$run$1(TreadleBackend.scala:144)
[info]   at chiseltest.internal.ThreadedBackend$TesterThread$$anon$1.$anonfun$run$1(ThreadedBackend.scala:453)
[info]   at chiseltest.backends.treadle.TreadleBackend.doTimescope(TreadleBackend.scala:103)
[info]   at chiseltest.internal.ThreadedBackend$TesterThread$$anon$1.run(ThreadedBackend.scala:453)
[info]   at java.lang.Thread.run(Thread.java:748)
```

chick:
This is fixed by [Chisel3 Fixed problem creating Interval literals with full ranges #1275](https://github.com/freechipsproject/chisel3/pull/1275)

PR #84, opened 2019-11-10
Fix external links in README

edwardcwang:
No functional changes.


edwardcwang:
Not sure, though it seems like the old syntax was for repo-relative links

PR #87, opened 2019-11-19
Suggestion for better errors on underspecifed bundle literal

chick:
- Add LiteralValue Exception
- Catch NoSuchElementException on Bundle poke to identify missing field.
- Add test to show new error

### DIFF
--- a/src/main/scala/chisel3/tester/package.scala
+++ b/src/main/scala/chisel3/tester/package.scala
@@ -9,6 +9,7 @@ import chisel3.experimental.BundleLiterals._
 import chisel3.util._
 
 class NotLiteralException(message: String) extends Exception(message)
+class LiteralValueException(message: String) extends Exception(message)
 class LiteralTypeException(message: String) extends Exception(message)
 class UnpokeableException(message: String) extends Exception(message)
 class UnsupportedOperationException(message: String) extends Exception(message)
@@ -39,11 +40,19 @@ package object tester {
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
       }
-      case (x: Bundle, value: Bundle) => {
+      case (xx: Bundle, value: Bundle) => {
         // TODO: chisel needs to expose typeEquivalent
-        require(x.elements.keys == value.elements.keys)  // TODO: this discards type data =(
-        (x.elements zip value.elements) foreach { case ((_, x), (_, value)) =>
-          x.poke(value)
+        require(xx.elements.keys == value.elements.keys)  // TODO: this discards type data =(
+        (xx.elements zip value.elements) foreach { case ((fieldName, xxx), (_, value)) =>
+          try {
+            xxx.poke(value)
+          }
+          catch {
+            case _: NoSuchElementException =>
+              val bundleFieldName = x.toString.split("[(]").head + "." + fieldName
+              throw new LiteralValueException(
+                s"""Error: Maybe bundle field $bundleFieldName was not given a value during poke""")
+          }
         }
       }
       case x => throw new LiteralTypeException(s"don't know how to poke $x")
--- a/src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
+++ b/src/test/scala/chisel3/tests/BundleLiteralsSpec.scala
@@ -69,4 +69,13 @@ class BundleLiteralsSpec extends FlatSpec with ChiselScalatestTester with Matche
       c.out.expect(chiselTypeOf(c.in).Lit(_.a -> 0.U, _.b -> 1.U))
     }
   }
+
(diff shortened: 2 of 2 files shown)


src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,11 +40,19 @@ package object tester {
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
       }
-      case (x: Bundle, value: Bundle) => {
+      case (xx: Bundle, value: Bundle) => {
```
ducky64: Why is this `xx`?

src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,11 +40,19 @@ package object tester {
...
         // TODO: chisel needs to expose typeEquivalent
-        require(x.elements.keys == value.elements.keys)  // TODO: this discards type data =(
-        (x.elements zip value.elements) foreach { case ((_, x), (_, value)) =>
-          x.poke(value)
+        require(xx.elements.keys == value.elements.keys)  // TODO: this discards type data =(
+        (xx.elements zip value.elements) foreach { case ((fieldName, xxx), (_, value)) =>
```
ducky64: I'm pretty sure chaining more `x`s into a variable name does not make it better

src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,11 +40,19 @@ package object tester {
...
+          try {
```
ducky64: Not a fan of try/catch if alternatives are available. NoSuchElementException does not seem to be a stable API to rely on either. Can you do litOption or something instead?

src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,11 +40,19 @@ package object tester {
...
+            xxx.poke(value)
+          }
+          catch {
+            case _: NoSuchElementException =>
+              val bundleFieldName = x.toString.split("[(]").head + "." + fieldName
```
ducky64: What is this weird code doing? It seems it's dependent on a debugging representation which isn't guaranteed to be stable. I'd prefer stability over making the display super pretty, unless it's unusable or something without this split thingy.

src/main/scala/chisel3/tester/package.scala
```diff
@@ -39,11 +40,19 @@ package object tester {
...
+              throw new LiteralValueException(
+                s"""Error: Maybe bundle field $bundleFieldName was not given a value during poke""")
```
ducky64: If you do use litOption, consider having this display both the underlying cause (incomplete bundle literal) and recommended resolution (consider filling in that field). Or we could support partial bundle literals, but that's another can of worms.

src/main/scala/chisel3/tester/package.scala
chick: Probably doesn't need to be but as I was trying to figure out which vars had the information I needed it was hard because you used x in three nested contexts. All the style checkers I have seen complain about variable shadowing in these cases. Firrtl adopts the convention of just duplicating the short var name so x -> xx -> xxx through the context drops.

src/main/scala/chisel3/tester/package.scala
chick: It's better than all of them being the same name. This was particularly complicated when I did some work for Adam and Jack on cleaning up style in some older firrtl sections. It was difficult to figure out which var was in play in many cases.

src/main/scala/chisel3/tester/package.scala
chick: I was mostly trying to illustrate an improved error message. The message without this code `None.get` is pretty unhelpful and will be even more of a problem with real complex nested bundles like one sees in riscv-mini and rocket. Stability and robustness are important of course. I did not see an immediate way of getting a better access to the name of the Bundle involved.

src/main/scala/chisel3/tester/package.scala
chick: Hack to get get bundle name.  I'm sure there's a better way

src/main/scala/chisel3/tester/package.scala
chick: I'm not quite sure how one would use litOption in this context. I think, for a minute, of proposing to relax the complete bundle requirement, but shied away from that.

ducky64:
Like is the intent of this PR for it to get merged or is it just a proof-of-concept that is not intended to be production code? I generally assume PRs to be the former, but if it's the latter then it should be clarified because the back and forth on reviewing and dehackifying code that is known by everyone to be hacky is not terribly productive.

chick:
Fair point, I will close this PR for now and resubmit later. I've discovered a few more edge cases where the messages on poke can be less than helpful and a more complete treatment should be done

PR #88, opened 2019-11-20
Rename testers2 to ChiselTest

ducky64:
We settled on ChiselTest (with package name `chiseltest`) during the previous chisel-dev meeting. This updates the README to reflect that.

One open question: while the sbt package will be `chiseltest`, the import is still `chisel3.tester`. I'm not sure about whether we want to rename that, and what to rename to. `import chiseltest._`? `import chisel3.test._`? `import chisel3.chiseltest._`? A compatibility shim should allow existing imports to work for a while.

@ucbjrl when this PR goes through, can you also change the repo name to ChiselTest and (maybe? not sure about implications) move it to freechipsproject? Also, is it possible to publish both under sbt package `chisel-testers2` and `chiseltest` for some limited migration period?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,23 @@
-# chisel-testers2
+# ChiselTest
+*formerly known as testers2*
+
 This is **alpha software** that is currently under development, and interfaces may be subject to change (see [stability](#stability) for details).
 However, it is very much in a usable state, so if you're fine living on the bleeding edge, give it a try. 
 
-
 ## Overview
-Testers2 is a test harness for [Chisel](https://github.com/freechipsproject/chisel3)-based RTL designs, currently supporting directed testing (all test stimulus manually specified - no constrained random and coverage-driven flows).
-Testers2 emphasizes tests that are lightweight (minimizes boilerplate code), easy to read and write (understandability), and compose (for better test code reuse).
+ChiselTest is a test harness for [Chisel](https://github.com/freechipsproject/chisel3)-based RTL designs, currently supporting directed testing (all test stimulus manually specified - no constrained random and coverage-driven flows).
+ChiselTest emphasizes tests that are lightweight (minimizes boilerplate code), easy to read and write (understandability), and compose (for better test code reuse).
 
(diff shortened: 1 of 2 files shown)


edwardcwang:
Is ChiselTest sufficiently disambiguated from chisel-testers?

ucbjrl:
I think of the `import chisel3.tester` as the API namespace. The `chiseltest` package is an implementation of this API and as such, it's fine (maybe even better) if the names are different.

Can we just freeze `chisel-testers2` in the 0.1.x branch and publish 0.2.x releases as `chiseltest` or do you want identical code published as both `chiseltest` and `chisel-testers2`? If the latter, and we use a single branch for both, I suggest we do the name change between `+publishSigned`  steps and don't commit the change to `build.sbt` that changes the name (i.e., going forward the official name will become `chiseltest` and we won't have committed versions of `build.sbt` with the name `chisel-testers2`).

ducky64:
Resolution from the meeting today:
- The (maven/ivy/sbt) package name will be `chiseltest`, as is standard practice with Scala projects
- The Scala package name will be `import chiseltest._`, more or less consistent with ScalaTest. 
  - Note, the chisel3 regression tests are in the package `chiselTests` but we can rename that since those are mainly internal. Also for reference, the chisel3 BasicTester is in `chisel3.tester.BasicTester`.
  - There was concern about package aliasing not playing well with incremental compilation, which was why we avoided a subpackage in the `chisel3` namespace.
- This change will go in ASAP, but we will maintain Scala package compatibility with `import chisel3.testers._` through a compatibility package (much like `import Chisel._`) and dual publish under `chisel-testers2` at least until Chisel 3.3.0. This allows those working under the existing namespaces to continue using those, and allows new dev to use the new namespaces immediately.

ucbjrl:
It's still a little unclear to me to what the name `ChiselTest` applies. Is it the github repo/directory containing the source code? Is the capitalization important?

ducky64:
It's just the repo name / nice name. ChiselTest (with capitalization) is not used in the source code at all.

Issue #89, opened 2019-11-25
implicit conver to DecoupledIO does not work


ShuyunJia:
The `TestAdapters`  suggest to use implicit conversion to `DecoupledDriver` 
https://github.com/ucb-bar/chisel-testers2/blob/ea5e81f2e54f2df4d5cd34d8e7ece2a69fab0f86/src/main/scala/chisel3/tester/TestAdapters.scala#L9

It woks fine for `DecoupledIO`. However, it does not work for `IrrevocableIO`. I have to force the type conversion to `DecoupledIO`: 
```
it should "run implicit convert" in {
    test(new Module{
      val io = IO( new Bundle{
        val in = Flipped(Irrevocable(UInt(8.W)))
        val out = Irrevocable(UInt(8.W))
      })
    }).withAnnotations(Seq(WriteVcdAnnotation)) { c =>
      Decoupled(c.io.in).initSource.setSourceClock(c.clock)
    }
  }
```

ducky64:
Resolved by #90, I think?

PR #90, opened 2019-11-26
support enqueue/dequeue of IrrevocableIO

ShuyunJia:
Signed-off-by: shuyun <shuyunjia@outlook.com>

issue related: #89

This commit is to: 
1. support enqueue/dequeue of IrrevocableIO
2. add implicit convert of IrrevocableIO to ReadyValidIO

### DIFF
--- a/src/main/scala/chisel3/tester/package.scala
+++ b/src/main/scala/chisel3/tester/package.scala
@@ -173,5 +173,7 @@ package object tester {
 
   implicit def decoupledToDriver[T <: Data](x: DecoupledIO[T]) = new DecoupledDriver(x)
 
+  implicit def irrevocableToDriver[T <: Data](x: IrrevocableIO[T]) = new DecoupledDriver(x)
+
   implicit def validToDriver[T <: Data](x: ValidIO[T]) = new ValidDriver(x)
 }
--- a/src/main/scala/chisel3/tester/package.scala
+++ b/src/main/scala/chisel3/tester/package.scala
@@ -171,9 +171,7 @@ package object tester {
     }
   }
 
-  implicit def decoupledToDriver[T <: Data](x: DecoupledIO[T]) = new DecoupledDriver(x)
-
-  implicit def irrevocableToDriver[T <: Data](x: IrrevocableIO[T]) = new DecoupledDriver(x)
+  implicit def decoupledToDriver[T <: Data](x: ReadyValidIO[T]) = new DecoupledDriver(x)
 
   implicit def validToDriver[T <: Data](x: ValidIO[T]) = new ValidDriver(x)
 }
--- a/src/test/scala/chisel3/tests/QueueTest.scala
+++ b/src/test/scala/chisel3/tests/QueueTest.scala
@@ -4,6 +4,8 @@ import org.scalatest._
 
 import chisel3._
 import chisel3.tester._
+import chisel3.util._
+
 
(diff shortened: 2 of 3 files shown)


src/main/scala/chisel3/tester/package.scala
```diff
@@ -173,5 +173,7 @@ package object tester {
 
   implicit def decoupledToDriver[T <: Data](x: DecoupledIO[T]) = new DecoupledDriver(x)
 
+  implicit def irrevocableToDriver[T <: Data](x: IrrevocableIO[T]) = new DecoupledDriver(x)
```
ducky64: Will an implicit conversion from ReadyValidIIO instead of DecoupledIO work instead of requiring an implicit conversion for each individually?

src/test/scala/chisel3/tests/QueueTest.scala
```diff
@@ -57,4 +59,28 @@ class QueueTest extends FlatSpec with ChiselScalatestTester {
...
+    test(new Module{
+      val io = IO(new Bundle{
+        val in = Flipped(Irrevocable(UInt(8.W)))
+        val out = Irrevocable(UInt(8.W))
+      })
+      val full = RegInit(false.B)
+      when(io.in.fire()){ full := true.B }.elsewhen(io.out.fire()){ full := false.B }
+      io.in.ready := !full
+      io.out.valid := full
+      io.out.bits := RegNext(io.in.bits)
```
ducky64: Would it make more sense to just instantiate a queue here, instead of essentially rolling your own one-element queue which doesn't appear to be fully correct (the data bits updates every cycle, regardless of io.in.ready)?

src/test/scala/chisel3/tests/QueueTest.scala
```diff
@@ -57,4 +59,28 @@ class QueueTest extends FlatSpec with ChiselScalatestTester {
...
+    }){c =>
+      c.io.in.initSource().setSourceClock(c.clock)
+      c.io.out.initSink().setSinkClock(c.clock)
+      c.io.in.enqueue(9.U)
+      c.io.out.expectDequeue((9.U))
+      parallel(
+        c.io.in.enqueueSeq(Seq(5.U, 2.U)),
```
ducky64: Would it make sense to add the 9.U above to the enqueueSeq? Or is the above meant to specifically test cycle-by-cycle behavior?

src/main/scala/chisel3/tester/package.scala
ShuyunJia: oh yes, it is much better to convert from ReadyValidIO.  
Fixed in 49d2857.

src/test/scala/chisel3/tests/QueueTest.scala
ShuyunJia: The original purpose of instantiate a different queue is to distinguish between `DecoupledIO` and `IrrevocableIO`. Since there are additional requirements on signaling of `valid` and `ready`. I looked into the queue object and it satisfy the requirement. 
Fixed in  49d2857

src/test/scala/chisel3/tests/QueueTest.scala
ShuyunJia: It is just some redundant cycly-by-cycle test. Removed in 49d2857

ShuyunJia:
And, I agree to dual-license my contribution under Apache 2 and BSD 3-clause.

PR #93, opened 2019-12-02
Rename package to chiseltest._

ducky64:
The title is self-explanatory. Adds a compatibility shim package so old code should continue compiling. Deprecated code like TestAdapters are kept in the old package.

I hope I didn't miss anything in the compatibility shim.

Related: #88. This is orthogonal, and only touches the package names in Scala code, not the ivy/maven package names.


ducky64:
TODO see if deprecations are possible in the compatibility package

PR #95, opened 2019-12-06
add verilog testbench dump trait

ShuyunJia:
Signed-off-by: shuyun <shuyunjia@outlook.com>
draft design for dumping verilog testbench
not fully tested


ekiwi:
I like the idea, but I agree with @ducky64 that this should ideally be implemented with less code duplication. In case anyone wants to pick up the work, let us know and we can give you some feedback!

PR #98, opened 2019-12-13
tests: check that elaboration happens only once

oharboe:
### Build Dependencies

### DIFF
--- a/src/test/scala/chiseltest/tests/SingleElaboration.scala
+++ b/src/test/scala/chiseltest/tests/SingleElaboration.scala
@@ -1,0 +1,41 @@
+// See LICENSE for license details.
+
+package chiseltest.tests
+
+import chisel3._
+import chiseltest._
+import chiseltest.experimental.TestOptionBuilder._
+import chiseltest.internal.VerilatorBackendAnnotation
+import firrtl.AnnotationSeq
+import org.scalatest._
+
+class SingleElaborationTests extends FlatSpec with ChiselScalatestTester with Matchers {
+
+  behavior of "elaboration should happen once"
+
+  var checkOnceVerilator = 0
+  it should ("elaborate once with Verilator") in {
+    new TestBuilder(() => {
+
+      checkOnceVerilator should be(0)
+      checkOnceVerilator = checkOnceVerilator + 1
+
+      new DummyModule
+    }, Seq.empty, Array[String]()).withAnnotations(Seq(VerilatorBackendAnnotation)) { dut =>
+      dut.clock.step(1)
+    }
+  }
+
+  var checkOnceTreadle = 0
+  it should ("elaborate once with Treadle") in {
+    new TestBuilder(() => {
+
+      checkOnceTreadle should be(0)
+      checkOnceTreadle = checkOnceTreadle + 1
+
+      new DummyModule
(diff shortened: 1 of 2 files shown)


src/test/scala/chiseltest/tests/SingleElaboration.scala
```diff
@@ -0,0 +1,50 @@
...
+
+class SingleElaborationTests extends FlatSpec with ChiselScalatestTester with Matchers {
+
+  behavior of "elaboration should happen once"
+
+  class ThingAMaBobBundle extends Bundle {
+    val condition = Input(Bool())
+  }
+
+  class ThingAMaBob extends Module {
```
ducky64: Is this module critical to test functionality, or can you use one of the standard modules in TestUtils (and maybe we should also have an empty DummyModule there too, since I think the module contents are orthogonal to this test case)?

src/test/scala/chiseltest/tests/SingleElaboration.scala
```diff
@@ -0,0 +1,50 @@
...
+    val io = IO(new ThingAMaBobBundle)
+    val counter = RegInit(0.U(2.W))
+  }
+
+  var checkOnceVerilator = 0
+  it should ("elaborate once with Verilator") in {
+    new TestBuilder(() => {
```
ducky64: Can you have the variable increment inside the Module body itself, like
```scala
test(new Module {
  checkOnceVerilator should be (0)
  checkOnceVerilator += 1
}) { c =>
  // empty test body
}
```

You can also use the `test(...).withAnnotations(...) { c => ... }` syntax as demonstrated in [VerilatorBasicTests.scala](https://github.com/ucb-bar/chisel-testers2/blob/master/src/test/scala/chiseltest/experimental/tests/VerilatorBasicTests.scala), which I think is cleaner than creating a TestBuilder object?

src/test/scala/chiseltest/tests/SingleElaboration.scala
oharboe: I created a DummyModule

src/test/scala/chiseltest/tests/SingleElaboration.scala
oharboe: test() syntax can't be used, because it doesn't take a producer of a module, but an actual module.

oharboe: > Can you have the variable increment inside the Module body itself, like

That breaks the test...

ducky64: Wait, why does it break the test? `test(...)` is call-by-name on the DUT argument (essentially it's the same as `() => ...` but without the explicit syntax), and also I don't think it's allowed to (should assert out) elaborate a Chisel module outside a special context (that gets initialized inside somewhere `test(...)`)

oharboe:
I dont know exactly. I am using vscode and it doesnt support debugging yet
. Also, I am learning Scala in spare moments so my understanding is shaky.

ons. 18. des. 2019, 20:28 skrev Richard Lin <notifications@github.com>:

oharboe:
@ducky64 I think that there is a unit-test that verifies that elaboration doesn't happen twice? Is there? Can I close this PR?

oharboe:
I can re-open if this still has any value.

PR #111, opened 2020-01-15
ChiselEnum support

### LINKED ISSUE #12: Support StrongEnum
$subject

ducky64:
Partial for #12

`poke` and `expect` work as expected. `peek` does not, because I don't think there's a programmatic literal construction API for ChiselEnum.

`pokeBits` has been expanded to Data in general. I'm not sure if this is the right solution (alternatively, ChiselEnum can be Bits - because it's very Bits-like). While expanding the pokeBits API makes it work on things like Clock (and eliminates the need for a pokeClock API) it becomes less type safe (should pokeBits be allowed on Bundle? probably not - that is handled at a higher level). Alternatives include pokeElement (but that wouldn't work on Analog), or overloading pokeBits to support Bits, Clock, and ChiselEnum.

### DIFF
--- a/src/main/scala/chiseltest/backends/treadle/TreadleBackend.scala
+++ b/src/main/scala/chiseltest/backends/treadle/TreadleBackend.scala
@@ -61,7 +61,7 @@ extends BackendInstance[T] with ThreadedBackend[T] {
     a > 0
   }
 
-  override def pokeBits(signal: Bits, value: BigInt): Unit = {
+  override def pokeBits(signal: Data, value: BigInt): Unit = {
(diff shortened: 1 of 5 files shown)


ducky64:
Expect already works, since it works on bits directly and does not need to construct a new literal.

Peek theoretically isn't too difficult, but it requires plumbing in chisel - given a hardware object and an integer value, return the corresponding enum literal.

kammoh:
Impatiently waiting for this to be merged in :)

ducky64:
Sure, I'll merge this as-is and we can deal with enum pokes later.

PR #121, opened 2020-02-13
Make scala-steward ignore edu.berkeley.cs deps

seldridge:
My reading of the scala-steward documentation means that adding this file will tel scala-steward to ignore any dependencies coming from `edu.berkeley.cs`. This should be the minimal change to stop the snapshot updating.

Completely untested, though...

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,0 +1,1 @@
+updates.ignore = [ { groupId = "edu.berkeley.cs" } ]


ucbjrl:
I say, GoFoRiT.

ducky64:
Is this PR outdated?

ucbjrl:
I think this is still relevant. If it does work here, something similar should be incorporated in the other ecosystem repositories, since existing attempts at achieving the same effect don't seem to be working.

ducky64:
But we already have a [`.scala-steward.conf`](https://github.com/ucb-bar/chisel-testers2/blob/master/.scala-steward.conf) file that seems to be a more targeted version of this PR?

PR #127, opened 2020-02-21
Bug fix for verilator backend

sequencer:
Currently verilator backend runs `chisel3.stage.phases.Elaborate` twice, first one is generating circuit, second is generating verilog.
In first elaborate phase will leave `chisel3.internal.Builder` to a wrong state, which leads to second elaborate phase throw a exception "attempted to instantiate a Module, but nothing happened".


sequencer:
Thanks:)

PR #131, opened 2020-02-23
User-friendly hint when forgetting join after fork

edwardcwang:
"non-enclosed timescope" can be a pretty cryptic message from a user's perspective


ducky64:
Yeah, so join isn't strictly necessary and (depending on where it is positioned) may not solve the problem.

The underlying issue that the (somewhat cryptic - suggestions welcome!) message is trying to communicate is that timescopes must be fully enclosed (think of structured programming constructs), even across child-parent threads. Every thread has an implicit root timescope - so in essence, a spawning thread can't close its timescope before its child threads finish (but there are no dependencies between child threads).

Timescopes are kind of a new construct, so feedback and suggestions on semantics (and how to communicate them) are welcome.

edwardcwang:
Well, it is a usability problem as it stands. Maybe you can add some additional messaging to the error? At the moment it is an extremely cryptic message for something as simple as forgetting `join`.

ekiwi:
Seems like we could not find consensus on this. Feel free to re-open if you want to revisit the discussion.

PR #133, opened 2020-02-27
Fix CI testing problem with OptionsPassingTest

chick:
- Calls to .withFlags and .withAnnotations now keeps values from previous invocations
- removed unused logger import
- made flags and annos vals in TestBuilder so they can be retained for subsequent .with* invocations
- updated scaladoc on `def test` to mention withFlags
- Fixed `it should "allow specifying configuration options using annotations and CLI style flags"` test
  - It needed all three output file annotations to make a test run directory with consistently named files

### DIFF
--- a/src/main/scala/chiseltest/ChiselScalatestTester.scala
+++ b/src/main/scala/chiseltest/ChiselScalatestTester.scala
@@ -14,9 +14,7 @@ import scala.util.DynamicVariable
 
 trait ChiselScalatestTester extends Assertions with TestSuiteMixin with TestEnvInterface { this: TestSuite =>
 
-  class TestBuilder[T <: MultiIOModule](val dutGen:        () => T,
-                                        val annotationSeq: AnnotationSeq,
-                                        val flags:         Array[String]) {
+  class TestBuilder[T <: MultiIOModule](val dutGen: () => T, val annotationSeq: AnnotationSeq, val flags: Array[String]) {
     def getTestName: String = {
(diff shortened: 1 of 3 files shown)


src/main/scala/chiseltest/ChiselScalatestTester.scala
```diff
@@ -9,13 +9,14 @@ import firrtl.AnnotationSeq
...
 import org.scalatest.exceptions.TestFailedException
 import internal.WriteVcdAnnotation
-import logger.Logger
 
 import scala.util.DynamicVariable
 
 trait ChiselScalatestTester extends Assertions with TestSuiteMixin with TestEnvInterface { this: TestSuite =>
 
-  class TestBuilder[T <: MultiIOModule](val dutGen: () => T, annotationSeq: AnnotationSeq, flags: Array[String]) {
+  class TestBuilder[T <: MultiIOModule](val dutGen:        () => T,
```
ducky64: what changed here other than formatting? did we reach consensus on this style?

chick: The introduction of `val`s made this line too long.  And although we have not nailed down our formatting style. This formatting is the one we were considering adopting.

ducky64: Not a fan of the one argument per line style, so unless that was one of the style points for which there was a consensus, I'd prefer it be backed out,

chick: backed it out

PR #135, opened 2020-03-03
Don't mess with version management.

ucbjrl:


ducky64:
I think there's some marginal utility to the bot reminding us to update major versions, but the snapshot bumping is useless. Is there a way to prevent the bot from seeing chisel (and related projects) snapshot versions across the board?

ucbjrl:
My understanding is this has to be done on a pre-repository basis. This PR attempts to do that for chisel-testers2.

PR #144, opened 2020-03-26
Fix a problem where negative values sign extension causes incorrect

chick:
behavior when the targeted input is not a width that is a power of 2.
provides test that will fail without this fix.


chick:
@ducky64 Test made simpler, println removed

PR #150, opened 2020-04-03
Add support for GitHub workflows.

ucbjrl:
Since no build dependencies are explicitly specified, GitHub will use whatever has been published on maven.


sequencer:
I pr test depends on the newest master version of chisel3?
I saw the error in: https://github.com/ucb-bar/chisel-testers2/pull/151/checks?check_run_id=580596759
But it has been already upstreamed.

PR #151, opened 2020-04-06
Partial poke to Record data. 

sequencer:
Depends on freechipsproject/chisel3#1400. In this commit, duplex `Record` peek/poke is enabled for complex IO like
`TLBundle` or `AutoBundle`.
1. replace `litValue` with `litOption.map(v => doSometing)`, for the use
case of Record partial assignment.
2. use `elements.keys` to match `x` and `value` field, this will result
   poke/except the element with same key and not `None`.

### Build Dependencies

### DIFF
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -12,6 +12,51 @@ import chisel3.util._
 package object chiseltest {
   import chisel3.internal.firrtl.{LitArg, ULit, SLit}
 
+  implicit class testableRecord[T <: Record](x: T) {
+    /** Poke the given signal with a [[Record.litValue()]]
+      * Literals of this Record can be instantiated with
+      * {{{
+      *   someRecord.Lit(_.elements("foo") -> 4.U)
+      * }}}
+      * `pokePartial` will only poke [[Input]] signals of `x`,
+      * and elements of `x` which contain no literal will be ignored.
+      * */
+    def pokePartial(value: T): Unit = {
+      require(DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
+      x.elements
+        .filter {
+          case (k, v) => DataMirror.directionOf(v) != ActualDirection.Output && {
+            value.elements(k) match {
+              case _: Record => true
+              case data: Data => data.isLit()
+            }
+          }
+        }
+        .foreach { case (k, v) => v match {
+          case record: Record => record.pokePartial(value.elements(k).asInstanceOf[Record])
+          case data: Data => data.poke(value.elements(k))
+        }}
+    }
+
+    /** Check the given signal with a [[Record.litValue()]];
+      * elements of `x` which contain no literal will be ignored.
+      * */
+    def expectPartial(value: T): Unit = {
+      require(DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
+      x.elements
+        .filter {
+          case (k, v) => value.elements(k) match {
+              case _: Record => true
+              case d: Data => d.isLit()
+            }
+        }
+        .foreach { case (k, v) => v match {
+          case record: Record => record.expectPartial(value.elements(k).asInstanceOf[Record])
+          case data: Data => data.expect(value.elements(k))
+        }}
+    }
+  }
+
   implicit class testableData[T <: Data](x: T) {
     protected def pokeBits(signal: Data, value: BigInt): Unit = {
       if (DataMirror.directionOf(signal) != Direction.Input) {
@@ -37,15 +82,14 @@ package object chiseltest {
       case (x: Interval, value: Interval) =>
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
-      case (x: Bundle, value: Bundle) => {
-        // TODO: chisel needs to expose typeEquivalent
-        require(x.elements.keys == value.elements.keys)  // TODO: this discards type data =(
+      case (x: Record, value: Record) => {
+        require(DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
         (x.elements zip value.elements) foreach { case ((_, x), (_, value)) =>
           x.poke(value)
         }
       }
       case (x: EnumType, value: EnumType) => {
-        require(x.getClass() == value.getClass())  // TODO: chisel needs to expose typeEquivalent
+        require(DataMirror.checkTypeEquivalence(x, value), s"EnumType mismatch")
         pokeBits(x, value.litValue)
       }
       case x => throw new LiteralTypeException(s"don't know how to poke $x")
@@ -66,9 +110,9 @@ package object chiseltest {
       }
       case x: Interval =>
         Context().backend.peekBits(x, stale).I(x.binaryPoint).asInstanceOf[T]
-      case (x: Bundle) => {
+      case (x: Record) => {
         val elementValueFns = x.elements.map { case (name: String, elt: Data) =>
-          (y: Bundle) => (y.elements(name), elt.peekWithStale(stale))
+          (y: Record) => (y.elements(name), elt.peekWithStale(stale))
         }.toSeq
         chiselTypeOf(x).Lit(elementValueFns: _*).asInstanceOf[T]
       }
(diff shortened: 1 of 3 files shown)


sequencer:
Not sure why this failed.
```
Testers2 thread safety checker
[info] [0.000] Elaborating design...
[info] [0.001] Done elaborating.
Computed transform order in: 65.6 ms
Total FIRRTL Compile Time: 13.3 ms
file loaded in 5.38972E-4 seconds, 4 symbols, 1 statements
test InputOnlyModule Success: 0 tests passed in 1 cycles in 0.003960 seconds 252.55 Hz
- should locate source lines for simultaneous pokes from parallel threads *** FAILED ***
  "Overlapping pokes from ListBuffer(chiseltest.package$testableData.$anonfun$poke$1$adapted(package.scala:26), chiseltest.package$testableData.$anonfun$poke$1$adapted(package.scala:26))" did not include substring "ThreadSafetyLocatorTest.scala:15" (ThreadSafetyLocatorTest.scala:23)
```

sequencer:
Hello @ducky64, would you mind give some comments on this PR?
How this could get merged :)

ducky64:
I'm very reluctant to allow partial pokes (at least without more consensus), because of the possibility for subtle errors and code bugs. In the early days of Chisel, there were many interfaces that had a focus on convenience for the immediate task at hand, rather than consistency and safety, and years later, they have come back to bite us. Worse is that we can't fix those since a change would break existing code.

So kind of the approach here is to build a consistency-first, safety-first interface - and if it turns out to be too strict, we can always expand an API later without breaking existing code (but the reverse doesn't work).

One solution might be a dedicated partial poke API - so you would define Record.partialPoke (using implicits), while leaving regular poke as strict.

The checkTypeEquivalence change looks fine, though.

sequencer:
Sure! I have added the `testableRecord.partialPoke` for Record partial poking.

ducky64:
Can you add a unit test for the partial poke?

ducky64:
Also maybe a nit, but I think I'd slightly prefer `pokePartial` as a name, unless anyone has strong preferences.

sequencer:
> Can you add a unit test for the partial poke?
> Also maybe a nit, but I think I'd slightly prefer `pokePartial` as a name, unless anyone has strong preferences.

Sure!

sequencer:
done~

src/test/scala/chiseltest/tests/TestUtils.scala
```diff
@@ -35,3 +39,16 @@ class QueueModule[T <: Data](ioType: T, entries: Int) extends MultiIOModule {
   val out = IO(Decoupled(ioType))
   out <> Queue(in, entries)
 }
+
+/** borrow from chiselTests/RecordSpec.scala */
```
edwardcwang, line 43:
```suggestion
/** Borrowed from chiselTests/RecordSpec.scala */
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
 package object chiseltest {
   import chisel3.internal.firrtl.{LitArg, ULit, SLit}
 
+  implicit class testableRecord[T <: Record](x: T) {
+    /** poke `x` signals with a [[Record.litValue()]]
```
edwardcwang, line 16:
```suggestion
    /** Poke the given signal with a [[Record.litValue()]]
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
...
+      * Literal of this Record can be instantiated with
+      * {{{
+      *   someRecord.Lit(_.elements("foo") -> 4.U)
+      * }}}
+      * `pokePartial` will only poke [[Input]] signal of `x`,
```
edwardcwang, line 21:
```suggestion
      * `pokePartial` will only poke [[Input]] signals of `x`,
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
 package object chiseltest {
   import chisel3.internal.firrtl.{LitArg, ULit, SLit}
 
+  implicit class testableRecord[T <: Record](x: T) {
+    /** poke `x` signals with a [[Record.litValue()]]
+      * Literal of this Record can be instantiated with
```
edwardcwang, line 17:
```suggestion
      * Literals of this Record can be instantiated with
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
...
+      * {{{
+      *   someRecord.Lit(_.elements("foo") -> 4.U)
+      * }}}
+      * `pokePartial` will only poke [[Input]] signal of `x`,
+      * and elements of `x` which contains no literal will be ignored.
```
edwardcwang, line 22:
```suggestion
      * and elements of `x` which contain no literal will be ignored.
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
...
+          }
+        }
+        .foreach { case (k, v) => v match {
+          case record: Record => record.pokePartial(value.elements(k).asInstanceOf[Record])
+          case data: Data => data.poke(value.elements(k))
+        }}
+    }
+
+    /** poke `x` signals with a [[Record.litValue()]]
+      * elements of `x` which contains no literal will be ignored.
```
edwardcwang, line 42:
```suggestion
      * elements of `x` which contain no literal will be ignored.
```

src/main/scala/chiseltest/package.scala
```diff
@@ -12,6 +12,51 @@ import chisel3.util._
...
+            }
+          }
+        }
+        .foreach { case (k, v) => v match {
+          case record: Record => record.pokePartial(value.elements(k).asInstanceOf[Record])
+          case data: Data => data.poke(value.elements(k))
+        }}
+    }
+
+    /** poke `x` signals with a [[Record.litValue()]]
```
edwardcwang, line 41:
```suggestion
    /** Check the given signal with a [[Record.litValue()]];
```

src/main/scala/chiseltest/package.scala
```diff
@@ -37,15 +82,15 @@ package object chiseltest {
       case (x: Interval, value: Interval) =>
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
-      case (x: Bundle, value: Bundle) => {
+      case (x: Record, value: Record) => {
         // TODO: chisel needs to expose typeEquivalent
-        require(x.elements.keys == value.elements.keys)  // TODO: this discards type data =(
+        DataMirror.checkTypeEquivalence(x, value)
```
ducky64, line 87: checkTypeEquivalence returns a boolean value, right? So this just gets a boolean and discards it, instead of asserting on it?

src/test/scala/chiseltest/tests/BasicTest.scala
```diff
@@ -30,6 +31,40 @@ class BasicTest extends FlatSpec with ChiselScalatestTester with Matchers {
...
+  it should "fail on partial expect mismatch" in {
+    val typ = new CustomBundle("foo" -> UInt(32.W), "bar" -> UInt(32.W))
+    assertThrows[exceptions.TestFailedException] {
+      test(new PassthroughModule(typ)) { c =>
+        c.in.pokePartial(typ.Lit(
+          _.elements("foo") -> 4.U
+        ))
+        c.out.expect(typ.Lit(
+          _.elements("foo") -> 4.U,
+          _.elements("bar") -> 5.U
```
ducky64, line 62: I think this actually depends on undefined initialization behavior?

It might also be worth adding a test for partial expect failure (as opposed to full expect failure)?

src/test/scala/chiseltest/tests/TestUtils.scala
```diff
@@ -35,3 +39,16 @@ class QueueModule[T <: Data](ioType: T, entries: Int) extends MultiIOModule {
...
+  val elements = ListMap(elts map { case (field, elt) =>
+    requireIsChiselType(elt)
+    field -> elt
+  }: _*)
+  def apply(elt: String): Data = elements(elt)
+  override def cloneType: this.type = {
+    val cloned = elts.map { case (n, d) => n -> DataMirror.internal.chiselTypeClone(d) }
+    (new CustomBundle(cloned: _*)).asInstanceOf[this.type]
+  }
+}
```
ducky64, line 54: missing newline at end of file

sequencer:
After I added `DataMirror.checkTypeEquivalence` to `FixedPoint` and `Interval`,
This test failed:
```
- should work with FixedPoint *** FAILED ***
  java.lang.IllegalArgumentException: requirement failed: FixedPoint type mismatch
  at scala.Predef$.require(Predef.scala:281)
  at chiseltest.package$testableData.poke(package.scala:79)
  at chiseltest.tests.ElementTest$$anon$7$$anon$8.expect(ElementTest.scala:104)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.base/java.lang.reflect.Method.invoke(Method.java:566)
  at chiseltest.tests.ElementTest.$anonfun$new$12(ElementTest.scala:111)
  at chiseltest.tests.ElementTest.$anonfun$new$12$adapted(ElementTest.scala:110)
  at chiseltest.backends.treadle.TreadleBackend.$anonfun$run$1(TreadleBackend.scala:144)
```
It means I cannot use type check on `FixedPoint`?

sequencer:
> It means I cannot use type check on FixedPoint?

Just checked the codes, that makes sense.

ducky64:
Is this a unknown width thing, or is something else about the implementation of typeEquivalent causing the error? I'm fine with using the old type checking code for FixedPoint / Interval, especially since I don't think we use typeEquivalent on integer types because of this issue.

sequencer:
Yes, so I keep the check codes of `FixedPoint` / `Interval` unchanged.

ucbjrl:
Let's see if this works as advertised. (I've edited the PR body to include build dependencies.)

ucbjrl:
I can't restart the workflow - GitHub claims it's queued, but it never starts. @sequencer could you push a trivial change to see if we can kick off a rebuild with the correct versions (master) of firrtl and chisel3?

Issue #106, opened 2019-12-22
Feature request - add caching of Verilator binaries


oharboe:
I have a test that takes ca. 10-20 minutes to prepare(Elaboration + generate FIR + Verilator + g++), but the elaboration itself takes only 42 seconds. The test itself is only 3000 cycles and quickly completed.

If Chisel could create a hash after 42 seconds(when elaboration is complete), then that could be used as a key in a cache of the Verilator binaries.

This could bring down the turnaround time for these tests from 10-20 minutes to 60 seconds.  Bringing down the turnaround time to 60 seconds would matter enormously, because there is then no need to switch tasks to try to make use of the 10-20 minute wait.

ducky64:
Interesting. I'm a bit wary about introducing external state (like the hash db) in the core, since there's more potential for failure, and when it fails it's difficult to debug. It might be an experimental / power use feature though. There would also be a not-insignificant amount of plumbing work, both to build the cache, and to refactor the Verilator backend to use it.

Side note: I'm thinking of adding an API for separating simulator compilation and test execution (to address #34), for example:
```scala
class BasicTest extends FlatSpec with ChiselScalatestTester with Matchers {
  val precompiled = test(new StaticModule(42.U))
  // or perhaps even manually creating the backend / simulator interface object

  it should "test a thing" in {
    precompiled.test { c =>
      c.out.expect(42.U)
    }
  }

  it should "test it again w/o recompiling" in {
    precompiled.test { c =>
      c.out.expect(42.U)
    }
  }
}
```
(names TBD, this could be `compile` and `run` instead of just `test`)

In theory, with the simulator interface constructor exposed, this could allow you to specify a Chisel elaborated design with a pre-existing Verilator binary (though there would still be a bit of plumbing and refactoring that needs to happen). The caching layer would be separate, but you could probably build it?

oharboe:
The above doesn't work for the use-case where one is iterating on the Scala part of a test as sbt is relaunched every time a change is made. 

For every little configuration/API niggle you add, you halve the number of people who benefit from this.

Turnaround times is an incredibly important non-functional property w.r.t. user experience and productivity.

For me it's the key "missing feature" of chisel-testers2 and one that I spend considerably energy doing something about, so whatever chisel-testers2 can do saves a ton of time for me.

W.r.t. a database to cache things, surely there's a Java/Scala database you could use, but even so I see that it's more that can go wrong.

oharboe:
[...]

PR #156, opened 2020-04-19
Move to dependencyAPI

sequencer:
In this PR. I'm going to implement a `ChiselTest` stage.

Since `chiseltester` use some informations from firrtl, eg: combinational logic, top ports.
In order to make `chiseltester` more compatible to firrtl Transforms, I will refactor `BackendExecutive` with `Stage`.

With this refactor, `chiseltester` can easily access the generated `Annotation` such as `portNames`, `pathsAsData`, and `Circuit` after transformed. Most of configuration can be set and get based on `Annotation` and `Phase` which will ease the difficulty to add a new plugin. 

## Implementation
This PR rewrite simulator backend interface and add Stages to unify with firrtl.

#### AddDefaults
This Phase add 4 default setting if not set.
1. `addDefaultBackend`: it will add `SimulatorBackendAnnotation("treadle")` by default.
2. `addDefaultBinary`: it will add `SimulatorBinary` based on `SimulatorBackendAnnotation` by default, which will directly execute this program from `PATH`, if not set by user.
3. `addWaveForm`: it will add `WaveFormAnnotation("none")` by default, `WriteVcdAnnotation` will be converted to `WaveFormAnnotation("vcd")`
4. `convertAbsoluteTargetDir` will convert `TargetDirAnnotation` to absolute path to get rid of path problem in simulator.

- If backend is vcs:
-- use `getLicense` to get vcs `LM_LICENSE_FILE` from `$PATH` is not being set explicitly,
-- If user not set `EnableCache` explicitly, `EnableCache(true)` will be added explicitly,
-- If user set `WaveFormAnnotation` to `vcd` or `fsdb`, it will be kept, otherwise set to `none`.
-- Add `SimulatorBinary` based on `SimulatorBackendAnnotation` by default, which will directly execute vcs from `PATH`, if not set by user.

- If backend is verilator:
-- if user set `WaveFormAnnotation` to `vcd`, it will be kept, otherwise set to `none`.
-- Add `SimulatorBinary` based on `SimulatorBackendAnnotation` by default, which will directly execute verilator from `PATH`, if not set by user.

- If backend is treadle:
-- convert `WaveFormAnnotation("vcd")` to `treadle.WriteVcdAnnotation`, and propagate all annotation to treadle.

#### AnalysisCircuit
- add `TopCombinationalPathAnnotation`, generate combinational paths of top IO Components.
- add `ExportedSingalsAnnotation`, a `[Data, String]` map between Scala `Data` and verilog name.

#### CompileDut
- If backend is vcs:
-- TBD

- If backend is verilator:
-- if user already defined `CommandAnnotation`, it will skip this Phase, and annotations will be directly passed to simulator executor.
-- if cache is enabled, it will generate verilator cpp harness based, and compile a dut, if cache is missed.

- If backend is treadle:
-- execute `PrepareAstFromLowFIRRTL`, and generate `TreadleTesterAnnotation` directly.

#### SimulationPhase
- run simulation itself

### cache #106
add `EnableCache` annotation to cache signal based on hash of firrtl circuit.

### remove `CommandEditor`
I believe no one is using this, since it's too tricky... we should add annotation s

### no more chirrtl
Refactor all usage to `chirrtl` to `firrtl`

### Build Dependencies


src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
```diff
@@ -29,98 +31,72 @@ object VcsExecutive extends BackendExecutive {
...
-    new VcsBackend(dut, portNames, pathsAsData, command)
+    val chiselAnnotations = (new chisel3.stage.phases.Elaborate).transform(annotationSeq :+ chisel3.stage.ChiselGeneratorAnnotation(dutGen))
+    val firrtlAnnotations = new PhaseManager(
+      Seq(Dependency[chisel3.stage.phases.MaybeAspectPhase], Dependency[chisel3.stage.phases.Convert], Dependency[firrtl.stage.FirrtlPhase]),
+      Seq(Dependency[chisel3.stage.phases.Elaborate])
+    ).transformOrder.foldLeft(chiselAnnotations :+ CompilerAnnotation(new VerilogCompiler()))((a, f) => f.transform(a))
+    val tester2Annotations = (new chiseltest.stage.VerilatorCompilerPhase).transform(firrtlAnnotations)
+    val dut = getTopModule(chiselAnnotations.collect { case x: ChiselCircuitAnnotation => x }.head.circuit).asInstanceOf[T]
+
+//    // Generate Harness
```
chick, line 47: Remove commented code

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCompilerPhase.scala
```diff
@@ -0,0 +1,258 @@
...
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.ir._
+import firrtl.options.{Phase, TargetDirAnnotation}
+import firrtl.stage.FirrtlCircuitAnnotation
+
+import scala.sys.process.{ProcessBuilder, ProcessLogger, _}
+
+case class CommandAnnotation(value: Seq[String]) extends NoTargetAnnotation
+
+class VerilatorCompilerPhase extends Phase {
```
chick, line 19: I think moving to stage/phase approach is good but this could use a general outline of what this phase does.

sequencer:
Should I keep `CommandEditor` still available in the new version?
I personally think adding functions should compose another `Phase` rather than use regex to fix the command, which seems to be a brittle way to hack low-level command.
Is no one object this, I'm going to remove it.

sequencer:
Hi, @seldridge, I noticed that Treadle doesn't honor `TargetDirAnnotation` to create a vcd file. when I inspecting codes into Treadle, I noticed that Treadle use `StageOptions.getBuildFileName` to create file, but when I set `TargetAnnotation` in `ChiselTesterStage`, `StageOptions` is not created, would you mind giving some idea to design this?

seldridge:
Treadle tests look like they're setting `TargetDirAnnotation`. I'm not sure what's happening here. My guess would be that the `TargetDirAnnotation` is either getting overwritten somewhere or stripped. You may need to publishLocal a modified version of Treadle to debug it.

I do know that `chisel-iotesters` does have a `filterAnnotations` kludge to get around the situation where multiple copies of a `TargetDirAnnotation` (and others) can get created. It's possible that something similar is going on here.

chick:
@sequencer @seldridge 
Treadle uses
```
stageOptions.getBuildFileName(topName, Some(".vcd"))
```
 to generate the VCD file name.  

TopName is generated via the `OutputFileAnnotation` not the `TargetDirAnnotation`
```
  val topName: String = annotationSeq.collectFirst { case OutputFileAnnotation(ofn) => ofn }.getOrElse(engine.ast.main)
```

sequencer:
@ekiwi-sifive sorry for the delay :( working on it this week, hope I can spare times to this.

ekiwi-sifive:
> @ekiwi-sifive sorry for the delay :( working on it this week, hope I can spare times to this.

Don't worry @sequencer . We decided in the last dev meeting that the GraphML addition can be part of a minor release, so there is no rush to get this in before the `1.4` release. Take your time.

ducky64:
Is there any movement on this mega-refactor? There's a few changes I want to make to internal APIs, that might conflict with this.

sequencer:
Oh, I think I can clean it up before next dev-meeting, hopefully this weekend. sorry for the delay!

src/main/scala/chiseltest/stage/phases/SimulationPhase.scala
```diff
@@ -0,0 +1,36 @@
...
+      *       the correct logic is:
+      *       we need to provide a api to user to let them change main clock.
+      *       and provide a analysis transform, if dut io has its own clock domain
+      * */
+    val backend = getThreadedBackend(a)
+    val testFn = getTestFunction(a)
+
+    require(context.value.isEmpty)
+    context.withValue(Some(new Instance(backend))) {
+      backend.run(testFn)
```
ducky64, line 32: Random thought: would it make sense for each backend to be a phase, extending some base backend-simulation-phase class? (it might not)

src/test/scala/chiseltest/experimental/tests/VerilatorClockPokeTest.scala
```diff
@@ -1,19 +1,18 @@
...
 import org.scalatest._
 import treadle.executable.ClockInfo
-import treadle.{ClockInfoAnnotation, WriteVcdAnnotation}
+import treadle.ClockInfoAnnotation
 import org.scalatest.flatspec.AnyFlatSpec
 
 class VerilatorClockPokeTest extends AnyFlatSpec with ChiselScalatestTester {
   behavior of "Testers2 with a clock input"
-
+  /*@todo*/
```
ducky64, line 15: Uh, what? (ditto for repeated instances below)

src/test/scala/chiseltest/tests/BasicTest.scala
```diff
@@ -26,13 +26,28 @@ class BasicTest extends AnyFlatSpec with ChiselScalatestTester with Matchers {
   }
 
   it should "fail on expect mismatch" in {
-    assertThrows[exceptions.TestFailedException] {
+    assertThrows[ExpectsException] {
```
ducky64, line 29: Not sure I agree with this change, I think there's a good reason for tests to look like ScalaTest.
Without the assertThrows, would this show up as a test error or test failure for ScalaTest? I think keeping the distinction is useful.

src/test/scala/chiseltest/tests/FaultLocatorTest.scala
```diff
@@ -41,13 +54,12 @@ class FaultLocatorTest extends AnyFlatSpec with ChiselScalatestTester with Match
         c.out.expectDequeueNow(true.B)
       }
     }
-    // Only check the filename to avoid this being too brittle as implementation changes
-    exc.failedCodeFileNameAndLineNumberString.get should startWith ("DecoupledDriver.scala:")
-    exc.getMessage should include regex ("""\(lines in FaultLocatorTest\.scala:[^\)]*41.*\)""")
+      // Only check the filename to avoid this being too brittle as implementation changes
+      .expects.head.trace.toString should include regex ".*(DecoupledDriver.scala:\\d+)"
```
ducky64, line 58: Style: I don't like the comment inbetween the object and the call. Similar below.

I'm not sure if we just want to check the name, since we want to make sure it's locating the right part of that file.

But if you only wanted to check the name, you could make that clear in the test name / description, instead of a comment.

src/test/scala/chiseltest/tests/NoScalatestTesterTest.scala
```diff
@@ -19,7 +20,9 @@ class NoScalatestTesterTest extends AnyFreeSpec with Matchers {
       out.expect(value)
     }
 
-    test(new ShifterModule(UInt(8.W), 4)) { c =>
+    test(new ShifterModule(UInt(8.W), 4),
+      Seq(TargetDirAnnotation("test_run_dir" + java.io.File.separator + sanitizeFileName(this.toString)))
```
ducky64, line 24: Why is this necessary?

src/test/scala/chiseltest/tests/OptionsPassingTest.scala
```diff
@@ -44,6 +43,7 @@ class OptionsPassingTest extends AnyFlatSpec with ChiselScalatestTester with Mat
     vcdFileOpt.get.delete()
   }
 
+  /*@todo*/
```
ducky64, line 46: What? (similarly below)

src/test/scala/chiseltest/tests/ThreadSafetyLocatorTest.scala
```diff
@@ -10,6 +10,7 @@ import org.scalatest.flatspec.AnyFlatSpec
 class ThreadSafetyLocatorTest extends AnyFlatSpec with ChiselScalatestTester {
   behavior of "Testers2 thread safety checker"
 
+  /*@todo*/
```
ducky64, line 13: What?

ducky64 (changes requested):
Architecturally (solely for Dependency API), this seems fine. But I don't really like that this is several PRs in one. I suggest it be split into these:
- migration to Dependency API
- scalafmt support
- dependent library updates
- mill support
- backend refactoring (as much as could be split from dependency API)
- changing ScalaTest Exceptions to ExpectsException. I also think this needs further discussion.
- cache support for VCS
- removal of legacy path for Verilator / VCS backends
- removal of ScalaTest writeVcd flag (I strongly disagree with this removal, but it's a change that is part of this PR nonetheless)

Looking at the PR as a whole, I see these recurring issues:
- some APIs seem not-well-defined/standardized. For example, there is backend-specific code in the AnalysisCircuit phase - a clear violation of separation of concerns.
- this appears to break existing APIs? eg, getSourceSlocks, getSinkClocks, TesterOptions removed. Even if it's experimental, I think this needs discussion (and ideally a deprecation phase, unless you know for sure no one is using it).

When the code is ready for a thorough review, I'd like feedback from the people most familiar with these features:
- how to best deal with annotation transforms, in particular the AddDefaults phase where the potential to duplicate annotations smells weird to me. @seldridge , @chick ?
- a look over the phases code - it looks sane to me, but I haven't dealt with phases at all. @seldridge ?

*Overall, I don't feel comfortable merging this mega-PR as-is (read: will not approve for merge in its current state)*, because it mixes in so many things and I can't rule out unintentional side effects with any degree of confidence. I'd like orthogonal features to be broken apart, so the diffs can be reviewed individually. Or an explanation of why certain features are intertwined (but I'd find it hard to believe that everything here is non-separable).

I get the feeling that this branch didn't start life as a PR, but as your personal work-in-progress-and-customize-things-to-your-liking-branch. Which is fine, but it really isn't conducive to the upstreaming and code review process. In the future, I'd recommend PRing things as you make changes, to avoid this kind of mega-PR that you'll have to split out and clean up. General good practice is to PR often (and PRs here should get reviewed relatively quickly) and keep in sync with master as much as possible. If you're comfortable with advanced git, you can also branch off branches to have a local version that integrates several orthogonal features, though it's not the easiest or cleanest thing to do.

build.sc
```diff
@@ -0,0 +1,98 @@
+import mill._
```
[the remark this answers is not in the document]
edwardcwang: One comment I have (about this in particular) is that it often makes sense to have build system changes live with code changes, especially if he is using these build systems to build the updated code. A small nit with doing this separately is that the mill PR can easily get out of date.

sequencer:
Thanks @ducky64, I'm splitting this PR into different PRs. and rework this huge PR.
1. mill support.
2. scalafmt support.(It may introduce a huge line change, but good for a standardization for PRs.)
3. migration to Dependency API.
4. backend refactoring (as much as could be split from dependency API) and removal of legacy path for Verilator / VCS backends
5. cache support.
6. changing ScalaTest Exceptions to ExpectsException. I also think this needs further discussion.

1 and 2 are separate.
4 depends on 3, and 5 depends on 4.
6 will be in the feature PRs.

> some APIs seem not-well-defined/standardized. For example, there is backend-specific code in the AnalysisCircuit phase - a clear violation of separation of concerns.

I'll add a standard API and clarifications.

> this appears to break existing APIs? eg, getSourceSlocks, getSinkClocks, TesterOptions removed. Even if it's experimental, I think this needs discussion (and ideally a deprecation phase, unless you know for sure no one is using it).

Sorry about this, it seems a long time a ago, which I just removed, I'll try add clock analysis pass for these features. I'll add back those mis-removed features.

> how to best deal with annotation transforms, in particular the AddDefaults phase where the potential to duplicate annotations smells weird to me.

It contains some directory track which has been discussed in the last dev meeting, but it is still under discussion.

> Overall, I don't feel comfortable merging this mega-PR as-is (read: will not approve for merge in its current state), because it mixes in so many things and I can't rule out unintentional side effects with any degree of confidence

Sure, I think this PR is too mess to be reviewed, I'll rework on this.

PR #168, opened 2020-06-01
DecoupledDriver: support user to specify the error message format

ljwljwljwljw:
It's very useful to allow the user to specify the error message format such as print the expected value and the error value in hexadecimal (by default the value was printed in decimal).

For example, if I want to print the error value in hexadecimal, I can use a msgGen function which maps a value `x` to its hexadecimal string:

`x => Some(x.litValue().toString(16))`

Use the following test case:

```
 it should "pass through elements, using enqueueSeq and dequeueSeq, with custom error message" in {
    test(new QueueModule(UInt(8.W), 2)) { c =>
      c.in.initSource().setSourceClock(c.clock)
      c.out.initSink().setSinkClock(c.clock)

      val seq = Seq(42.U, 43.U, 44.U)
      val errSeq = seq map (x => (x.litValue() + 1).U)

      fork {
        c.in.enqueueSeq(seq)
      }.fork {
        // print error value and expected value in hex
        c.out.expectDequeueSeq(errSeq, x => Some(x.litValue().toString(16)))
      }.join()
    }
  }
```
The error message would be:
```
out_bits=42 did not equal expected=43: Some(
Expected:2b
Actual:2a
) (lines in QueueTest.scala: 76, 65, 76)
ScalaTestFailureLocation: chiseltest.DecoupledDriver at (DecoupledDriver.scala:83)
```

### DIFF
--- a/src/main/scala/chiseltest/DecoupledDriver.scala
+++ b/src/main/scala/chiseltest/DecoupledDriver.scala
@@ -74,35 +74,35 @@ class DecoupledDriver[T <: Data](x: ReadyValidIO[T]) {
     }
   }
 
-  def expectDequeue(data: T): Unit = timescope {
+  def expectDequeue(data: T, msgGen: T => Option[String] = _ => None): Unit = timescope {
     // TODO: check for init
     x.ready.poke(true.B)
     fork.withRegion(Monitor) {
       waitForValid()
       x.valid.expect(true.B)
-      x.bits.expect(data)
+      x.bits.expect(data, msgGen)
     }.joinAndStep(getSinkClock)
   }
 
-  def expectDequeueNow(data: T): Unit = timescope {
+  def expectDequeueNow(data: T, msgGen: T => Option[String] = _ => None): Unit = timescope {
     // TODO: check for init
     x.ready.poke(true.B)
     fork.withRegion(Monitor) {
       x.valid.expect(true.B)
-      x.bits.expect(data)
+      x.bits.expect(data, msgGen)
     }.joinAndStep(getSinkClock)
   }
 
-  def expectDequeueSeq(data: Seq[T]): Unit = timescope {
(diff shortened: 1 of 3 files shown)


ducky64:
So one potential issue I have with this is that it doesn't really compose well - for example, were one to build a higher-level transaction driver on top of (that calls) dequeueSeq, it would have these problems:
1. boilerplate to thread through the message handler through function calls - a very imperfect programming analogy would be explicit return codes, and the need to thread them through functions, as opposed to modern exception handling
2. likely an either-or to take in a message handler from a upper level, or provide a 'hard-coded' pre-specified error message, but not both (which would require merging logic, further complicated by the use of functions rather than strings)
3. potential code bloat because of the need to write custom error messages - possibly (probably?) scattered and inconsistent ones, since I'd imagine that custom errors would be (primarily?) added ad-hoc when debugging a particular error, or (more rarely?) in infrastructure where there was a lot of foresight

As an alternative solution for your case: would it make sense to print values in both decimal and hex. since these are probably the two representations that are the most common? This might limit flexibility in some instances, but that might also be an indication to use more appropriate data structures like bundles and enums (which, if they aren't being printed in a sane way, should be changed to)?

Separate from printing the different values is providing failure context, eg this decoupled-expect failed while in a xyz-link transaction. While out of scope of your current PR, one proposal to solve that would be to attach a short description to timescopes, and have the failure print out the timescope stack.

Do these alternative proposals address your use cases? I'd like to have a clean solution, if it works.

ljwljwljwljw:
Print values in both decimal and hex will solve my problem well, and this solution will work well not only in 'expectDequeue' but also every 'expect', but I don't know if there are more flexible requirements for printing error messages?

ducky64:
So what I'm hoping to do is to provide more structured (and built-in) mechanisms that compose well, instead of quick-fixes that don't provide good abstraction and possibly cause code bloat. The proposed solution probably won't address all use cases, but we can consider clean solutions for more use cases, as they come up.

PR #170, opened 2020-06-08
Display hex and type-specific values on expect failures

ducky64:
For expect failures, the error now includes the hex value, decimal value, and a type-specific decoding (where applicable).
For example:
- UInt case: `out=42 (0x2a) did not equal expected=0 (0x0)`
- Bool case: `out=true (1, 0x1) did not equal expected=false (0, 0x0)`
- Fixed case: `out=-2.0 (-4, -0x4) did not equal expected=-0.5 (-1, -0x1)`
- Enum case: `out=[unimplemented enum decode] (0, 0x0) did not equal expected=[unimplemented enum decode] (1, 0x1)` (@hngenc, I think StrongEnum needs some kind of reflection API to get the name from a integer? - related: #111)

Architecturally, this passes a decode function through the internals of `wire.expect` -> `BackendInterface.expectBits` -> `TestEnvInterface.testerExpect`. Does not expose any additional APIs. Bundles aren't affected, since the specific sub-field should fail.

Addresses the underlying issue from #168 with infrastructural support.

### DIFF
--- a/src/main/scala/chiseltest/internal/TestEnvInterface.scala
+++ b/src/main/scala/chiseltest/internal/TestEnvInterface.scala
@@ -59,7 +59,8 @@ trait TestEnvInterface {
   /** Expect a specific value on a wire, calling testerFail if the expectation isn't met.
     * Failures queued until the next checkpoint.
     */
-  def testerExpect(expected: Any, actual: Any, signal: String, msg: Option[String]): Unit = {
+  def testerExpect(expected: BigInt, actual: BigInt, signal: String,
+                   msg: Option[String], decode: Option[BigInt => String]): Unit = {
     if (expected != actual) {
       val appendMsg = msg match {
         case Some(_) => s": $msg"
@@ -75,7 +76,14 @@ trait TestEnvInterface {
       val trimmedTrace = trace.getStackTrace.drop(expectStackDepth + 2)
       val detailedTrace = topFileName.map(getExpectDetailedTrace(trimmedTrace.toSeq, _)).getOrElse("")
 
-      val message = s"$signal=$actual did not equal expected=$expected$appendMsg$detailedTrace"
+      val (actualStr, expectedStr) = decode match {
+        case Some(decode) =>
+          (f"${decode(actual)} ($actual, 0x$actual%x)", s"${decode(expected)} ($expected, 0x$actual%x)")
+        case None =>
+          (f"$actual (0x$actual%x)", s"$expected (0x$actual%x)")
+      }
+
+      val message = s"$signal=$actualStr did not equal expected=$expectedStr$appendMsg$detailedTrace"
       val stackIndex = expectStackDepth + 1
       batchedFailures += new FailedExpectException(message, stackIndex)
     }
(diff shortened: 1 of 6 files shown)


src/main/scala/chiseltest/internal/TestEnvInterface.scala
```diff
@@ -56,10 +56,19 @@ trait TestEnvInterface {
     }
   }
 
+  protected def bigintToHex(x: BigInt): String = {
```
chick, line 59: Note that some viewers like GtkWave will pad leading zeros on hex strings to represent their widths.

src/main/scala/chiseltest/package.scala
```diff
@@ -57,6 +57,29 @@ package object chiseltest {
...
+            val bpInteger = 1 << binaryPoint
+            (bits.toFloat / bpInteger).toString
+          case UnknownBinaryPoint => "[unknown binary point]"
+        }
+      }
+      inner
+    }
+
+    def enumToString(record: EnumType): BigInt => String = {
+      def inner(bits: BigInt): String = "[unimplemented enum decode]"
```
chick, line 78: ChiselGUI gets the names of enum values, see EnumManager in that repo, it seems to be picking them up 
by collecting the `EnumDefAnnotation` and the EnumComponentAnnotation`

src/main/scala/chiseltest/internal/TestEnvInterface.scala
ducky64: Is there a consensus on whether this should be done or not, especially within the Chisel ecosystem?

src/main/scala/chiseltest/package.scala
ducky64: Good to know, but I think (unless there's like overwhelming demand for this) I'll wait for underlying enum reflection-like support in base Chisel, instead of adding more complexity while only addressing part of the problem (I don't think enum annotations can address #111, at least not without more hacks?)

src/main/scala/chiseltest/internal/TestEnvInterface.scala
aswaterman: Quantizing widths to multiples of the log of the radix isn't ideal, which IIRC was part of the rationale for moving away from doing things like inferring that `"h1".U` has a width of 4.

ducky64: Yeah, I can see that it would be a deal-breaker for specifying a literal, but it seems less clear that it's an obvious bad idea to pad zeros based on width for display purposes? That leading zero might represent more bits than there actually are, but I'm not sure if that edge case is worth it (or not?) to have an (imperfect) indicator of wire width?

aswaterman: It's worth thinking about the implications for signed numbers.  If the width is inferred to be greater than actual, but the value isn't sign-extended to the new width, it'll appear to be a positive number.

ducky64: Ok, leaving as-is (not padded), if anyone really cares / has more information on general norms, we can revisit this later. Feel free to open an issue.

Some alternate options:
- Don't zero-pad (currently the case)
- Zero-pad hex only
- Zero-pad both hex and decimal
- Some other type of width indicator, perhaps displaying the data type and width, instead of leading zeros

Issue #177, opened 2020-07-20
Verilated::flushCall renamed in newer versions of verilator


jiegec:
Code change in verilator: https://github.com/verilator/verilator/commit/fac89c5d6233f900d0e382bde21526400427e388#diff-6c80b812607617c162e46fd56e6579c8

Related code in chisel-testers2:
https://github.com/ucb-bar/chisel-testers2/blob/db85e6fcb5ce2d8397ea0d9ed1745cdd5b7b9e35/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala#L148

ljwljwljwljw:
I have same problem, maybe "Verilated::flushCall()" need to be changed to "Verilated::runFlushCallbacks()"?

ducky64:
I think it makes sense for us to keep our Verilator version in sync with the [version recommended by chisel3's setup](https://github.com/freechipsproject/chisel3/blob/master/SETUP.md), which is currently 4.016.

Though, feel free to make a PR if you have code that updates behavior for newer versions while remaining 4.016 compatibility?

ljwljwljwljw:
> I think it makes sense for us to keep our Verilator version in sync with the [version recommended by chisel3's setup](https://github.com/freechipsproject/chisel3/blob/master/SETUP.md), which is currently 4.016.
> 
> Though, feel free to make a PR if you have code that updates behavior for newer versions while remaining 4.016 compatibility?

Thanks for you advice, here(#187) is my PR, I just simply use `verilator --version` to check verilator's version as a temporary solution, I noticed that in #156 we can use our own cpp harness or `SimulatorBinary`, after that PR this problem will be well solved.

PR #187, opened 2020-08-08
Verilator: use newer flush call back api to support newer version of verilator

ljwljwljwljw:
Use newer flush call back api in newer version of verilator to avoid the compile error (#177).

### DIFF
--- a/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
+++ b/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
@@ -1,6 +1,7 @@
 package chiseltest.legacy.backends.verilator
 
 import chisel3.MultiIOModule
+import scala.sys.process._
 
 /**
   * Generates the Module specific verilator harness cpp file for verilator compilation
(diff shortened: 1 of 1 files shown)


nrother:
This change caused some issues for me: My Verilator (default from Ubuntu 18.04) reports its self as
```
$ verilator --version
Verilator 3.916 2017-11-25 rev verilator_3_914-65-g0478dbd
```
The extracted version is `verilator_3_914-65-g0478dbd` which is considered greater than `v4.038`. Due to this, the new code is generated, which of course does not work in the old Verilator version.

ducky64:
Can you upgrade to 4.038 (or 4.x in general)?

Also, Chisel3's [officially supported version](https://github.com/freechipsproject/chisel3/blob/master/SETUP.md#ubuntu-linux) is v4.016. Is the rest of the Chisel flow still working on a Verilator so old, or are you using some other flow?

We can definitely fix this bug in this code, but I'm curious as to the underlying use case. I think the fix would be to use the first element (eg, `3.916` or `4.038` - which seems more consistent between versions). Was there a reason we were trying to parse the more complex version string?

nrother:
I used the Verilator version that comes with my Ubuntu version, "because it just works" :) I use the default flow and haven't encountered any other problems yet.

If Verilator 3.x is officially unsupported, this can probably left as "wontfix". Our institute will upgrade to Ubuntu 20.04 "soon", which comes with Verilator 4.028, so this issue shouldn't be around long anymore.

nrother:
Ok, i just notices another problem with this: The version [in the docker image I use for CI](https://hub.docker.com/r/zmar/sbt-verilator) identifies itself as 
```
Verilator 4.012 2019-03-23 rev UNKNOWN_REV
```
So parsing the second component instead of the last might be a good idea anyway.

ducky64:
@nrother can you see if #218 fixes your issue?

nrother:
Yes, the code from #218 detects the correct version in all my use-case :tada:

I still have a little bit of trouble due to https://github.com/verilator/verilator/issues/2067: The `verilator` executable is missing a shebang line, therefore it cannot be executed without a shell. Ubuntu seems to fix this before packaging, but in my docker container this is still a problem. This is not a huge issue though, because it can be fixed with a bit of `sed` magic :)

FWIW: `sed -i '1,3c#!/usr/bin/perl' /usr/local/bin/verilator`

PR #190, opened 2020-08-14
Scalatest 3.2.0

nullobject:
This PR updates scalatest to 3.2.0.

Most of the work was done automatically by scala-steward, although there were a few tests I had to fix manually.

The fixes were mostly to do with checking the output from exceptions raised by scalatest, which had changed slightly since 3.2.0.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,8 @@ resolvers ++= Seq(
 )
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.2.0",
+  "org.scalatest" %% "scalatest" % "3.0.8",
+  "org.scalatest" %% "scalatest" % "3.2.0" % "test",
   "com.lihaoyi" %% "utest" % "latest.integration"
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ resolvers ++= Seq(
 )
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.0.8",
+  "org.scalatest" %% "scalatest" % "3.2.0",
   "com.lihaoyi" %% "utest" % "latest.integration"
 )
 
--- a/src/test/scala/chiseltest/experimental/tests/NegativeInputValuesTest.scala
+++ b/src/test/scala/chiseltest/experimental/tests/NegativeInputValuesTest.scala
@@ -20,7 +20,7 @@ import chisel3._
 import chiseltest._
 import chiseltest.experimental.TestOptionBuilder._
 import chiseltest.internal.VerilatorBackendAnnotation
-import org.scalatest.FreeSpec
+import org.scalatest.freespec.AnyFreeSpec
 
 class HasOddWidthSInt extends MultiIOModule {
   val in = IO(Input(SInt(15.W)))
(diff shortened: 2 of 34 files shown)


ducky64:
The code looks fine.

Are there any implications about compatibility with older versions of ScalaTest? I imagine this isn't a concern for most users of ScalaTest who are upgrading, since they are end-users, but since this is a library that provides tools that interact with ScalaTest, things might get weird? Put another way, will users who want to upgrade chiseltest past this PR also need to upgrade their ScalaTest to latest? And if so, are there lightweight mechanisms for backwards compatibility?

ducky64:
Also: if there is agreement that this should be merged (likely if there are no backward compatibility issues, or consensus on a non-backward-compatible upgrade possibly aligned with a major release like Chisel 3.4), these documentation items also need to be done:
- Update examples in README to reflect the scalatest changes.
- Possibly a migration guide for the Scalatest changes, either as a standalone document or a README section.

build.sbt
```diff
@@ -41,7 +41,7 @@ resolvers ++= Seq(
 )
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.0.8",
+  "org.scalatest" %% "scalatest" % "3.2.1",
```
jackkoenig, line 44:
```suggestion
  "org.scalatest" %% "scalatest" % "3.2.0",
```

I don't know if it's relevant here but we had to roll back to `3.2.0` on FIRRTL: https://github.com/freechipsproject/firrtl/pull/1845

nullobject: No probs, I'll revert the commit to upgrade to 3.2.1.

nullobject: Done

nullobject:
> Put another way, will users who want to upgrade chiseltest past this PR also need to upgrade their ScalaTest to latest? And if so, are there lightweight mechanisms for backwards compatibility?

Good point. I'm not too sure. I guess the best way to tell is for me to try it and see. Will report back...

build.sbt
```diff
@@ -41,7 +41,7 @@ resolvers ++= Seq(
 )
 
 libraryDependencies ++= Seq(
-  "org.scalatest" %% "scalatest" % "3.0.8",
+  "org.scalatest" %% "scalatest" % "3.2.0",
```
ducky64, line 44:
```suggestion
  "org.scalatest" %% "scalatest" % "3.0.8",
  "org.scalatest" %% "scalatest" % "3.2.0" % "test",
```

This should allow the ChiselTest infrastructure to be used with projects requiring 3.0.8 (or later), but having the ChiselTest tests run on ScalaTest 3.2.0. Projects with a dependency on ChiselTest will import ScalaTest 3.0.8 by default (I think? - at least compiles with 3.0.8-style code without deprecation warnings), but can use a later version of ScalaTest by adding a later versioned dependency.

ducky64: FYI, here are the results of testing with a project written for ScalaTest 3.0.8 and depending on chiseltest, with chiseltest published locally with the above modification.

---

build.sbt contains:
```scala
libraryDependencies += "edu.berkeley.cs" %% "chisel3" % "3.4-SNAPSHOT"
libraryDependencies += "edu.berkeley.cs" %% "chiseltest" % "0.3-SNAPSHOT"
```
compiles fine, with no ScalaTest related warnings

---

build.sbt contains:
```scala
libraryDependencies += "edu.berkeley.cs" %% "chisel3" % "3.4-SNAPSHOT"
libraryDependencies += "org.scalatest" %% "scalatest" % "3.1.0" % "test"
libraryDependencies += "edu.berkeley.cs" %% "chiseltest" % "0.3-SNAPSHOT"
```
produces deprecated warnings from ScalaTest

---

build.sbt contains:
```scala
libraryDependencies += "edu.berkeley.cs" %% "chisel3" % "3.4-SNAPSHOT"
libraryDependencies += "org.scalatest" %% "scalatest" % "3.2.0" % "test"
libraryDependencies += "edu.berkeley.cs" %% "chiseltest" % "0.3-SNAPSHOT"
```
produces ScalaTest name related compile time errors

ducky64:
@nullobject how does the suggested change look, and does it also work for you? If it's fine, we can get this merged. If you'd like, I can also make the readme changes.

nullobject:
@ducky64 Sounds like a good solution, please make the changes and merge. I’m away on a camping trip so have intermittent internet access 🏕

ducky64:
Thanks for keeping our ScalaTest up to date! Merging now...

PR #195, opened 2020-09-18
Fix legacy support for ExecutionOptions

jackkoenig:
CompilerAnnotation has been deprecated and is now rarely used in FIRRTL,
replaced with RunFirrtlTransformAnnotation of an Emitter. The
(deprecated) support for ExecutionOptions now handles this change.

https://github.com/freechipsproject/firrtl/pull/1870 necessitates this change. Concretely, `FirrtlExecutionOptions` no longer create `CompilerAnnotation`, instead they create `RunFirrtlTransformAnnotation` with an `Emitter`.


ducky64 (approved):
lgtm

jackkoenig:
I don’t have write access but @chick this fixes the release issue

ducky64:
Oh interesting. Will merge then.

Issue #65, opened 2019-09-16
Test On All Backends


seldridge:
- All tests should be run on all backends
- Backends supported should be (open to debate):
  - Verilator
  - Treadle
  - Icarus Verilog (once supported)

VCS testing also has to be figured out.

edwardcwang:
If we can support Hammer simulation as a backend that would also be great...

chick:
@edwardcwang can you describe what that would mean?

edwardcwang:
Hammer has a simulation API/abstraction that allows a number of different simulator backends to be driven, such as VCS, ncsim, etc.

ekiwi:
We have much better coverage of different backends in the tests now.

PR #196, opened 2020-09-20
tests: Alu: add example of reusing the same test for different data

parzival3:
This pull request adds a simple example of how scalatest enables the possibility to "share" the same test for different data.
The ALU example was taken from [chisel-examples](https://github.com/schoeberl/chisel-examples/blob/master/src/main/scala/simple/Alu.scala)
[Here](https://www.scalatest.org/user_guide/sharing_tests) is the official documentation from scaltest.

Test output
```
[info] AluTest:
[info] ALU
[info] - should + 1, 2 and the result == 3
[info] - should - 1, 2 and the result == 15
[info] - should | 1, 2 and the result == 3
[info] - should & 1, 2 and the result == 0
[info] - should + 3, 4 and the result == 7
[info] - should - 3, 4 and the result == 15
[info] - should | 3, 4 and the result == 7
[info] - should & 3, 4 and the result == 0
[info] - should + 4, 5 and the result == 9
[info] - should - 4, 5 and the result == 15
[info] - should | 4, 5 and the result == 5
[info] - should & 4, 5 and the result == 4
[info] ScalaTest
[info] Run completed in 7 seconds, 946 milliseconds.
[info] Total number of tests run: 12
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.

```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------
### Why I think this is a good addition for the examples in chisel-tester2.

This test by itself is nothing interesting, but is the first step for achieving CRV(Constrained Random Verification) in Chisel.
For example, we can obtain the same behaviour described in  [ASIC WORLD Random Constraints Tutorial - part 6.](http://www.asic-world.com/systemverilog/random_constraint6.html) by combining a CSP solver like [Choco Solver](https://github.com/chocoteam/choco-solver) and Tester2 to test a set of data.

by adding
```
 "org.choco-solver" % "choco-solver" % "4.10.2"
```
to the library dependency in `build.sbt` and changing the code in `AluTest.scala`

```scala
class AluTest extends AnyFlatSpec with AluBehavior with ChiselScalatestTester with Matchers {
  behavior of "ALU"

  val model = new Model("Constraint problem")
  val bit_domain = (0x0 to 0xFF).toArray
  val src_port_domain = (0x0 to 0x0A) ++ Set(0x14, 0x18) toArray
  val src_dest_domain = (0x0 to 0xFF).toSet.diff(0x04 to 0xFF toSet).toArray

  val src_port = model.intVar("src_port", bit_domain)
  val src_dest = model.intVar("src_dest", bit_domain)

  model.member(src_port, src_port_domain).post()
  model.member(src_dest, src_dest_domain).post()
  val solver = model.getSolver
  val sol_port = ArrayBuffer[Int]()
  val sol_dest = ArrayBuffer[Int]()

  while(solver.solve()) {
    it should behave like testAddition(src_port.getValue, src_dest.getValue, 10)
    it should behave like testSubtraction(src_port.getValue, src_dest.getValue, 10)
    it should behave like testOr(src_port.getValue, src_dest.getValue, 10)
    it should behave like testAnd(src_port.getValue, src_dest.getValue, 10)
  }
}
```

We obtain the same result shown in [ASIC WORLD Random Constraints Tutorial - part 6.](http://www.asic-world.com/systemverilog/random_constraint6.html) without code duplication.

The interface to Choco solver is not very friendly and the idea is to create a wrapper for it or to write a simple CSP from scratch in scala similar to the one used in [cocotb-coverage](https://github.com/mciepluc/cocotb-coverage).

Keep up with the good work :smiley: 
Cheers,
Enrico

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ See the test cases for examples:
     ```scala
     import chisel3.experimental.BundleLiterals._
     ```
+- [AlutTest](src/test/scala/chiseltest/tests/AluTest.scala) shows an example of re-using the same test for different data
 - [ShiftRegisterTest](src/test/scala/chiseltest/tests/ShiftRegisterTest.scala) shows an example of using fork/join to define a test helper function, where multiple invocations of it are pipelined using `fork`.
 - [VerilatorBasicTests](src/test/scala/chiseltest/experimental/tests/VerilatorBasicTests.scala) shows an example using Verilator as the simulator.
   - Note: the simulator is selected by passing an annotation into the `test` function, which requires experimental imports:
--- a/src/test/scala/chiseltest/tests/AluTest.scala
+++ b/src/test/scala/chiseltest/tests/AluTest.scala
@@ -1,0 +1,85 @@
+package chiseltest.tests
+
+import chisel3._
+import chiseltest._
+import chiseltest.ChiselScalatestTester
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+/** Example of test that is "shared by multiple fixture objects
+  * More information on https://www.scalatest.org/user_guide/sharing_tests
+  */
+
+trait AluBehavior {
+  this: AnyFlatSpec with ChiselScalatestTester =>
+
+  def mask(s: Int): Int = (1 << s) - 1
+
+  def testAddition(a: Int, b: Int, s: Int): Unit = {
+    val result = (a + b) & mask(s)
+    it should s"+ $a, $b and the result == $result" in {
+      test(new Alu(s)) { c =>
+        c.io.fn.poke(0.U)
+        c.io.a.poke(a.U(s.W))
+        c.io.b.poke(b.U(s.W))
+        c.clock.step()
+        c.io.result.expect(result.U(s.W))
+      }
+    }
+  }
+
+  def testOr(a: Int, b: Int, s: Int): Unit = {
+    val result = (a | b) & mask(s)
+    it should s"| $a, $b and the result == $result" in {
+      test(new Alu(s)) { c =>
+        c.io.fn.poke(2.U)
+        c.io.a.poke(a.U(s.W))
+        c.io.b.poke(b.U(s.W))
+        c.clock.step()
+        c.io.result.expect(result.U(s.W))
+      }
+    }
+  }
+
+  def testAnd(a: Int, b: Int, s: Int): Unit = {
+    val result = (a & b) & mask(s)
+    it should s"& $a, $b and the result == $result" in {
+      test(new Alu(s)) { c =>
+        c.io.fn.poke(3.U)
+        c.io.a.poke(a.U(s.W))
+        c.io.b.poke(b.U(s.W))
+        c.clock.step()
+        c.io.result.expect(result.U(s.W))
+      }
+    }
+  }
+
+  def testSubtraction(a: Int, b: Int, s: Int): Unit = {
+    val result = (a - b) & mask(s)
+    it should s"- $a, $b and the result == $result" in {
+      test(new Alu(s)) { c =>
+        c.io.fn.poke(1.U)
+        c.io.a.poke(a.U(s.W))
+        c.io.b.poke(b.U(s.W))
+        c.clock.step()
+        c.io.result.expect(result.U(s.W))
+      }
+    }
+  }
+}
+
+
+class AluTest extends AnyFlatSpec with AluBehavior with ChiselScalatestTester with Matchers {
+  behavior of "ALU"
+  val testData: List[(Int, Int)] = List[(Int, Int)](
+    (1, 2),
+    (3, 4),
+    (4, 5)
+  )
+  testData.foreach { data =>
+    it should behave like testAddition(data._1, data._2, 4)
+    it should behave like testSubtraction(data._1, data._2, 4)
+    it should behave like testOr(data._1, data._2, 4)
+    it should behave like testAnd(data._1, data._2, 4)
+  }
+}
--- a/src/test/scala/chiseltest/tests/TestUtils.scala
+++ b/src/test/scala/chiseltest/tests/TestUtils.scala
@@ -54,3 +54,23 @@ final class CustomBundle(elts: (String, Data)*) extends Record {
     (new CustomBundle(cloned: _*)).asInstanceOf[this.type]
   }
 }
+/** taken from https://github.com/schoeberl/chisel-examples/blob/master/src/main/scala/simple/Alu.scala */
+class Alu(size: Int) extends Module {
+  val io = IO(new Bundle {
+    val fn = Input(UInt(2.W))
+    val a = Input(UInt(size.W))
+    val b = Input(UInt(size.W))
+    val result = Output(UInt(size.W))
+  })
+
+  val result = Wire(UInt(size.W))
+  result := 0.U
+
+  switch(io.fn) {
+    is(0.U) { result := io.a + io.b }
+    is(1.U) { result := io.a - io.b }
+    is(2.U) { result := io.a | io.b }
+    is(3.U) { result := io.a & io.b }
+  }
+  io.result := result
+}


ducky64 (commented):
So the linked ScalaTest documentation (https://www.scalatest.org/user_guide/sharing_tests) is more on sharing test fixtures, instead of test data. Would this application be better done with ScalaCheck? For example: https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25

Also, practical considerations:
- Does this produce a test case (at least for ScalaTest's list of test cases run at the end) for every data input? When the number of cases is small, this probably makes sense, but if not it might just get spammy.
- Spinning up a new test for each case seems a bit wasteful / inefficient. For more complicated examples, you might want to reset the DUT, but I think you generally won't need to re-elaborate or re-compile RTL? Related: #106 (possibly coming "soon"TM)
- Would the structure in the ScalaCheck example (https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25) be more suitable? I'd argue the example is much more straightforward and achieves pretty much the same thing. Complexity aside, the main differences are whether you get a separate test cases listed in ScalaTest, and whether you recompile the DUT every time.

parzival3:
Hi @ducky64 , and thank you for your comments :smiley: 

1.

> `So the linked ScalaTest documentation (https://www.scalatest.org/user_guide/sharing_tests) is more on sharing test fixtures,`

Regarding this point, I think the documentation of scaltest is a bit confusing.  But in general, my understanding is that you have a test, for example, `testAddition`(in my code) and you want to provide to it different data or fixture. So you "share" the test between the fixtures. Citing the documentation in https://www.scalatest.org/user_guide/sharing_tests
> In other words, you may want to write tests that are "shared" by different fixture objects

2)

> Spinning up a new test for each case seems a bit wasteful / inefficient. For more complicated examples, you might want to reset the DUT, but I think you generally won't need to re-elaborate or re-compile RTL? Related: #106 (possibly coming "soon"TM)

Yes, you are absolutely right, speaning up a new test is expensive when you have to re-elaborate the design for each test. While I was looking at the codebase, I saw the various comments relate to a possible caching of Verilator binaries [here](https://github.com/ucb-bar/chisel-testers2/blob/master/src/main/scala/chiseltest/ChiselScalatestTester.scala#L37). But I think that this is generally good and not only for "sharing tests".

3)

> Would the structure in the ScalaCheck example (https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25) be more suitable? I'd argue the example is much more straightforward and achieves pretty much the same thing. Complexity aside, the main differences are whether you get a separate test case listed in ScalaTest and whether you recompile the DUT every time.

Why do I think ScalaCheck is not the best fit here?
IMHO why scalatest tries to achieve is to follow the DRY principle. When you are "sharing" test like in my example you are specifying a behaviour, and from that behaviour, you can run one specific test, multiple test etc. With the example that you provided 
https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25 you have two tests that do the same thing, one test for checking only one value and the second one to generate a set of random values and validate the GCD on it. Plus you don't have an idea on which value the test fail in case of failure. 

Again, this is my humble opinion, I'm not an expert in verification and testing. I just thought that this simple example could help other people that are using chisel-tester2.

Thanks again for your time :smiley: .
Cheers,
Enrico

ducky64:
So particularly for the linked example [ccc18-testing-intensive/.../T2GcdSpec.scala](https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25), the two tests are meant to be illustrative of two ways to do things, one with an explicit testvector, and one with a more constrained-random style (which is apparently popular in industry - though not exactly supported by ScalaCheck and chisel at the moment).

For example, in line 40 we could also write:
```scala
val testData: List[(Int, Int)] = List[(Int, Int)](
  (1, 2),
  (3, 4),
  (4, 5)
)

testData.foreach { (a: Int, b: Int) =>
```
using your static testvectors, instead of using ScalaCheck to generate inputs.

I do agree that this might make failures harder to diagnose, since the values that failed aren't reflected in the test name. Though, there's also a balance here, since if you tried to make the test names so descriptive, there might be so many individual tests that the test output becomes spammy. So far, this is kind of being addressed with expect failure messages (though the user does have to write them, and functions like drivers must propagate them).

For error messages, in the future, I think it would be interesting to debugging messages associated with timescopes, and errors can print out the timescope stack. So if we continue modifying the above example,
```scala
testData.foreach { (a: Int, b: Int) =>
  timescope(s"GCD $a, $b") {
    ... test body here
  }
}
```
Additionally, since the message can be in library functions, it means that the top-level test writer might not have to think about it and still get useful error messages.

In general, I like minimalistic approaches that make use of basic constructs (such as function calls), to minimize the learning curve. I think your example is legit, but I'm not sure if the ScalaTest feature provides significant benefits over regular function calls compared to the learning curve of the new construct?

parzival3:
HI @ducky64 and thanks again for the comments :smiley: ,

> So particularly for the linked example ccc18-testing-intensive/.../T2GcdSpec.scala, the two tests are meant to be illustrative of two ways to do things, one with an explicit testvector, and one with a more constrained-random style (which is apparently popular in industry - though not exactly supported by ScalaCheck and chisel at the moment).

Yes, you are totally right, the end goal as I've said in the original message is to achieve CRV or (Constrained Random Verification). 

> Why I think this is a good addition for the examples in chisel-tester2.
This test by itself is nothing interesting but is the first step for achieving CRV(Constrained Random Verification) in Chisel.
For example, we can obtain the same behaviour described in ASIC WORLD Random Constraints Tutorial - part 6. by combining a CSP solver like Choco Solver and Tester2 to test a set of data.
[...]

Yes you can generate a set of values in with ScalaCheck,s but this is not very useful, what is helpful is to combine CRV and Test Coverage in order to have a complete picture of your current design. So the question is how to implement coverage and CRV?

For CRV, the first step is to have a CSP solver https://en.wikipedia.org/wiki/Constraint_satisfaction_problem , with a CSP solver you can specify a problem by giving a set of variables, a domain in which these variable "lives" and a series of constraint that are applied to these variables. By giving this input, the CSP solver will output the solution from the current problem. When you have the set of solutions, you can test your design with this set of values. There are many types of problem that can be specified and supplied to a CSP solver, but I don't think that this is the place to discuss CSP in-depth, a good tutorial is available [here]( http://www.asic-world.com/systemverilog/random_constraint6.html) if you want more information.

The second step after the CSP solver is to have a coverage metrics of what you are testing. For my understanding, you supply a random amount of values generated by the CSP solver to you DUT until you reach 100% of coverage, and you know for sure that the design is validate.

That said, by browsing the net, I didn't really find anything interesting regarding CSP solver in Scala. On the contrary, in the Java world, there are quite a few projects and libraries. One of the most famous is the [Choco-Solver](https://github.com/chocoteam/choco-solver). The interface to this library is not very attractive, like any other library written in Java compared to a library in Scala,
 (going a bit of topic) or a CSP library for [python](https://labix.org/python-constraint) (this is also the library used in cocotb to achieve CRV).


Interface aside, in my example posted in the first message, I wanted to show you how you can create a CSP problem and test a design on the set of the solution generated by the solver :smiley: . The problem that I'm solving in the example is the same problem that is reported [here](http://www.asic-world.com/systemverilog/random_constraint6.html).


I think here we are trying to solve the problem on a different scale.

> I do agree that this might make failures harder to diagnose, since the values that failed aren't reflected in the test name. Though, there's also a balance here, since if you tried to make the test names so descriptive, there might be so many individual tests that the test output becomes spammy. So far, this is kind of being addressed with expecting failure messages (though the user does have to write them, and functions like drivers must propagate them).
For error messages, in the future, I think it would be interesting to debugging messages associated with timescopes, and errors can print out the timescope stack. So if we continue modifying the above example,

Passing the time scopes is an excellent idea 100% in favour of it, the problem is that with your approach, I don't see how you can avoid code duplication and test duplication. As an example, we can take a medium size project in which you have a CI/CD environment that takes care of checking the changes in your design on every commit. The tests run on every commit are a small set. You don't want to care about coverage or run complex tests on every commit. What you actually care is have a rough idea that the changes you made during your working day are working on a set of tests that are considered "safe". After you make sure that the last changes are correct, the heavy tests are done during the night. The CI/CD will run exhaustive testing checking that the desired coverage of the design is reached and it will also run those tests that are important but takes a considerable amount of time. The only important part of the Nightly test is to check if the CI/CD reports a green light or not the day after. And only if the output of the CI/CD is red, you go and check the logs. And you want to have the tests clear, you don't want to go and search on others etc. These are standard practices in the software world, and another is to follow the DRY (do not repeat yourself) principle.
So going back to your example https://github.com/ducky64/ccc18-testing-intensive/blob/master/src/test/scala/T2GcdSpec.scala#L25 , and considering it as a test that is part of a large project.
Let's say that we want to test the Gcd every commit by trying only a few values, and during night we want to run an exhaustive test that checks the coverage of your design.
The only way to not create two different tests, is to run the "each commit " test with a small set of input and then during night passing a large number of inputs.
By how do you select the ammount of data you need to use for the current test? Do you use a command-line argument? Do you use a configuration file?

IMHO, I think the inventor of scalatest face these problems a lot of times before us, and that's why he is suggesting in his documentation to create a trait that specifies the behaviour of the test that you want to run.  By following the trait design, you can achieve much higher flexibility and reduce code duplication while maintaining the test output as it was intended. Plus, ( but again this is my opinion), I don't think that understanding how to create a trait has a learning curve. 

As a last note :smiley:, what I wanted to say is that I'm not against ScalaCheck. Actually, ScalaCheck is a very, very useful library. What I wanted to point out whit my pull request is to provide a simple example on how to share a test between fixtures. How you generate those fixtures is not a problem, the can be static or dynamic what matters for me is how you write the tests in order to avoid code duplication :smiley:.

If you made until here, I really appreciate your effort, and probably I'm taking to much of your time for something not really important. So if you are not sure about my approach,  go ahead and close the pull request. I don't want to bother you about tests and how I see things.

Thanks again for your comments, work and time that you are offering :smiley:.
Cheers
Enrico.

ducky64:
So splitting fast tests / long tests is something that's been discussed on and off, but I don't think ever resolved. For reference, the context usually was developers would run tests under treadle, but CI might run them (per PR? nightly?) through all the backends. Again, we haven't found a elegant solution to this, though if you have ideas, those are welcome. And while the underlying mechanisms might be similar to this PR, it is a different problem and motivation.

Anyways, here was how I was thinking of writing the same test, which I think addresses the main needs that the example in the PR addresses:
```scala
"alu should do arithmetic" in {
  val testData: List[(Int, Int)] = List[(Int, Int)](
    (1, 2),
    (3, 4),
    (4, 5)
  )

  test(new Alu(s)) { c =>
    // this might be defined on a AluIO Bundle, or in the Alu Class, possibly with PML
    // alternatively, this can be defined to take in an Alu `c` and defined in the test class, but outside a particular test run
    def testAdd(a: Int, b: Int): Unit = {
      result = ...
      timescope(s"$a+$b=$result") {  // optional, mainly to generate a debug trace via timescope stack, whenever this gets implemented
        c.io.fn.poke(0.U)
        c.io.a.poke(a.U)
        c.io.b.poke(b.U)
        c.clock.step()
        c.io.result.expect(result.U)
      }
    }
    def testSub(...): ...

    testData.foreach { (a: Int, b: Int) =>
      testAdd(a, b)
      // maybe add a reset inbetween all operations
      // also maybe do the subtraction in a different test
      testSub(a, b)
      ...
  } }
}
```
For nightly CI runs, I think failure data is useful, but I'd imagine it's a rare bug that passes your quick unit tests, yet fails on a nightly CI, and you can get enough information out of just the test name to debug it. I'd imagine the usual process is: see failure from last night (or whatever), stare at the log a bit confused, repeat the test on your local machine, break out the heavy debugging tools like waveform viewers, and go from where. At the end of the day, I think it's a trade-off of stuffing more debugging information into the test name vs. having a giant list of very tiny tests. I don't know where that line is exactly, but I think that one op per test might be overkill.

I would consider ScalaCheck vs. user-defined testvectors an orthogonal issue. That being said, we haven't done much with constrained-random; the ScalaCheck example is a quick and dirty example using Scala-ey constructs and libraries towards constrained-random, though it probably doesn't match the features and needs of industry-level constrained-random. Currently ChiselTest focuses on supporting hand-written testvectors (kind of the motivation for ChiselTest is to make the easy things really easy), but I think much of the infrastructure and libraries also works for constrained-random.

parzival3:
Hi @ducky64 and thanks again for your comments :smiley: 

> So splitting fast tests / long tests is something that's been discussed on and off, but I don't think ever resolved. For reference, the context usually was developers would run tests under treadle, but CI might run them (per PR? nightly?) through all the backends. Again, we haven't found a elegant solution to this, though if you have ideas, those are welcome. And while the underlying mechanisms might be similar to this PR, it is a different problem and motivation.

This is a good question, and my idea is to tag your test, as an example
```scala
trait AluBehavior {
  this: AnyFlatSpec with ChiselScalatestTester =>

  def mask(s: Int): Int = (1 << s) - 1

  def testAddition(a: Int, b: Int, s: Int, ann: Seq[BackendAnnotation]): Unit = {
    val result = (a + b) & mask(s)
    it should s"+ $a, $b and the result == $result" in {
      test(new Alu(s)) { c =>
        c.io.fn.poke(0.U)
        c.io.a.poke(a.U(s.W))
        c.io.b.poke(b.U(s.W))
        c.clock.step()
        c.io.result.expect(result.U(s.W))
      }
    }
  }
```
You pass the sequence of backends that you want to use in your test as a parameter, and then you tag your tests
```scala
@tags.Commit
class AluTestCommit extends AnyFlatSpec with AluBehavior with ChiselScalatestTester with Matchers {
  behavior of "ALU"
  val ann = Seq(TreadleBackendAnnotation)
  // This is for simulating a small test to run each commit
  val data = (4, 5)
  it should behave like testAddition(data._1, data._2, 4, ann)
  it should behave like testSubtraction(data._1, data._2, 4, ann)
  it should behave like testOr(data._1, data._2, 4, ann)
  it should behave like testAnd(data._1, data._2, 4, ann)
}

@tags.Nightly
class AluTestNightly extends AnyFlatSpec with AluBehavior with ChiselScalatestTester with Matchers {
  val ann = Seq(VerilatorBackendAnnotation, TreadleBackendAnnotation) // Here you can add all the backends or just a few
  behavior of "ALU"
  // this is just to simulate a very long test
  val testData: List[(Int, Int)] = List[(Int, Int)](
    (1, 2),
    (3, 4),
    (4, 5)
  )
  testData.foreach { data =>
    it should behave like testAddition(data._1, data._2, 4, ann)
    it should behave like testSubtraction(data._1, data._2, 4, ann)
    it should behave like testOr(data._1, data._2, 4, ann)
    it should behave like testAnd(data._1, data._2, 4, ann)
  }
}
```

And then in your CI configuration, you specify that you want to run sbt on every commit like this
```shell
sbt "testOnly * -- -n chiseltest.tags.Commit"
```

And every Pull request and Nightly 
```scala
sbt "testOnly * -- -n chiseltest.tags.Nightly"
```

As you see the code in the `trait AluBehaviour` didn't change, of course, I had to add a new parameter, but nothing else and I created two test suites that are run in specific situations.  I don't know how you would solve the problem if you only hade one big test case that does everything like in the [code above](https://github.com/ucb-bar/chisel-testers2/pull/196#issuecomment-697100371). This code is highly coupled, one test that tests everything. Personally, I don't think it is a good idea.


> For nightly CI runs, I think failure data is useful, but I'd imagine it's a rare bug that passes your quick unit tests, yet fails on a nightly CI, and you can get enough information out of just the test name to debug it. I'd imagine the usual process is: see failure from last night (or whatever), stare at the log a bit confused, repeat the test on your local machine, break out the heavy debugging tools like waveform viewers, and go from where. At the end of the day, I think it's a trade-off of stuffing more debugging information into the test name vs. having a giant list of very tiny tests. I don't know where that line is exactly, but I think that one op per test might be overkill.

Yeah probably you are right, but you can easily extend `chisel-tester2` by including `flexmark-java`, I tried with version `0.35.10`
```scala
libraryDependencies ++= Seq(
  "com.vladsch.flexmark" % "flexmark-all" % "0.35.10"
)
```

With this library, you can obtain an excellent html output of your tests
image
And of course, there is an option to use a custom `CSS`, and it would be nice to have a custom `CSS` for chisel-tester2 which includes an svg of the waveform generated in a failed test :wink: Or a direct link to the waveform.

To obtain this output I've run the test like

```shell
sbt "testOnly * -- -n chiseltest.tags.Nightly -h output_folder"   
```
And [Here](https://www.scalatest.org/user_guide/using_scalatest_with_sbt) you can find the scala test documentation with all the options.

Moreover, having the output of the waveform in front of your screen without having to run the simulation is always a good idea. Also, you have a trace of your design. You can always go back in time. How and why our design was failing one month ago? Just browse the CI interface, find the test that was failing one month ago and grab the waveform :smiley:



Regarding this point
> I would consider ScalaCheck vs. user-defined testvectors an orthogonal issue. That being said, we haven't done much with constrained-random; the ScalaCheck example is a quick and dirty example using Scala-ey constructs and libraries towards constrained-random, though it probably doesn't match the features and needs of industry-level constrained-random. Currently ChiselTest focuses on supporting hand-written testvectors (kind of the motivation for ChiselTest is to make the easy things really easy), but I think much of the infrastructure and libraries also works for constrained-random.

Yeah you are totally right, I don't think chisel-tester2 should integrate a CSP solver or focus on CRV, the infrastructure is there, is very flexible, and it is very well structured. My comment in the PR message was just to give some context in why I wanted to open this PR. My idea was to share with other people how to "share test" across multiple data. The train of thoughts behind this PR was like: Why do you want to share a test across multiple data? Because you want to stress-test your design. What is the best way to stress-test your design? The best way is to use CRV ..... 

Again what you are doing is amazing, and these are just my ideas on how tests should be carried, and I really appriciate all your comments.

Thanks again for your comments and time :smiley: 
Cheers,
Enrico

ducky64:
So this PR looks like it's becoming an omnibus discussion for all kinds of different features. I get that they're all kind of related, but not sure if it's helpful in terms of organization. It might be worth discussing them in separate issues (or PRs, if there's prototype code).

## This PR specifically
I think the two main issues I have with this PR's test structure are:
1. The structure of `testAddition` (and similar) as top-level tests means they're less composable and re-usable than if they were a test-snippet (eg, testAddition is just the stuff inside the `test() {...}` block, and would be called inside an enclosing `test() {...}` block - much like DecoupledDriver. I do get that this is maybe more demo / toy example code specifically trying to generate top-level unit tests, so it might really be a matter of this next point...
2. Each test case is very small, which has drawback of spamming the console (or other test listing) and incurring large per-test spin-up time (which might be alleviated with re-using pre-compiled modules / tests)

That being said, both really come down to matters of taste. I'll ask for wider thoughts at the next chisel-dev meeting.


## Multiple Backends
I think this is where I'd like more input from the rest of the community. There's already an active issue #65, though I think the focus is slightly different there (run all tests on all backends, where your example adds backends to specific tests).


## Tagged Tests
Tags have been discussed (but not fully resolved) in chisel3's test suite at freechipsproject/chisel3#791, though it seems like a stale PR. One issue there was a lack of standardization on the tag syntax, because ScalaTest offers 3 different syntaxes for tagging tests. It would be nice if there was one syntax that we could standardize on while addressing all the use cases (or at least have clear recommendations for which to use when).


## Waveforms
I can definitely see how waveforms is useful, but is it typical practice to run tests with waveforms? I think one reason waveforms are off by default is that they're expensive computationally - so you almost certainly don't want them running when you're running tests interactively, or as a pre-commit.

I don't know what accepted best practice is for nightly tests, though - you're not directly burning engineer-hours, though it does burn compute time and probably the vast majority of the time the waveforms are going to be for successful tests that no one will look at. Though having the waveforms removes the friction cost of needing to re-run a failing test locally with waveforms enabled, after the fact - not sure how that trade-off is evaluated, especially at scale.

ducky64 (commented):
Ok, we discussed this at the dev meeting earlier today, and it seems there's many different opinions about test granularity - so I've added a comment suggestion to make this clear. I've also added a TODO to re-use compiled simulators for when #212 gets merged. (while having the test cases run one after the other works here, in general we want a reset for independence between test cases)

Otherwise, the code looks good, and if the proposed changes are acceptable, then we can get this merged.

jackkoenig:
@Mergifyio backport 0.3.x

PR #204, opened 2020-09-28
Add Coverage Annotation for VCS and Verilator

ryan-lund:
Add new annotations for tests that allow coverage (line, toggle) to be enabled when backend is Verilator or VCS. Created a corresponding unit test for verilator that confirms existence of coverage file.


sequencer:
Basically, for coverage or other verification annotations,
I think we may need to figure out a way to push these annotation to FIRRTL, and process it in chiseltest or treadle. since it's nice to avoid the problem what current `WriteVcdAnnotation` did(we have two version of `WriteVcd`). What's your idea @chick @ducky64.
Another idea is, from knowledge of mine, coverage may need to annotate to specific module/signal? we may need to annotate a bunch of modules/signals, convert them to verilator/vcs configuration files like [verilator](https://www.veripool.org/projects/verilator/wiki/Manual-verilator#CONFIGURATION-FILES) did.
For example if we wanna disable coverage for a specific module, we need to annotate to that module and generate a configuration with `coverage_block_off -module "<modulename>" -block "<blockname>"`

ryan-lund:
I've updated this to rebase on current master.

@sequencer with regards to annotate specific modules/signal, my long term view of how this would work is that an annotation can be used to specify coverage on a per item basis (with the default coverage setting being off for all modules). This is something that I'm currently looking into.

ducky64 (approved):
The code looks fine, and I think this is a good near-term solution to get this functionality while being limited to internal. Considering adding a short comment to the LineCoverageAnnotation and ToggleCoverageAnnotation classes that indicate that these might be superseded in the future.

I think in the long term, since these flag annotations might be common infrastructure, it does make sense to have them in FIRRTL. But I think unless there's a FIRRTL PR doing the same thing being fast-tracked right now, it makes sense to merge this functionality while planning for it to be superseded when done right.

ducky64:
Let me know when you think this is ready for merge.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -77,18 +77,35 @@ object VerilatorExecutive extends BackendExecutive {
       .collectFirst { case VerilatorCFlags(f) => f }
       .getOrElse(Seq.empty)
     val writeVcdFlag = if(compiledAnnotations.contains(WriteVcdAnnotation)) { Seq("--trace") } else { Seq() }
+    val coverageFlags = if(compiledAnnotations.contains(StructuralCoverageAnnotation)) {Seq (
+      "--coverage-line",
+      "--coverage-toggle",
+      if(compiledAnnotations.contains(UserCoverageAnnotation)) {"--coverage-user"} else {""}
```
ducky64, line 83: Ditto, is a `collect` better here?

src/test/scala/chiseltest/experimental/tests/VcsCoverageTests.scala
```diff
@@ -0,0 +1,177 @@
...
+
+class VcsCoverageTests extends AnyFlatSpec with ChiselScalatestTester with Matchers {
+  behavior of "Testers2"
+
+  it should "allow specifying line coverage for Vcs" in {
+    assume(firrtl.FileUtils.isVCSAvailable)
+
+    val outputStream = new ByteArrayOutputStream()
+    Console.withOut(new PrintStream(outputStream)) {
+      test(new Module {
```
ducky64, line 22: For all these tests, if you're not actually looking at the circuit behavior (and I don't see anything here that is doing that?), does it make sense to test on an empty module (perhaps there should even be an EmptyModule test utility class - more as a way of saying "this module intentionally left empty")? I'd prefer tests to be orthogonal - if the circuit stuff broke but this was fine, I'd prefer this not to fail.

And ditto for all the instances below.

src/test/scala/chiseltest/experimental/tests/VcsCoverageTests.scala
```diff
@@ -0,0 +1,177 @@
...
+        c.io.b.expect(1.U)
+        c.clock.step()
+        c.io.a.poke(42.U)
+        c.io.b.expect(42.U)
+      }
+    }
+    val output = outputStream.toString
+    output.contains("-cm line+tgl+branch+cond+assert") should be(true)
+  }
+}
```
ducky64, line 177: nit: newline at end of file preferred I think. Same for other files.

ducky64 (changes requested):
A few stylistic comments. Functionally still looks fine.

src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
```diff
@@ -95,16 +95,38 @@ object VcsExecutive extends BackendExecutive {
...
+        "tgl",
+        "branch",
+        "cond",
+        if (compiledAnnotations.contains(UserCoverageAnnotation)) {"assert"} else {""})
+      .filter(item => item != "")
+      .mkString("+"))
+    } else if (compiledAnnotations.intersect(Seq(LineCoverageAnnotation, ToggleCoverageAnnotation,
+      BranchCoverageAnnotation, ConditionalCoverageAnnotation, UserCoverageAnnotation)).nonEmpty) {
+      Seq("-cm " + Seq(
+        if (compiledAnnotations.contains(LineCoverageAnnotation)) {"line"} else {""},
```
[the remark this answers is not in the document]
ryan-lund: Resolved all in newest push

src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
```diff
@@ -95,16 +95,29 @@ object VcsExecutive extends BackendExecutive {
     val moreVcsCFlags = compiledAnnotations
       .collectFirst { case VcsCFlags(flagSeq) => flagSeq }
       .getOrElse(Seq())
+    val coverageFlags = Seq((compiledAnnotations collect Map(
```
ducky64, line 98: Would something like the second code block in the first answer to this [StackOverflow post](https://stackoverflow.com/questions/4784051/scala-partition-collect-usage) be cleaner? Basically, can you (and similarly for the code below):
- instead of collect with a Map, can each be a case statement, and
- instead of putting everything in a Seq and then flattening it, can you use the output of the collect directly, which should be an iterable?

Also, with use of collect, which should drop irrelevant annotations instead of emitting empty strings, can you drop the `if (item.isEmpty)`?

ducky64 (commented):
One more thing

src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
ryan-lund: Map to case done.

For the flattening the process is Seq(collect -> flatten -> distinct -> string joined by "+") which is why we then have the if (item.isEmpty) on in map, effectively prepending "-cm" to the string in the Seq (type used for flag passing) if it exists.

ducky64 (approved):
lgtm

jackkoenig:
@Mergifyio backport 0.3.x

PR #211, opened 2020-10-10
Add thread ordering documentation and supporting test

### LINKED ISSUE #202: Cuncrrency question
This is more a clarification question instead of an issue. Is it correct that fork actually implements somthing like a coroutine, meaning that all forked code is executed sequentically and the switch between the code blocks happens on step()? Threads are only used to implement this functionallity and nothing is reall running in parllel.

Therefore, is it safe to do following:
```
var cnt = 0

fork {
  for (...) {
    cnt  += 1
    c.step()
  }
}

for (...) {
  cnt  += 1
  c.step()
}
```

Meaning there is no race condition on the increment of ```cnt```.

ducky64:
As described in the title, adds the fork behavior to the README, and a test for it.

Resolves #202

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,14 @@ See the test cases for examples:
 ## New Constructs
 - `fork` to spawn threads, and `join` to block (wait) on a thread.
   Pokes and peeks/expects to wires from threads are checked during runtime to ensure no collisions or unexpected behavior.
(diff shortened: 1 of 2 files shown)


chick (approved):
lgtm

jackkoenig:
@Mergifyio backport 0.3.x

PR #218, opened 2020-10-30
Use second element in Verilator version string for compatibility detection

ducky64:
Seems more stable across Verilator versions, but still, watch as this breaks something.

Fixes the after-party of #187


chick (approved):
Looks good to me

jackkoenig:
@Mergifyio backport 0.3.x

PR #221, opened 2020-11-20
Update SBT to 1.4.3

### LINKED ISSUE #209: Issue with scalatest transitive dependency
It appears as if the published artifact for `0.3.0` has has something wrong with its `scalatest` transitive dependency.

Running just again `chiseltest` as in https://scastie.scala-lang.org/gdDNgFnJRtOB95nRdUOJCA gives a linkage error in a `scalatest` API. This can be resolved by adding an explicit dependency on `scalatest` `3.2.x`: https://scastie.scala-lang.org/0C0JUwuJROKouurFavLvbg

The build for `chiseltest` is `scalatest` `3.0.8` in the main configuration and `3.2.0` in the test configuration. However, it appears as if the main artifact is actually compiled against `3.2.0`. This main be something wrong in our SBT build.

jackkoenig:
Fixes #209

I did some digging, and it turns out that on `master` with SBT `1.3.13`, you can see the error by looking at the `Compile / fullClasspath` vs. the published pom. See discussion here: https://gitter.im/sbt/sbt?at=5fb8374c29cc4d73481fb6b3

I found that updating SBT fixes the issue, the compile classpath and pom are properly in sync.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.3.13
+sbt.version=1.4.3


jackkoenig:
Failure is the same CI issue Treadle had, I'll look into cleaning up the CI a bit.

jackkoenig:
Closing in favor of https://github.com/ucb-bar/chisel-testers2/pull/226 which bumps to `1.4.4`

PR #222, opened 2020-11-20
Simplify CI

jackkoenig:


jackkoenig:
It's not clear if caching will work when running inside of the chisel-tools Docker container so this should be fun.

jackkoenig:
Okay so this "works" but caching doesn't yet

chick (approved):
lgtm

jackkoenig:
This currently does not cache downloaded dependencies, but it's fast enough that it isn't a huge problem. Future work is to make caching work properly.

jackkoenig:
@Mergifyio backport 0.3.x

PR #227, opened 2020-11-24
Add .mergify.yml

jackkoenig:
Currently only with backporting to 0.3.x

As described on the related Treadle PR, no need to backport this: https://github.com/chipsalliance/treadle/pull/266#issuecomment-730042187


chick (approved):
Excellent

PR #239, opened 2020-11-30
Pass Coverage Flags to VCS Simulation

TsaiAnson:
Adding the missing coverage flag arguments to the VCS simulation command. Without it, the simulation won't monitor for coverage.

### DIFF
--- a/src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
+++ b/src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
@@ -122,7 +122,7 @@ object VcsExecutive extends BackendExecutive {
       .collectFirst[Seq[String]] {
         case TestCommandOverride(f) => f.split(" +")
       }
-      .getOrElse { Seq(new File(targetDir, circuit.name).toString) }
+      .getOrElse { Seq(new File(targetDir, circuit.name).toString) } ++ coverageFlags
 
     val paths = compiledAnnotations.collect { case c: CombinationalPath => c }
 


src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
```diff
@@ -122,7 +122,7 @@ object VcsExecutive extends BackendExecutive {
       .collectFirst[Seq[String]] {
         case TestCommandOverride(f) => f.split(" +")
       }
-      .getOrElse { Seq(new File(targetDir, circuit.name).toString) }
+      .getOrElse { Seq(new File(targetDir, circuit.name).toString) } ++ coverageFlags
```
ducky64, line 125: `coverageFlags` is part of `vcsFlags` on line 108, why does it need to be duplicated here? Why does this need to be handled differently from the other flags?

ducky64 (commented):
Not sure I understand this change, see one inline comment.

src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
TsaiAnson: Yeah, the duplication is needed due to how coverage works in the VCS flow. The coverage flags passed into the vcs command on line 108 tells the tool to compile the design for the types of coverage. The same flags are also needed in the simulation command to tell the simulator which coverage to actively monitor.

This seems kind of redundant, but it allows for designs to be compiled for several types of coverage while the simulator can monitor them in isolation (e.g. while the design is compiled with line + branch coverage, a specific test can only monitor for line coverage).

TsaiAnson: I found this piece of documentation helpful: http://read.pudn.com/downloads343/ebook/1502221/cov_ug.pdf (page 47, 2-35)

ducky64: Makes sense. Can you add an abbreviated version of that explanation in the comments so it's obvious to others looking at the code why it's being duplicated?

Also, the book link doesn't work (anymore?) 🙃.

I think that's the only change I'd like.

ducky64 (changes requested):
Add the explanation as a short comment and it's good!

src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
TsaiAnson: Sounds good. Just committed!

ducky64 (approved):
lgtm, thanks!

PR #240, opened 2020-11-30
Fix VCS Double Elaboration Bug

TsaiAnson:
I recently ran into an error when switching from the VerilatorBackendAnnotation to the VcsBackendAnnotation:
```
[info] TLRAMTest:
[info] - should Driver/Monitor Master Hardcoded Burst TLRAM *** FAILED ***
[info]   firrtl.options.StageError:
[info]   at chisel3.stage.ChiselStage.run(ChiselStage.scala:49)
[info]   at chiseltest.legacy.backends.vcs.VcsExecutive$.start(VcsExecutive.scala:61)
[info]   at chiseltest.defaults.package$.createDefaultTester(defaults.scala:24)
[info]   at chiseltest.ChiselScalatestTester$TestBuilder.apply(ChiselScalatestTester.scala:33)
[info]   at verif.TLRAMTest.$anonfun$new$1(TLRAMTest.scala:42)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   ...
[info]   Cause: chisel3.internal.ChiselException: Error: attempted to instantiate a Module, but nothing happened. This is probably due to rewrapping a Module instance with Module(). See @[ChiselAnnotations.scala 50:46]
[info]   at chisel3.internal.throwException$.apply(Error.scala:85)
[info]   at chisel3.Module$.do_apply(Module.scala:58)
[info]   at chisel3.stage.ChiselGeneratorAnnotation.$anonfun$elaborate$1(ChiselAnnotations.scala:50)
[info]   at chisel3.internal.Builder$.$anonfun$build$1(Builder.scala:414)
[info]   at scala.util.DynamicVariable.withValue(DynamicVariable.scala:62)
[info]   at chisel3.internal.Builder$.build(Builder.scala:412)
[info]   at chisel3.stage.ChiselGeneratorAnnotation.elaborate(ChiselAnnotations.scala:50)
[info]   at chisel3.stage.phases.Elaborate.$anonfun$transform$1(Elaborate.scala:19)
[info]   at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:245)
[info]   at scala.collection.immutable.List.foreach(List.scala:392)
[info]   ...
```
Given that it worked with the Verilator backend, I resolved the issue by comparing and copying over the structure from the VerilatorExecutive. I'm not exactly sure on what the underlying error was, but it seems like the previous error was due to a "double elaboration" on `generatorAnnotation` during the second pass, although I might be wrong. This PR fixes the error I was getting. 

Other details:
The test that I ran (with the VcsBackendAnnotation): https://github.com/TsaiAnson/verif/blob/8f47c8f5674343db8ee754a076ea0047695b29bb/src/test/scala/TLRAMTest.scala#L13

### DIFF
--- a/src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
+++ b/src/main/scala/chiseltest/legacy/backends/vcs/VcsExecutive.scala
@@ -43,7 +43,8 @@ object VcsExecutive extends BackendExecutive {
     val targetDirFile = new File(targetDir)
 
     val generatorAnnotation = chisel3.stage.ChiselGeneratorAnnotation(dutGen)
-    val circuit = generatorAnnotation.elaborate.collect { case x: ChiselCircuitAnnotation => x }.head.circuit
+    val elaboratedAnno = (new chisel3.stage.phases.Elaborate).transform(annotationSeq :+ generatorAnnotation)
+    val circuit = elaboratedAnno.collect { case x: ChiselCircuitAnnotation => x }.head.circuit
     val dut = getTopModule(circuit).asInstanceOf[T]
 
     // Generate the verilog file and some or all of the following annotations
@@ -58,10 +59,7 @@ object VcsExecutive extends BackendExecutive {
 
     val compiledAnnotations = (new ChiselStage)
       .run(
-        annotationSeq ++ Seq(
-          generatorAnnotation,
-          RunFirrtlTransformAnnotation(new VerilogEmitter)
-        )
+        elaboratedAnno :+ RunFirrtlTransformAnnotation(new VerilogEmitter)
       )
       .filterNot(_.isInstanceOf[DeletedAnnotation])
 


ducky64 (approved):
This looks fine.

@chick, any thoughts, since I think you migrated these backends over initially?

chick (approved):
I think this looks ok. It should follow pretty much the same flow as `VerilatorExecutive` and this seems to do that.

PR #252, opened 2020-12-29
Peek / Poke records check type equivalence for both data fields

parzival3:
Currently I have to write code similar to this

```scala
class AluInput(val size: Int) extends Bundle with RandBundle {
  val a = UInt(size.W)
  val b = UInt(size.W)
  val fn = UInt(2.W)

override def typeEquivalent(data: Data): Boolean = {
    data match {
      case _: (AluInput @ AluInputConstraint(size)) => true
      case _ => false
    }
  }
}

class AluInputTransaction(size: Int) extends AluInput(size) {

  def expectedResult(): AluOutput = {
      var result: BigInt = 0
      if (fn.litValue() == 0) {
        result = a.litValue() + b.litValue()
      } else if (fn.litValue() == 1) {
        result = a.litValue() - b.litValue()
      } else if (fn.litValue() == 2) {
        result = a.litValue() | b.litValue()
      } else {
        result = a.litValue() & b.litValue()
      }
      new AluOutput(8).Lit(_.result -> result.U)
  }
}
```

It would be nice if all the logic would be moved from the `AluInput` bundle to `AluInputTransaction` and overriding the `typeEquivalent` method only in `AluInputTransaction`

```scala
class AluInput(val size: Int) extends Bundle with RandBundle {
  val a = UInt(size.W)
  val b = UInt(size.W)
  val fn = UInt(2.W)

}

class AluInputTransaction(size: Int) extends AluInput(size) {

override def typeEquivalent(data: Data): Boolean = {
    data match {
      case _: (AluInput @ AluInputConstraint(size)) => true
      case _ => false
    }
  }

  def expectedResult(): AluOutput = {
      var result: BigInt = 0
      if (fn.litValue() == 0) {
        result = a.litValue() + b.litValue()
      } else if (fn.litValue() == 1) {
        result = a.litValue() - b.litValue()
      } else if (fn.litValue() == 2) {
        result = a.litValue() | b.litValue()
      } else {
        result = a.litValue() & b.litValue()
      }
      new AluOutput(8).Lit(_.result -> result.U)
  }
}
```
So I propose the changes in `peek` and `poke` methods for the `Record` type.

### DIFF
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -113,7 +113,7 @@ package object chiseltest {
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         pokeBits(x, value.litValue)
       case (x: Record, value: Record) => {
-        require(DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
+        require(DataMirror.checkTypeEquivalence(value, x) || DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
         (x.elements.zip(value.elements)).foreach {
           case ((_, x), (_, value)) =>
             x.poke(value)
@@ -176,7 +176,7 @@ package object chiseltest {
         require(x.binaryPoint == value.binaryPoint, "binary point mismatch")
         Context().backend.expectBits(x, value.litValue, message, Some(BitsDecoders.fixedToString(x.binaryPoint)), stale)
       case (x: Record, value: Record) => {
-        require(DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
+        require(DataMirror.checkTypeEquivalence(value, x) || DataMirror.checkTypeEquivalence(x, value), s"Record type mismatch")
         (x.elements.zip(value.elements)).foreach {
           case ((_, x), (_, value)) =>
             x.expectWithStale(value, message, stale)


ducky64:
Uh, I'm not sure this is a clean solution in general - `typeEquivalent` in general should not be overridden, and doing so might break other things in chisel or result in brittle code, especially since now `a.typeEquivalent(b)` can be different than `b.typeEquivalent(a)`.

What's the specific use case (what code is this supposed to enable)? I think we should consider other possible solutions, especially ones that don't do weird things with core constructs.

parzival3:
Humm, yeah, the fact that it is not symmetric is probably a no go. 

I would like to extend a basic input Bundle like `AluInput` and encapsulate some logic in it, and then use this "extended" bundle to poke some data in the module under test. 

The final goal would be to create a random constraint transaction, where a "constraint random transaction" is simply an extended bundle that contains some logic and can be randomized. Something like this https://github.com/chiselverify/chisel-crv/blob/RandomBundles/src/test/scala/chisel/bundle/random/AluRandomTest.scala. The problem with my example is that the generated random `Bundle` has a different type than the `AluInput` and thus poking it into the module fails. Currently, the two possible ways (that I'm aware of ) to solve this problem are to override `typeEquivalent` in the `AluInput` bundle or use `asInstanceOf[AluInput]`.

Or probably this is not the correct way of generating random transactions... @ducky64 do you think this is achievable  without using `asInstanceOf` or `typeEquivalent`

ducky64:
I think I see what you're trying to do (and correct me if I'm wrong): you're trying to make a class that acts like another Bundle, but is the output of a randomize operator. Would it make sense to instead have `.randomBundle` return a randomized Bundle Literal of type `AluInput` (instead of `AluInputConstraint`)? You could then generate the output somewhere else (possibilities might be `Alu.getOutputBundleOf(AluInput) -> AluOutput` or `AluInput.outputOf -> AluOutput` or `AluInputConstraint.outputOfInput(AluInput) -> AluOutput`, or having `.randomBundle` return an input and output pair).

In general, I think it makes sense to directly use chisel types where possible, instead of having custom types masquerade as something else?

parzival3:
Happy new year,

> I think I see what you're trying to do (and correct me if I'm wrong): you're trying to make a class that acts like another Bundle, but is the output of a randomize operator. Would it make sense to instead have .randomBundle return a randomized Bundle Literal of type AluInput (instead of AluInputConstraint)?

Yeah exactly I would like to return a Bundle Literal of type `AluInput` but I'm struggling to create a generic function that returns it...

> In general, I think it makes sense to directly use chisel types where possible, instead of having custom types masquerade as something else?

Yeah, the problem with chisel types is that I find it hard to handle them outside of a `Module`, or am I missing something?

But I think we are going out of topic, should we continue the discussion somewhere else and close this issue?

ducky64:
Happy new year to you as well.

I think continuing discussion in this PR is fine (even if it's unlikely this specific PR is going to be merged).

So for doing chisel ops outside of a Module context: generally, you can't, but literals (including Bundle literals) are the exception and may be created anywhere. So for example, you should be able to do this outside a Module context:
```scala
MyBundle.Lit(_.a -> 0.U, _.b -> 1.U)
```
because it constructs a Bundle literal, but you can't do this:
```scala
val bundle = Wire(new MyBundle())
bundle.a := 0.U
bundle.b := 1.U
```
because this generates hardware assignments - even if the outcome of both is similar in hardware.

Note that the [Bundle literals example tests](https://github.com/ucb-bar/chisel-testers2/blob/master/src/test/scala/chiseltest/tests/BundleLiteralsSpec.scala) construct Bundle literals outside a Module context.

If this isn't the underlying problem you're having, example code (even if not compiling or erroring) of what you'd like to do might be helpful.

parzival3:
Yeah, my randomization function already uses literal bundles https://github.com/chiselverify/chisel-crv/blob/RandomBundles/src/main/scala/crv/backends/jacop/VerificationContext.scala#L32. The problem is that I found it hard to programmatically generate bundle literals at runtime. The bundle that I want to randomize is composed of other bundles or other chisel types like `Vec` etc... I didn't find any simple way of doing that.

From my perspective, the only two possible ways are to parse the output of FIRRTL at runtime and create a sort of tree structure of the bundle or use compile-time macros. Am I correct, or are there more simple ways of doing that?

For example, how would you create a generic function that takes as input a  generic `Bundle` and returns a `Bundle` literal with all the values initialized to a default value?

```scala
    def initialization[T <: Bundle](b: T): T = {
      [...]
      b.Lit(...)
    }
```

Another thing that I find challenging (but most likely, this doesn't have a solution) because I would like to specify the constraint for the random bundles with chisel types.
Something like this:
```scala
class AluInputConstraint(size: Int) extends AluInput(size) with RandBundle {
  val constraints = new Constraint {
    a + b <= 255.U
    a - b >= 0.U
    fn <= 3.U
  }
}
```
Right now, I've defined my set of "operators" that starts with a `#` specifically for defining constraint.  If you have any idea on how specifying constraint like in the example above, please do 😃

ducky64:
You might also want to check out some of the recent work at Berkeley that deals with constrained random: https://github.com/TsaiAnson/verif/blob/master/core/test/ConstrainedRandomTest.scala

I'm not too familiar with the details - but it seems to me that there's some overlap with what you're trying to do?

ducky64:
Also, you can programmatically get the fields of a Bundle using `Record.elements` in [Aggregate.scala](https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala/chisel3/Aggregate.scala#L633). Also, note that the way `.Lit(...)` is defined is it takes in a function with one argument (object of its own type) that produces a pair of a element / field in that bundle to the literal value. So it should be possible to programmatically construct Bundle literals on arbitrary Bundles through this underlying mechanism, though it might not be immediately obvious given the syntactic sugar around it.

ekiwi:
Seems like this PR is stale. Feel free to re-open though.

PR #254, opened 2021-01-05
Dependencyapi phase2

sequencer:
1. refactor stage/phase introduced in #247.
1. refacotr ThreadedBacked and BackendInterface.

### DIFF
--- a/src/main/scala/chiseltest/backends/SimulatorInterface.scala
+++ b/src/main/scala/chiseltest/backends/SimulatorInterface.scala
@@ -1,0 +1,30 @@
+package chiseltest.backends
+
+import firrtl.AnnotationSeq
+
+/** API to Simulator. */
+trait SimulatorInterface {
+
+  /** start time of this test. */
+  final val startTime: Long = System.nanoTime()
+
+  /** how many nano seconds has elapsed since this test start. */
+  def elapsedNanoSeconds(): Long = System.nanoTime() - startTime
+
+  def annotations: AnnotationSeq
+
+  /** poke a signal name with a [[BigInt]]. */
+  def poke(signal: String, value: BigInt): Unit
+
+  /** peek a signal name, return a BigInt if found. */
+  def peek(signal: String): Option[BigInt]
+
+  /** step the main clock. */
+  def step(n: Int): Unit
+
+  /** end simulator. */
+  def finish(): Unit
+
+  /** interface to get current clock. */
+  def cycleCount: BigInt
+}
--- a/src/main/scala/chiseltest/exceptions.scala
+++ b/src/main/scala/chiseltest/exceptions.scala
@@ -7,7 +7,8 @@ class NotLiteralException(message: String) extends ChiselTestException(message)
 class LiteralTypeException(message: String) extends ChiselTestException(message)
 class UnpokeableException(message: String) extends ChiselTestException(message)
 class UnsupportedOperationException(message: String) extends ChiselTestException(message)
-class FailedExpectException(val message: String, val failedCodeStackDepth: Int) extends ChiselTestException(message)
+class FailedExpectException(val message: String, val stackTraceElement: StackTraceElement)
+    extends ChiselTestException(message)
 
 class ClockResolutionException(message: String) extends ChiselTestException(message)
 
--- a/src/main/scala/chiseltest/internal/BackendInterface.scala
+++ b/src/main/scala/chiseltest/internal/BackendInterface.scala
@@ -7,26 +7,6 @@ import chisel3._
 
 import scala.collection.mutable
 
-trait AbstractTesterThread
-
-class TesterThreadList(protected val elts: Seq[AbstractTesterThread]) {
-  def toSeq(): Seq[AbstractTesterThread] = elts
-
-  def join() {
-    Context().backend.doJoin(elts, None)
-  }
-
-  def joinAndStep(clock: Clock) {
-    Context().backend.doJoin(elts, Some(clock))
-  }
-
-  def ++(others: TesterThreadList): TesterThreadList = {
-    new TesterThreadList(elts ++ others.elts)
-  }
-
-  val fork: ForkBuilder = new ForkBuilder(None, None, elts)
-}
-
 /** Common interface definition for tester backends. Internal API.
   */
 trait BackendInterface {
--- a/src/main/scala/chiseltest/internal/BackendInterface.scala
+++ b/src/main/scala/chiseltest/internal/BackendInterface.scala
@@ -2,91 +2,93 @@
 
 package chiseltest.internal
 
-import chiseltest.Region
 import chisel3._
 
-import scala.collection.mutable
+import scala.math.BigInt
 
-/** Common interface definition for tester backends. Internal API.
-  */
-trait BackendInterface {
+/** SimulatorInterface usages in ThreadedBackend. */
+trait BackendInterface { this: ThreadedBackend =>
+
+  def getClock(clk: Clock): Boolean =
+    simulatorInterface.peek(dataNames(clk)) match {
+      case Some(x) if x == BigInt(1) => true
+      case _                         => false
+    }
 
   /** Writes a value to a clock.
     */
-  def pokeClock(signal: Clock, value: Boolean): Unit
+  def pokeClock(signal: Clock, value: Boolean): Unit = {
+    val intValue = if (value) 1 else 0
+    simulatorInterface.poke(dataNames(signal), intValue)
+    logger.debug(s"${resolveName(signal)} <- $intValue")
+  }
 
   /** Read the value of a clock.
     */
-  def peekClock(signal: Clock): Boolean
+  def peekClock(signal: Clock): Boolean = {
+    doPeek(signal, new Throwable)
+    val a = simulatorInterface.peek(dataNames(signal)).getOrElse(BigInt(0))
+    logger.debug(s"${resolveName(signal)} -> $a")
+    a > 0
+  }
 
   /** Writes a value to a writable wire.
     * Throws an exception if write is not writable.
     */
-  def pokeBits(signal: Data, value: BigInt): Unit
+  def pokeBits(signal: Data, value: BigInt): Unit = {
+    doPoke(signal, value, new Throwable)
+    val dataName = dataNames(signal)
+    simulatorInterface.peek(dataName) match {
+      case Some(peekValue) =>
+        if (peekValue != value) {
+          idleCycles.clear()
+        }
+        simulatorInterface.poke(dataNames(signal), value)
+        logger.debug(s"${resolveName(signal)} <- $value")
+      case None =>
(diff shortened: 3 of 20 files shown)


src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -0,0 +1,30 @@
...
+import firrtl.AnnotationSeq
+
+/** API to Simulator. */
+trait SimulatorInterface {
+
+  /** start time of this test. */
+  final val startTime: Long = System.nanoTime()
+
+  /** how many nano seconds has elapsed since this test start. */
+  def elapsedNanoSeconds(): Long = System.nanoTime() - startTime
```
ducky64, line 12: Should probably clarify this (by renaming) as walltime (to distinguish it from simulation time).
Also does it really need to be here? Unless it's meant to be a high accuracy API or something (and it doesn't seem like it), wouldn't it be cleaner to have whatever upper layer needs the performance data to handle the timing?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -0,0 +1,30 @@
...
+  def peek(signal: String): Option[BigInt]
+
+  /** step the main clock. */
+  def step(n: Int): Unit
+
+  /** end simulator. */
+  def finish(): Unit
+
+  /** interface to get current clock. */
+  def cycleCount: BigInt
```
ducky64, line 29: Cycle count on what? What if there are multiple clocks? What is the top-level isn't a module with an implicit clock?

src/main/scala/chiseltest/exceptions.scala
```diff
@@ -7,7 +7,8 @@ class NotLiteralException(message: String) extends ChiselTestException(message)
 class LiteralTypeException(message: String) extends ChiselTestException(message)
 class UnpokeableException(message: String) extends ChiselTestException(message)
 class UnsupportedOperationException(message: String) extends ChiselTestException(message)
-class FailedExpectException(val message: String, val failedCodeStackDepth: Int) extends ChiselTestException(message)
+class FailedExpectException(val message: String, val stackTraceElement: StackTraceElement)
```
ducky64, line 10: Consider making another exception superclass that is ChiselTestFailedException or something, to distinguish a test failure from a bad test or internal error? And then when you propagate failures to ScalaTest you can match on a more general superclass?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  /** start time of this test. */
+  final val startTime: Long = System.nanoTime()
+
+  /** how many nano seconds has elapsed since this test start. */
+  def elapsedNanoSeconds(): Long = System.nanoTime() - startTime
+
+  def annotations: AnnotationSeq
+
+  /** poke a signal name with a [[BigInt]]. */
```
ducky64, line 16: Consider defining behavior if the signal doesn't exist - this could be an exception or a return code. I think the exception makes more sense, so things fail noisily?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -0,0 +1,30 @@
...
+  def poke(signal: String, value: BigInt): Unit
+
+  /** peek a signal name, return a BigInt if found. */
+  def peek(signal: String): Option[BigInt]
```
ducky64, line 20: Similarly, perhaps this should throw an exception instead of returning an Option?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
ducky64: Also, define the naming convention (this can refer to some other documentation / ScalaDoc) and/or leave a TODO if this is to be refactored to a non-Stringly-typed interface in the future.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -2,111 +2,93 @@
...
-  val fork: ForkBuilder = new ForkBuilder(None, None, elts)
-}
-
-/** Common interface definition for tester backends. Internal API.
-  */
-trait BackendInterface {
+  def getClock(clk: Clock): Boolean =
+    simulatorInterface.peek(dataNames(clk)) match {
+      case Some(x) if x == BigInt(1) => true
+      case _                         => false
```
ducky64, line 15: Should this really be the result for the None case? Unless it makes sense to fail silently here (I think this usually should never be the case), I think an exception might make more sense? Also see above comment in the definition of peek.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -2,111 +2,93 @@
...
+    simulatorInterface.poke(dataNames(signal), intValue)
+    logger.debug(s"${resolveName(signal)} <- $intValue")
+  }
 
   /** Read the value of a clock.
     */
-  def peekClock(signal: Clock): Boolean
+  def peekClock(signal: Clock): Boolean = {
+    doPeek(signal, new Throwable)
+    val a = simulatorInterface.peek(dataNames(signal)).getOrElse(BigInt(0))
```
ducky64, line 30: Similarly, don't like silent failures.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -2,111 +2,93 @@
...
+    val dataName = dataNames(signal)
+    simulatorInterface.peek(dataName) match {
+      case Some(peekValue) =>
+        if (peekValue != value) {
+          idleCycles.clear()
+        }
+        simulatorInterface.poke(dataNames(signal), value)
+        logger.debug(s"${resolveName(signal)} <- $value")
+      case None =>
+        logger.debug(s"${resolveName(signal)} is eliminated by firrtl, ignore it. ")
```
ducky64, line 49: Similarly, would prefer this throws an exception. The exception text can provide diagnostic hints (eg, "this signal doesn't exist, was it eliminated by firrtl?"), but it should fail very noisily.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -2,111 +2,93 @@
...
-  def getParentTraceElements: Seq[StackTraceElement] = Seq()
-
-  protected val testMap = mutable.HashMap[Any, Any]()
-
-  /** Sets the value associated with a key in a per-test map.
-    */
-  def setVar(key: Any, value: Any): Unit = {
-    testMap.put(key, value)
+  def peekBits(signal: Data, stale: Boolean): BigInt = {
+    require(!stale, "Stale peek not yet implemented")
```
ducky64, line 58: Feel free to remove the stale stuff. I don't think it's going to happen (or that it even was a good idea).

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -2,111 +2,93 @@
...
+
+    doPeek(signal, new Throwable)
+    val dataName = dataNames(signal)
+    val a = simulatorInterface.peek(dataName) match {
+      case Some(peekValue) =>
+        logger.debug(s"${resolveName(signal)} -> $peekValue")
+        peekValue
+      case None =>
+        logger.debug(s"${resolveName(signal)} is eliminated by firrtl, default 0.")
```
ducky64, line 67: Ditto about silent failures.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -36,14 +35,183 @@ case class ForkBuilder(name: Option[String], region: Option[Region], threads: Se
...
+  /** Returns the value associated with the key in a per-test map.
+    */
+  def getVar(key: Any): Option[Any] = {
+    testMap.get(key)
+  }
+
+  /** Check a signal is equal to a value,
+    * if not equal, store this exception to [[chiselTestExceptions]].
+    */
+  def expectBits(
```
ducky64, line 169: I think expectBits belongs in the simulator interface. Note that in the current design, BackendInterface is meant to be the simulator interface, so that backends could implement optimized expects instead of needing to roundtrip to Scala each time. Having expect as a "primitive" operation also allows things like SteppedHWIOTester which generates test using a certain subset of primitives down to hardware. I'm not convinced we want to architecturally foreclose that route, even if we haven't used it yet.

ducky64 (changes requested):
The two main things I disagree with is:
1. Silent failure (or debug messages) when signal is not found - I think it should error noisily. If the false positive rate is so high that this would cause usability errors, then the false positives should be fixed at the sources. I think giving a wrong answer is worse than tests failing, because the wrong answer will probably be confusing and consume lots of debugging time.
2. The new splitting of BackendInterface into a separate simulator interface and threading infrastructure. See comment in code for details - but the current architecture (as imperfect as the implementation is) was meant to support synthesizing subsets of tests to hardware instead of being exclusive to Scala-as-top software simulations.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
sequencer: maybe expose an API to return `wallTime` in second?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
sequencer: this API should be removed.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
ducky64: So my argument is that this API doesn't belong here in the first place. Not that time isn't useful anywhere - but that timing code should go where it is used.

If you really feel that most SimulatorInterface use cases need timing data (which I don't think so), consider making a mixin trait that provides the timing data, while leaving SimulatorInterface as a bare essentials simulator interface API.

src/main/scala/chiseltest/internal/BackendInterface.scala
sequencer: all stale removed.

src/main/scala/chiseltest/internal/BackendInterface.scala
sequencer: fixed

src/main/scala/chiseltest/internal/ThreadedBackend.scala
sequencer: I think so. but in current behavior, `expectBits` is implemented in Scala to directly compare.
I also think in the future. After Verification API landing to Chisel. I think `expect` API can finally deprecated. let ScalaTest only `peek`/`poke`, let `assume` `assert` `cover` stuffs to simulator, and expose correspond error to chisletest.
but in current implementation, maybe we just can leave this as it is?

ekiwi:
This PR has gotten stale, so I will close is for now. Feel free to re-open.

PR #256, opened 2021-01-05
Dependencyapi phase4

sequencer:
Fixes for reviews and CIs


src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
+  private[chiseltest] def poke(signal: String, value: BigInt): Unit
+
+  /** peek a signal name, return BigInt if found.
+    * If signal is not found in circuit, return [[None]].
+    */
+  private[chiseltest] def peek(signal: String): Option[BigInt]
 
-  /** peek a signal name, return a BigInt if found. */
-  def peek(signal: String): Option[BigInt]
+  /** Ask Simulator to step to next n delta cycle. */
```
ducky64, line 32: Does it make sense to define delta cycles? What does it even mean to step multiple delta cycles? As far as I understand, the current way things work is that there's an update that propagates all combinational operations - and there isn't really a concept of how many updates there is, it's updated until essentially fixed-point?

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
+  /** start time of this test. */
+  final private[chiseltest] lazy val endTime: Long = elapsedNanoSeconds()
 
-  def annotations: AnnotationSeq
+  /** how many nano seconds has elapsed since this test start. */
+  final private[chiseltest] def elapsedNanoSeconds(): Long = System.nanoTime() - startTime
 
   /** poke a signal name with a [[BigInt]]. */
-  def poke(signal: String, value: BigInt): Unit
+  private[chiseltest] def poke(signal: String, value: BigInt): Unit
```
ducky64, line 25: As (probably?) mentioned in a prior phase, signal name format needs to be defined, especially because this is an interface trait.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
+
+  /** peek a signal name, return BigInt if found.
+    * If signal is not found in circuit, return [[None]].
+    */
+  private[chiseltest] def peek(signal: String): Option[BigInt]
```
ducky64, line 30: ditto

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
-  def step(n: Int): Unit
+  /** start simulator. */
+  private[chiseltest] def start(): Unit = {
+    logger.debug(s"Simulation start at $startTime")
+  }
 
   /** end simulator. */
-  def finish(): Unit
+  private[chiseltest] def finish(): Unit = {
+    logger.debug(s"Simulation end at $startTime")
```
ducky64, line 42: I'm pretty sure this code is just straight up wrong.

Also should be orthogonal.

src/main/scala/chiseltest/exceptions.scala
```diff
@@ -17,3 +17,5 @@ class TimeoutException(message: String) extends ChiselTestException(message)
 
 // when interfacing with the testdriver before stepping the clock after moving to an earlier region
 class TemporalParadox(message: String) extends ChiselTestException(message)
+
+case class TestApplicationException(exitVal: Int, lastMessage: String) extends RuntimeException(lastMessage)
```
ducky64, line 21: Does it really make sense to have the last message here - or is it even guaranteed that the last message is useful? It might make more sense to just generically define this as a message (like the above exceptions), and whatever generates this exception should try to put something useful there?

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -81,13 +97,8 @@ class ThreadedBackend(annotations: AnnotationSeq) extends BackendInterface with
...
     val mainThread = new TesterThread(
       () => {
-
-        /** @todo make user be able to set his with his own init logic,
-          *       for example reset, local last state or etc.
-          */
-        simulatorInterface.poke("reset", 1)
-        simulatorInterface.step(1)
-        simulatorInterface.poke("reset", 0)
+        simulatorInterface.start()
```
ducky64, line 100: I think this should happen in the main thread instead of the tester thread.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -188,18 +199,11 @@ class ThreadedBackend(annotations: AnnotationSeq) extends BackendInterface with
...
       }
       val trace = new Throwable
-      val expectStackDepth = trace.getStackTrace.indexWhere(ste =>
-        ste.getClassName == "chiseltest.package$testableData" && ste.getMethodName == "expect"
-      ) + 1
-      require(
-        expectStackDepth != 0,
-        s"Failed to find expect in stack trace:\r\n${trace.getStackTrace.mkString("\r\n")}"
+      val fullStackTraces = (trace.getStackTrace ++ getParentTraceElements).filterNot(ste =>
+        ste.getClassName.startsWith("chiseltest.internal.")
```
ducky64, line 203: This significantly changes the behavior - this is filtering out, instead of filtering all before or after. The criteria is also looser. Both are subtle, but makes the code more brittle.

Yes, I know that under the old one, changes elsewhere means this also needs to be changed too. But we have tests to catch that, and I'd prefer code errs on the stricter and safer side.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -877,4 +881,88 @@ class ThreadedBackend(annotations: AnnotationSeq) extends BackendInterface with
     }
     processThread(currentThread.get)
   }
+
+  def getClock(clk: Clock): Boolean =
+    simulatorInterface.peek(dataNames(clk)) match {
+      case Some(x) if x == BigInt(1) => true
+      case _                         => false
+    }
```
ducky64, line 889: As mentioned in a previous phase, these should fail noisily. Similarly below.

src/main/scala/chiseltest/package.scala
```diff
@@ -233,6 +232,9 @@ package object chiseltest {
     Context().doTimescope(() => contents)
   }
 
+  /** get current elapsedTime in nano second. */
+  def elapsedNanoSeconds: Long = Context().simulatorInterface.elapsedNanoSeconds()
```
ducky64, line 236: I don't think this should be part of the general public API.

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
```diff
@@ -7,79 +7,64 @@ import java.io.{ByteArrayOutputStream, File, PrintStream}
...
-    val output = outputStream.toString
+    val coverage = new File(targetDir + s"${java.io.File.separator}logs${java.io.File.separator}coverage.dat")
     coverage.exists() should be(true)
-    output.contains("--coverage-toggle") should be(true)
-    output.contains("--coverage-line") should be(false)
-    output.contains("--coverage-user") should be(false)
+    outputStream.toString
+  }
+
+  it should "allow specifying toggle coverage for Verilator" in {
```
ducky64, line 30: Instead of commenting out tests they should be marked as ignored.

src/test/scala/chiseltest/tests/FaultLocatorTest.scala
```diff
@@ -39,13 +39,13 @@ class FaultLocatorTest extends AnyFlatSpec with ChiselScalatestTester with Match
...
-        c.in.bits.poke(false.B)  // Have this be a data failure only
+        c.in.bits.poke(false.B) // Have this be a data failure only
         c.out.expectDequeueNow(true.B)
       }
     }
     // Only check the filename to avoid this being too brittle as implementation changes
-    exc.failedCodeFileNameAndLineNumberString.get should startWith ("DecoupledDriver.scala:")
-    exc.getMessage should include regex ("""\(lines in FaultLocatorTest\.scala:[^\)]*43.*\)""")
+    exc.failedCodeFileNameAndLineNumberString.get should startWith("DecoupledDriver.scala:")
+    exc.getStackTrace.mkString("\n") should include regex ("""\(FaultLocatorTest\.scala:[^\)]*43.*\)""")
```
ducky64, line 48: TODO see if this behaves properly in ScalaTest.

ducky64 (changes requested):
I'm getting this odd feeling of deja vu that some of these comments I'm making here we've already discussed prior during calls or in previous iterations of this PRs. Just because it's a different PR doesn't mean the issues don't still need to be addressed.

I have significant issues with how you've defined `SimulatorInterface`, details in the line-by-line comments. This really seems like the kind of thing that should be discussed beforehand, probably even at the concept stage before you've started writing lots of code around it - because otherwise (and as we're seeing now) you might have a lot of code to re-write.

src/main/scala/chiseltest/backends/SimulatorBackend.scala
```diff
@@ -1,13 +1,28 @@
...
+
 package chiseltest.backends
 
 import firrtl.AnnotationSeq
 import logger._
 
 import scala.sys.process.ProcessLogger
 
+/** [[SimulatorBackend]] is used to generate a binary executable.
+  * It should will compile FIRRTL to a [[SimulatorInterfaceAnnotation]].
```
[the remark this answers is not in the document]
sequencer: will fix.

src/main/scala/chiseltest/backends/SimulatorBackend.scala
```diff
@@ -1,13 +1,28 @@
...
+  * [[SimulatorInterfaceAnnotation]] will be used by [[SimulatorInterface]] to enable communication between generated executable and ChiselTest.
+  */
 trait SimulatorBackend extends LazyLogging {
-  def processLogger(foutFunc: String => Unit = out => logger.info(out),
-                    ferrFunc: String => Unit = out => logger.error(out)
-                   ): ProcessLogger = ProcessLogger(foutFunc, ferrFunc)
-  def compileDut(annotations: AnnotationSeq): AnnotationSeq
+
+  /** [[processLogger]] is used to convert subprocess to [[LazyLogging.logger]].
```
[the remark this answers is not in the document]
sequencer: will fix.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
+import firrtl.annotations.NoTargetAnnotation
+import logger.LazyLogging
+
+/** annotation to store [[SimulatorInterface]], which will be used in [[chiseltest.internal.ThreadedBackend]]. */
+case class SimulatorInterfaceAnnotation(interface: SimulatorInterface) extends NoTargetAnnotation
 
-/** API to Simulator. */
-trait SimulatorInterface {
+/** Standard API which should be implemented by a [[SimulatorInterface]], */
+trait SimulatorInterface extends LazyLogging {
```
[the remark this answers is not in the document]
sequencer: will fix.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
 
   /** start time of this test. */
-  final val startTime: Long = System.nanoTime()
+  final private[chiseltest] lazy val startTime: Long = System.nanoTime()
```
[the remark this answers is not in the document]
sequencer: will fix.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
```diff
@@ -1,30 +1,46 @@
...
 
-  /** how many nano seconds has elapsed since this test start. */
-  def elapsedNanoSeconds(): Long = System.nanoTime() - startTime
+  /** start time of this test. */
+  final private[chiseltest] lazy val endTime: Long = elapsedNanoSeconds()
```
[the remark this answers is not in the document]
sequencer: > Is this even accurate or functionally correct?

This will be evaluated at `finish`. so it's correct.

> This depends on `endTime` being only called at the end of a test - so at the least it's super brittle?

I think it's a internal API. I will make it more safe.

> Also should be orthogonal functionality.

will fix.

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
sequencer: will fix.

src/main/scala/chiseltest/package.scala
sequencer: I think maybe it can be an API addition. we can discuss later.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
sequencer: I'll add an additional API to enable/disable this.
In the future, it's possible to peek a non-exist(constant-propagated) target inside Module.

ducky64: Can you give me a use case where you would want to peek a non-existent target and where some arbitrary result would be useful?

Like, I can see the utility of maybe `peekOption -> Option[BigInt]`, and punt to the upper layer to deal with what happens on a None / nonexistent wire, but I don't think (and I feel strongly about this, unless you have solid arguments the other way) anything should ever create results out of thin air, because this kind of non-intuitive data (even if expedient now) can lead down a fruitless debugging rabbit hole.

And even with peekOption, I'm not sure what would be guidelines for when to use it vs. when not to, and I'd prefer for APIs not to be introduced until we have a reasonable use case for them.

I also feel strongly against a stateful API to toggle this kind of behavior.

sequencer: For example: a API design here:
https://github.com/sequencer/diplomatictester/blob/4284f6919426186cc4d3e752bb0578d3104ce3cd/src/Driver.scala#L99
But `corrupt` is eliminated during constant propagation.
And if it directly fails:
https://github.com/sequencer/diplomatictester/blob/4284f6919426186cc4d3e752bb0578d3104ce3cd/tests/src/dutIOSpec.scala#L46
won't be able to run.

sequencer: The key issue of this is: If we are peeking/poking a constant-propagated target, what behavior should be?
Or we can `dontTouch` those signal explicitly, and throw it like you said.

ducky64: I would say the behavior should be to crash. Sure, a zero might work in the use case you have now, but might not work in another use case - so it's a bad idea to hardcode values created out of nothing. With a crash, at least it's noisy and the user gets a stack trace indicating what failed, and can add the needed dontTouch annotations. Compared to creating values - the user might just see a failed test, with an assertion that might trigger until much later (so nothing hinting at the true root cause), with RTL that is actually good, and might not even think about needing the debug the test infrastructure itself.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
sequencer: Fixed.

src/main/scala/chiseltest/backends/SimulatorInterface.scala
sequencer: Fixed

src/main/scala/chiseltest/backends/SimulatorInterface.scala
sequencer: fixed

src/main/scala/chiseltest/package.scala
sequencer: API removed

sequencer:
I'm going to create a commit with all fixes about this review.
It should happen later today.

ducky64:
@sequencer what's going on with this refactor? Are things still a work-in-progress, or has this been abandoned? If the former, let me know how I can help speed things along so other changes aren't blocked; if the latter then 🤷 we should at least clarify and get some closure.

edwardcwang:
Didn't he already reply to all the code review issues that came up? I see the commit above.

ducky64:
I think we discussed this during a meeting, and I think the resolution was to avoid incidental refactors so the diff doesn't bloat, and that not everything needed to be an annotation or serializable. I haven't seen any movement since, though.

Since this was a one-of meeting (instead of the regularly recurring chisel-dev), I don't know if there was a notes doc. A quick search didn't surface anything.

sequencer:
In our last meeting, IIRC, these PRs are rejected since it's hard to review.
I was busy with other thing recently, I think since this week I can spare 1 or 2 hours per day to rebase and sort my commits into a series of meaningful PRs and we can discuss after next dev-meeting.

ducky64:
Yes, I recall these PRs were rejected because of feature creep too.

The original post was more regarding whether you were working on a refactor for dependencyapi that didn't have all the feature / 'cleanup' creep - because my understanding from that meeting was that you were going to do a minimal set of changes to migrate to dependencyapi.

Whether you accomplish that be rebasing these commits or starting from 'scratch' (perhaps copying over lines / files to another branch) is up to you, but please smaller, self-contained, single-feature PRs...

ekiwi:
This PR has gotten stale, so I will close is for now. Feel free to re-open.

PR #261, opened 2021-02-02
Use only a single version of scalatest

chick:
Fixes problems found in sbt and intellij when using two different versions


ducky64 (approved):
This is fine. Note that this can potentially require users to rewrite code, since ScalaTest 3.1 refactored some imports with deprecation, and ScalaTest 3.2 removed those deprecations.

PR #262, opened 2021-02-13
Chisel 3.5: MultiIOModule is now just Module

ekiwi:
This leaves us with a single deprecation warning when compiling `chiseltest`.

### DIFF
--- a/src/main/scala/chisel3/tester/package.scala
+++ b/src/main/scala/chisel3/tester/package.scala
@@ -57,7 +57,7 @@ package object tester {
     def sanitizeFileName(name: String): String = chiseltest.experimental.sanitizeFileName(name)
 
     object TestOptionBuilder {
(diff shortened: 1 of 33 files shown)


ducky64:
Code looks fine, but appears tests are failing because it's still picking up old Module?

ekiwi:
> Code looks fine, but appears tests are failing because it's still picking up old Module?

I think it has something to do with Scala 2.11. Maybe we stopped publishing 2.11 snapshots of Chisel/firrtl? @jackkoenig

jackkoenig:
> > Code looks fine, but appears tests are failing because it's still picking up old Module?
> 
> I think it has something to do with Scala 2.11. Maybe we stopped publishing 2.11 snapshots of Chisel/firrtl? @jackkoenig

We did for Chisel master yeah, so we need to drop it here.

ekiwi:
Now it passes :tada:

ducky64 (approved):

PR #263, opened 2021-02-15
build: match chisel3 behavior by only building for scala 2.12.13

ekiwi:
Since Chisel 3.5 recently dropped Scala 2.11 support, there is no point in keeping it in chiseltest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .settings
 .idea
 *.iml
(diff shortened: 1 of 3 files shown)


ekiwi:
This PR is a prereq for: https://github.com/ucb-bar/chisel-testers2/pull/262

ducky64 (approved):
lgtm

Issue #212, opened 2020-10-10
RFC: re-using compiled testbenches


ducky64:
This seems to be the most in-demand feature request...

Here, I'll sketch out several ideas with trade-offs, and we'll see if we can figure out how useful each one is. Note: these are API sketches for now, no guarantees these are implement-able.

## Simple precompiled units
This one is pretty basic, provide an alternate API that separates the `test(...) { c => ... }` into two parts: elaboration / compilation, which returns an object, and the test body, which can be invoked on the object, multiple times, with different tests.

Main downside: you need a handle to the object to run tests on it, this could make it more complicated when you have several tests across different files that could share the same object. Possibly worked around with global variables, or a library around that mechanism.

Briefly discussed in #106

Mockup:
```scala
class BasicTest extends FlatSpec with ChiselScalatestTester {
  val precompiled = build(new StaticModule(42.U))  // same interface as module construction for test(...)

  it should "test a thing" in {
    precompiled.test { c =>
      c.out.expect(42.U)
    }
  }

  it should "test it again w/o recompiling" in {
    precompiled.test { c =>
      c.out.expect(42.U)
    }
  }
}

class AnotherTest extends FlatSpec with ChiselScalatestTester {
  val precompiled = build(new StaticModule(42.U))  // can't see inside BasicTest, so need to re-create it
  ...
}
```

## Separation of interface (Chisel Circuit) and backend
A variation on the above, this further separates the `build(...)` into two parts: one part that would provide the Chisel elaborated Module as the interface, and another part that provides the backend. 

Two potential uses are:
1. Manually load a backend from disk (eg, a precompiled Verilator binary, or pre-elaborated FIRRTL file for treadle), to avoid the recompilation time. The main difference from above is that the precompiled units can persist across JVM invocations. This would be an advanced user API, because it relies on the elaborated Module interface being in-sync with the loaded backend, and the loaded backend being up-to-date (if it's a Chisel design). All that being said, I'm not sure how often you want to run a different test suite where you haven't updated the RTL.
2. Testing non-Chisel RTL (by using a Chisel Module shim or BlackBox), possibly including post-syn netlists. One issue with using Module shims is that some features of ChiselTest rely on circuit data, eg checking combinational paths for inter-thread safety, or (in the hopefully-near-future) checking clock sources to get clocks on wires.

Possible solution to #34 

Mockup:
```scala
val simulator = VerilatorBackend.loadFrom(...)

[...]

PR #267, opened 2021-02-23
Add binary reuse to Verilator tests via annotation

ryan-lund:
Adds a CachingAnnotation which enables binary reuse across verilator test runs if high FIRRTL, elaborated annotations, and underlying versions match. In testing reduces the VerilatorBasicTests runtime from 35 seconds on first run to 5 seconds on second (cached) run.

Addresses: https://github.com/ucb-bar/chisel-testers2/issues/106 and https://github.com/ucb-bar/chisel-testers2/issues/212

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -72,6 +72,7 @@ libraryDependencies ++= Seq(
   "edu.berkeley.cs" %% "treadle"       % defaultVersions("treadle"),
   "org.scalatest"   %% "scalatest"     % "3.2.8",
   "com.lihaoyi"     %% "utest"         % "0.7.9",
+  "org.json4s" %% "json4s-jackson" % "3.6.11",
   "org.scala-lang"   % "scala-reflect" % scalaVersion.value,
   compilerPlugin("edu.berkeley.cs" % "chisel3-plugin"    % defaultVersions("chisel3") cross CrossVersion.full)
 ) ++ {
--- a/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
+++ b/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
@@ -1,16 +1,24 @@
 package chiseltest.legacy.backends.verilator
 
-import java.io.{File, FileWriter}
+import java.io.{BufferedReader, File, FileReader, FileWriter}
 
 import chiseltest.backends.BackendExecutive
 import chiseltest.internal._
 import chisel3.{assert, Module}
 import chisel3.experimental.DataMirror
+import chisel3.internal.firrtl.Circuit
 import chisel3.stage.{ChiselCircuitAnnotation, ChiselStage}
-import firrtl.annotations.ReferenceTarget
+import firrtl.annotations.{JsonProtocol, ReferenceTarget}
+import firrtl.ir.StructuralHash
 import firrtl.stage.RunFirrtlTransformAnnotation
 import firrtl.transforms.CombinationalPath
 import firrtl.util.BackendCompilationUtilities
+import java.security.MessageDigest
+import javax.xml.bind.DatatypeConverter
+
+import org.json4s._
+import org.json4s.jackson.Serialization
+import org.json4s.jackson.Serialization.{read, write}
 
 object VerilatorExecutive extends BackendExecutive {
   import firrtl._
@@ -27,6 +35,147 @@ object VerilatorExecutive extends BackendExecutive {
       case "clock" => "clock"
       case _ =>
         s"${component.module}.${component.name}"
+    }
+  }
+
+  def getCachedSim[T <: Module](
+    dut:            T,
+    circuit:        Circuit,
+    elaboratedAnno: AnnotationSeq,
+    targetDir:      String
+  ): Option[BackendInstance[T]] = {
+    if (elaboratedAnno.contains(CachingAnnotation)) {
+      val highFirrtlAnnos = (new ChiselStage).run(
+        elaboratedAnno :+ RunFirrtlTransformAnnotation(new HighFirrtlEmitter)
+      )
+
+      val rawFirrtl = highFirrtlAnnos.collect { case EmittedFirrtlCircuitAnnotation(e) => e.value }
+        .map(l => firrtl.Parser.parse(l))
+        .head
+
+      val sortedModuleHashStrings = new firrtl.stage.transforms.Compiler(targets = firrtl.stage.Forms.HighForm)
+        .transform(firrtl.CircuitState(rawFirrtl, Seq()))
+        .circuit
+        .modules
+        .map(m => m.name -> StructuralHash.sha256WithSignificantPortNames(m))
+        .sortBy(_._1)
+        .map(_._2.hashCode().toString) // TODO: Replace .hashCode().toString with .str
+
+      val sortedModuleMessageDigest = MessageDigest
+        .getInstance("SHA-256")
+        .digest(
+          sortedModuleHashStrings
+            .toString()
+            .getBytes("UTF-8")
+        )
+
+      val moduleHashHex = DatatypeConverter
+        .printHexBinary(sortedModuleMessageDigest)
+
+      val moduleHashFile = new File(targetDir, "module.hash")
+
+      val oldModuleHashHex = if (moduleHashFile.exists()) {
+        new BufferedReader(new FileReader(moduleHashFile)).readLine()
+      } else {
+        " "
+      }
+
+      val moduleWriter = new FileWriter(moduleHashFile)
+      moduleWriter.write(moduleHashHex)
+      moduleWriter.close()
+
+      // Hash the deterministic elements of elaboratedAnno for comparison, this should
+      // only catch if coverage or vcd flags change
+      val elaboratedAnnoString = elaboratedAnno.toSeq.filter {
+        case _: chisel3.stage.ChiselCircuitAnnotation | _: chisel3.stage.DesignAnnotation[T] |
+            _: firrtl.options.TargetDirAnnotation =>
+          false
+        case _ => true
+      }
+        .map(anno => anno.toString)
+      val firrtlInfoString = Seq(firrtl.BuildInfo.toString)
+      val systemVersionString = Seq(System.getProperty("java.version"), scala.util.Properties.versionNumberString)
+
+      val elaboratedAnnoHashHex = DatatypeConverter
+        .printHexBinary(
+          MessageDigest
+            .getInstance("SHA-256")
+            .digest(
+              (elaboratedAnnoString ++ firrtlInfoString ++ systemVersionString).toString
+                .getBytes("UTF-8")
+            )
+        )
+
+      val annoHashFile = new File(targetDir, "anno.hash")
+
+      val oldElaboratedAnnoHashHex = if (annoHashFile.exists) {
+        new BufferedReader(new FileReader(annoHashFile)).readLine
+      } else {
+        " "
+      }
+
+      val elaboratedAnnoWriter = new FileWriter(annoHashFile)
+      elaboratedAnnoWriter.write(elaboratedAnnoHashHex)
+      elaboratedAnnoWriter.close()
+
+      if (moduleHashHex == oldModuleHashHex && elaboratedAnnoHashHex == oldElaboratedAnnoHashHex) { // Check if we have built a simulator with matching module and annotations
+        val portNames = DataMirror
+          .modulePorts(dut)
+          .flatMap { case (name, data) =>
+            getDataNames(name, data).toList.map {
+              case (p, "reset") => (p, "reset")
+              case (p, "clock") => (p, "clock")
+              case (p, n)       => (p, s"${circuit.name}.$n")
+              //          case (p, n) => (p, s"$n")
+            }
+          }
+          .toMap
+
+        implicit val formats = DefaultFormats
+
+        val pathsJson = new BufferedReader(new FileReader(new File(targetDir, "paths.json"))).readLine()
+        val paths = read[Seq[CombinationalPath]](pathsJson)
+
+        val pathsAsData =
+          combinationalPathsToData(dut, paths, portNames, componentToName)
+
+        val commandJson = new BufferedReader(new FileReader(new File(targetDir, "command.json"))).readLine()
+        val command = read[Seq[String]](commandJson)
+
+        val coverageAnnotations = AnnotationSeq(
+          JsonProtocol.deserialize(new File(targetDir, "coverageAnnotations.json"))
+        )
+
+        return Some(new VerilatorBackend(dut, portNames, pathsAsData, command, targetDir, coverageAnnotations))
+      }
+    }
+    None
+  }
+
+  def cacheSim[T <: Module](
+    paths:               Seq[CombinationalPath],
+    command:             Seq[String],
+    coverageAnnotations: AnnotationSeq,
+    elaboratedAnno:      AnnotationSeq,
+    targetDir:           String
+  ): Unit = {
+    if (elaboratedAnno.contains(CachingAnnotation)) {
+      implicit val formats = DefaultFormats
+
+      // Cache paths as json as they depend on the compiledAnnotations
+      val pathsWriter = new FileWriter(new File(targetDir, "paths.json"))
+      pathsWriter.write(write(paths))
+      pathsWriter.close()
+
+      // Cache command
+      val commandWriter = new FileWriter(new File(targetDir, "command.json"))
+      commandWriter.write(write(command))
+      commandWriter.close()
+
+      // Cache coverage annotations
+      val coverageAnnotationsWriter = new FileWriter(new File(targetDir, "coverageAnnotations.json"))
+      coverageAnnotationsWriter.write(JsonProtocol.serialize(coverageAnnotations.toSeq))
+      coverageAnnotationsWriter.close()
     }
   }
 
@@ -47,6 +196,11 @@ object VerilatorExecutive extends BackendExecutive {
     val elaboratedAnno = (new chisel3.stage.phases.Elaborate).transform(annotationSeq :+ generatorAnnotation)
     val circuit = elaboratedAnno.collect { case x: ChiselCircuitAnnotation => x }.head.circuit
     val dut = getTopModule(circuit).asInstanceOf[T]
+
+    getCachedSim(dut, circuit, elaboratedAnno, targetDir) match {
+      case Some(sim) => return sim
+      case _         =>
+    }
 
     // Create the header files that verilator needs
     CopyVerilatorHeaderFiles(targetDir)
@@ -144,6 +298,8 @@ object VerilatorExecutive extends BackendExecutive {
       combinationalPathsToData(dut, paths, portNames, componentToName)
 
     val coverageAnnotations = VerilatorCoverage.collectCoverageAnnotations(compiledAnnotations)
+
+    cacheSim(paths, command, coverageAnnotations, elaboratedAnno, targetDir)
 
     new VerilatorBackend(dut, portNames, pathsAsData, command, targetDir, coverageAnnotations)
   }
--- a/src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
+++ b/src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
@@ -1,0 +1,214 @@
+// SPDX-License-Identifier: Apache-2.0
+package chiseltest.experimental.tests
+
+import chisel3._
+import chiseltest.tests.{PassthroughModule, StaticModule}
+import chiseltest._
+import chiseltest.experimental.TestOptionBuilder._
+import chiseltest.experimental.sanitizeFileName
+import chiseltest.internal.{CachingAnnotation, VerilatorBackendAnnotation, WriteVcdAnnotation}
+import org.scalatest._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import java.io.File
+
+class VerilatorCachingTests extends AnyFlatSpec with ChiselScalatestTester with Matchers {
+  behavior of "Testers2 with Verilator and caching"
+
+  val annos = Seq(VerilatorBackendAnnotation, CachingAnnotation)
+
+  it should "test static circuits" in {
+    test(new StaticModule(42.U)).withAnnotations(annos) { c =>
+      c.out.expect(42.U)
+    }
+  }
+
+  it should "fail on poking outputs" in {
+    assertThrows[UnpokeableException] {
+      test(new StaticModule(42.U)).withAnnotations(annos) { c =>
+        c.out.poke(0.U)
+      }
+    }
+  }
+
+  it should "fail on expect mismatch" in {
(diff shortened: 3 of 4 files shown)


ducky64 (commented):
This looks fine at a high level, though I'd prefer it be explicitly in the experimental or advanced user package, because tests now depend on state on disk. This also differs from the proposal in the linked issues, which would provide users with an explicit reusable handle to a compiled simulator (essentially separating the compilation and simulation stages), instead of implicitly re-using cached binaries.

For implementation: unless this is only for super advanced users running a *nix-like OS, consider seeing if there is a Java/Scala API for file operations, so it's more platform-neutral.

I'd also prefer that there's no printing, since unit tests should not require users to check the console output (and console spam also isn't helpful for regressions),

ducky64:
Also, consider adding documentation on the mechanism of action, such as in the scaladoc for the annotation.

build.sbt
```diff
@@ -16,7 +16,9 @@ resolvers ++= Seq(
 
 libraryDependencies ++= Seq(
   "org.scalatest" %% "scalatest" % "3.2.2",
-  "com.lihaoyi" %% "utest" % "latest.integration"
+  "com.lihaoyi" %% "utest" % "latest.integration",
+  "com.github.plokhotnyuk.jsoniter-scala" %% "jsoniter-scala-core"   % "2.6.4",
+  "com.github.plokhotnyuk.jsoniter-scala" %% "jsoniter-scala-macros" % "2.6.4" % "provided"
```
ekiwi, line 21: I don't think we should be adding a second json library to our dependencies.
Could you check to see how hard it would be to use [json4s](https://github.com/json4s/json4s) instead?
That is the library used by `firrtl` and thus already pulled into `chiseltest`.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -48,6 +52,74 @@ object VerilatorExecutive extends BackendExecutive {
...
+      val highFirrtlPath = s"${targetDir}/${circuit.name}.hi.fir"
+      val oldHighFirrtlPath = s"${targetDir}/${circuit.name}.hi.fir.old"
+      s"mv ${highFirrtlPath} ${oldHighFirrtlPath}".!
+
+      (new ChiselStage).run(
+        elaboratedAnno :+ RunFirrtlTransformAnnotation(new HighFirrtlEmitter)
+      )
+
+      // Diff high FIRRTL and delete old file
+      val diffExit = s"diff ${highFirrtlPath} ${oldHighFirrtlPath}".!
```
ekiwi, line 66: Instead of relying on GNU tools here, you could try to use the hashing mechanism that we already have for firrtl:
https://github.com/chipsalliance/firrtl/blob/edb91f7bc613026f824519786c3ce25740bb21c3/src/main/scala/firrtl/ir/StructuralHash.scala#L29

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -48,6 +52,74 @@ object VerilatorExecutive extends BackendExecutive {
...
+
+      // Hash the deterministic elements of elaboratedAnno for comparison
+      // currently since this is guaranteed to be in Verilator and have the same test run dir, this should
+      // only catch if coverage or vcd flags change
+      val elaboratedAnnoHash = elaboratedAnno
+        .toSeq
+        .filter(anno => !List("chisel3.stage.ChiselCircuitAnnotation", "chisel3.stage.DesignAnnotation")
+          .contains(anno.getClass.toString.split(" ").last))
+        .map(anno => anno.hashCode)
+      val firrtlInfoHash = Seq(firrtl.BuildInfo.hashCode)
```
ekiwi, line 80: This isn't really good enough since `hashCode` is not a cryptographic hash, it is only really meant to quickly discover that two things are _not_ equal, but it isn't good enough for equality.
The easiest approach here might be to serialize all annotations to a JSON string and then use a Java `MessageDigest` to calculate a sha256 hash.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -48,6 +52,74 @@ object VerilatorExecutive extends BackendExecutive {
...
+      val systemVersionHash = Seq(System.getProperty("java.version"), scala.util.Properties.versionNumberString)
+        .map(version => version.hashCode)
```
ekiwi, line 82: again, `hashCode` is wrong

ekiwi (changes requested):
I think the general idea is good. The two things that I would like to ask you to change are:
1. use json4s instead of jsoniter
2. rethink how you do the hash, see my comments

ekiwi:
@jackkoenig:  We could add the build time to the firrtl build info: https://github.com/sbt/sbt-buildinfo

oharboe:
This would speed up our tests tremendously. 

Note that when running tests in parallel, Verilator and Chisel elaboration are the memory hogs. 

There should be a way to run a pre-processing stage that does Verilator and Chisel elaboration in parallel to the extent that there's memory available. Normally with Chisel I run out of memory before cores when running things in parallel. Especially in the cloud where there's ca. 1gByte per core for e.g. Google Cloud Build machine profiles.

build.sbt
ryan-lund: For now I've kept jsoniter because I've found its feature set is better set-up for serializing generic objects without having to write extra converters. (I.E as it is I can serialize the CombinationPath object without the need to write anything special). I haven't been able to replicate that with json4s but I'll keep investigating.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
ryan-lund: Changed to use your structural hash (thank you!) and replaced Unix operations in general with native Scala / Java file operations.

ryan-lund:
The largest update I've made in the latest push is a change to the way in which the cache is stored. The goals of these changes are to allow tests with different names or classes but identical DUTs to match the same cache entry. To achieve this, a new directory called cache_dir is created. In this director, the generated Verilog and Verilator sources are stored in a directory named test_run_dir/cache_dir/<high firrtl hash>/<annotations hash>. Thus any DUTs that has the same high firrtl hash and the same annotations will load the same sources. This source location is then simlinked to the specific test directory so that a user can view the sources used for testing (i.e. test_run_dir/should_<test name>/sources -> est_run_dir/cache_dir/<high firrtl hash>/<annotations hash>). Waveform files are always written to the specific test directory so that they are not overwritten by the results of un-related tests.

As before, these features are only turned on when the CachingAnnotation is supplied. I've tested the VerilatorBasicTests without the annotation and verified that the sources generate in the typical  test_run_dir/should_<test name> location.

I will also write up detailed documentation as per Ducky's comment about scaladocs.

ekiwi:
> In this director, the generated Verilog and Verilator sources are stored in a directory named test_run_dir/cache_dir//.

Quick question: How are you making sure that there are no race conditions between two different threads that both try to compile the same DUT?

build.sbt
ekiwi: I don't want to sign off on adding a new dependency. Removing that dependency later will be a great pain, since users might start depending on it. If @ducky64  is OK, with adding it, then we can. However, if it up to just me, then I believe that we should not add this dependency.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
```diff
@@ -190,8 +190,8 @@ int main(int argc, char **argv, char **env) {
 #endif
 #if VM_COVERAGE
     VL_PRINTF(\"Writing Coverage..\");
-    Verilated::mkdir("$targetDir/logs");
-    VerilatedCov::write("$targetDir/logs/coverage.dat");
+    Verilated::mkdir("$coverageDir");
+    VerilatedCov::write("$coverageDir/coverage.dat");
```
ekiwi, line 194: How is this change related to this PR? Did you add this by accident?

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -38,16 +45,101 @@ object VerilatorExecutive extends BackendExecutive {
...
+      val highFirrtlAnnos = (new ChiselStage).run(
+        elaboratedAnno :+ RunFirrtlTransformAnnotation(new HighFirrtlEmitter)
+      )
+
+      val rawFirrtl = highFirrtlAnnos
+        .collect { case EmittedFirrtlCircuitAnnotation(e) => e.value}
+        .map(l => firrtl.Parser.parse(l))
+        .head
+
+      val moduleHashString = DatatypeConverter
```
ekiwi, line 72: please split this up with some more named `val`s

build.sbt
ducky64: I generally would prefer to not add dependencies, and additionally prefer to standardize libraries across the chisel projects where possible. But if there's some strong rationale for a new library (or if we want to standardize on a new library project-wide, perhaps without necessarily migrating old code), it's worth discussing.

jackkoenig: We're fairly close to being able to build Chisel and FIRRTL without any runtime reflection except for generic serialization and deserialization of annotations. This isn't like, an important goal, but it would be cool to be able to emit ScalaJS as well as create GraalVM Native images. jsoniter appears to have better support for non-JVM targets than json4s, so it's worth considering using it throughout instead. This is a high code migration cost though since some users have written custom [de]serializers.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
ryan-lund: Done!

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
ryan-lund: This was an intentional change. I realized when making the more global prototype that the coverage directory would always default to $targetDir/logs. I felt that it would be a more useful API to let to caller of the C++ harness generator fully specify the path to a coverage directory.

ekiwi: I believe this should go into a separate PR

ryan-lund: Removed

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -48,6 +55,115 @@ object VerilatorExecutive extends BackendExecutive {
...
+      }
+
+      val moduleWriter = new FileWriter(moduleHashFile)
+      moduleWriter.write(moduleHashHex)
+      moduleWriter.close()
+
+      // Hash the deterministic elements of elaboratedAnno for comparison, this should
+      // only catch if coverage or vcd flags change
+      val elaboratedAnnoString = elaboratedAnno.toSeq
+        .filter(anno =>
```
ekiwi, line 101: could you replace this with something like?:

```.scala
.filter {
  case _ : chisel3.stage.ChiselCircuitAnnotatio | _ :  chisel3.stage.DesignAnnotation | _ : firrtl.options.TargetDirAnnotation => false
  case _ => true
}
```

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -48,6 +55,115 @@ object VerilatorExecutive extends BackendExecutive {
     val circuit = elaboratedAnno.collect { case x: ChiselCircuitAnnotation => x }.head.circuit
     val dut = getTopModule(circuit).asInstanceOf[T]
 
+    if (elaboratedAnno.contains(CachingAnnotation)) {
```
ekiwi, line 58: Since the whole thing has gotten quite long, could you please move this into its own function?

src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
```diff
@@ -0,0 +1,130 @@
...
+      c.io.out.expect(0.U)
+
+      c.reset.poke(false.B)
+      c.io.in.poke(43.U)
+      c.clock.step()
+      c.io.out.expect(43.U)
+    }
+  }
+
+  it should "test waveform generation" in {
```
ekiwi, line 122: This should do two checks:

1. check that the VCD is not there before `test`
2. check that the VCD now exists after `test`

src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
```diff
@@ -0,0 +1,130 @@
...
+import chisel3._
+import chiseltest.tests.{PassthroughModule, StaticModule}
+import chiseltest._
+import chiseltest.experimental.TestOptionBuilder._
+import chiseltest.internal.{CachingAnnotation, VerilatorBackendAnnotation, WriteVcdAnnotation}
+import org.scalatest._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class VerilatorCachingTests extends AnyFlatSpec with ChiselScalatestTester with Matchers {
```
ekiwi, line 13: Please add a test with multiple calls to `test` inside a single scalatest unit.

ekiwi (changes requested):

src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
ryan-lund: Done, added multiple calls with the same module in one case and different modules in another case.

src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala
ryan-lund: Updated this test to match the VCD test in OptionsPassingTest.scala

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
ryan-lund: Moved this into `getCachedSim`. Moved caching generation into `cacheSim`.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
ryan-lund: Done

ekiwi (approved):
Thanks @ryan-lund !

Let's get this in :shipit:

oharboe:
@ryan-lund Thanks! I'm looking forward to testing this! :-)

ekiwi:
> @ryan-lund Thanks! I'm looking forward to testing this! :-)

Yeah, please do let us know if you find any issues with this. That is why we are publishing the `SNAPSHOT` releases, so that the community can test things early and identify any issues that we missed in our testing.

oharboe:
@ekiwi ETA for SNAPSHOT? Is there any doc to go with this or do I just run a test and time it?

ekiwi:
>  ETA for SNAPSHOT? Is there any doc to go with this or do I just run a test and time it?

It is already out: https://github.com/ucb-bar/chisel-testers2/runs/2634441405
Just make sure to clear your coursier cache. sbt might only be checking for new releases every 24h or so.

You can have a look at Ryan's tests to see how to make use of the new feature:
https://github.com/ryan-lund/chisel-testers2/blob/master/src/test/scala/chiseltest/experimental/tests/VerilatorCachingTests.scala

oharboe:
I'm running into build problems when I upgrade...
```
Symbol 'type org.scalatest.FlatSpec' is missing from the classpath.\nThis symbol is required by 'class chisel3.iotesters.ChiselFlatSpec'.\nMake sure that type FlatSpec is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.\nA full rebuild may help if 'ChiselFlatSpec.class' was compiled against an incompatible version of org.scalatest.
```

```
val defaultVersions = Map(
  "chisel3" -> "3.5-SNAPSHOT",
  "chisel-iotesters" -> "1.5-SNAPSHOT",
  "chiseltest" -> "0.5-SNAPSHOT"
)

```

oharboe:
Deleted some tests and now I can build, but I get this error message:

```
[error] java.lang.NoClassDefFoundError: com/vladsch/flexmark/ast/Node
[error]         at org.scalatest.tools.ReporterFactory.createHtmlReporter(ReporterFactory.scala:194)
[error]         at org.scalatest.tools.ReporterFactory.getReporterFromConfiguration(ReporterFactory.scala:245)
[error]         at org.scalatest.tools.ReporterFactory.$anonfun$createReportersFromConfigurations$1(ReporterFactory.scala:254)
[error]         at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:273)
```

oharboe:
I finally ran into a problem that appears to be related to this new feature:

```
 error: redefinition of ‘void chisel_insertp(const char*, const char*, const char*, const char*, const char*, int, const char*, int, const char*, const string&, const char*, const char*, const char*, const char*, const char*, const char*)’
   34 | static void chisel_insertp(
      |             ^~~~~~~~~~~~~~
```

oharboe:
Tried a different smaller test:

```
[info] ascenium.tiles.ALUTests *** ABORTED *** (1 second, 633 milliseconds)
[info]   java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
[info]   at chiseltest.legacy.backends.verilator.VerilatorExecutive$.getCachedSim(VerilatorExecutive.scala:73)
[info]   at chiseltest.legacy.backends.verilator.VerilatorExecutive$.start(VerilatorExecutive.scala:200)
[info]   at chiseltest.defaults.package$.createDefaultTester(defaults.scala:24)
[info]   at chiseltest.ChiselScalatestTester$TestBuilder.apply(ChiselScalatestTester.scala:33)
[info]   at ascenium.tiles.ALUTests.$anonfun$new$5(ALUTests.scala:171)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   ...
[info]   Cause: java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter
[info]   at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
[info]   at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
[info]   at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
[info]   at chiseltest.legacy.backends.verilator.VerilatorExecutive$.getCachedSim(VerilatorExecutive.scala:73)
[info]   at chiseltest.legacy.backends.verilator.VerilatorExecutive$.start(VerilatorExecutive.scala:200)
[info]   at chiseltest.defaults.package$.createDefaultTester(defaults.scala:24)
[info]   at chiseltest.ChiselScalatestTester$TestBuilder.apply(ChiselScalatestTester.scala:33)
[info]   at ascenium.tiles.ALUTests.$anonfun$new$5(ALUTests.scala:171)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
```

oharboe:
J! First life! Was able to get caching for a small test. :-)

Some notes:

The first time I run the testsuite, I have many tests that use the same Verilator binary, I get no caching.

The second time, all tests are cached. 

My expectation is that during the first run of the testsuite, the Verilator binaries should be cached.

```
rm -rf test_run_dir
sbt "test:testOnly *ALUT*"
--- no Verilator binaries cached ---
sbt "test:testOnly *ALUT*"
--- all Verilator binaries cached ---
```

oharboe:
I'm happy to set up an interactive Google Meet, if that's easier to figure out what's going on. I can't share our code easily...

ekiwi:
> I finally ran into a problem that appears to be related to this new feature:

This is probably related to an earlier change. Could you show us the output of `verilator --version`?

oharboe:
> > I finally ran into a problem that appears to be related to this new feature:
> 
> This is probably related to an earlier change. Could you show us the output of `verilator --version`?

This is the Ubuntu 20.04 verilator:

```
$ verilator --version
Verilator 4.028 2020-02-06 rev v4.026-92-g890cecc1
```

ekiwi:
Awesome `4.0281 is still supported: https://github.com/ucb-bar/chisel-testers2/runs/2634410004?check_suite_focus=true

Does the project where you encounter this error have any special features that we might be missing in our CI? Like does it have black boxes, are you changing verilator flags? Anything out of the ordinary?

oharboe:
> Does the project where you encounter this error have any special features that we might be missing in our CI? Like does it have black boxes, are you changing verilator flags? Anything out of the ordinary?

I tried to find out what adds the verilator `--coverage-user`  argument, but I can't find it...

Some extra environment variables I add, but no VerilatorFlags().

```
// faster tests by using ccache for Verilator
envVars in Test := Map(
  "OBJCACHE" -> "ccache",
  "MAKEFLAGS" -> "VM_PARALLEL_BUILDS=0 -j8"
)
```

ekiwi:
> I tried to find out what adds the verilator `--coverage-user` argument, but I can't find it...

That is now done by default since we added the `cover` statement to firrtl and we want that to be supported out of the box.

kammoh:
Use of `javax.xml.bind.DatatypeConverter` makes this incompatible with JRE versions > 8. For example on OpenJDK 11 I get this:
```
*** ABORTED ***
  java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
  at chiseltest.legacy.backends.verilator.VerilatorExecutive$.getCachedSim(VerilatorExecutive.scala:73)
  at chiseltest.legacy.backends.verilator.VerilatorExecutive$.start(VerilatorExecutive.scala:200)
  at chiseltest.defaults.package$.createDefaultTester(defaults.scala:24)
  at chiseltest.ChiselScalatestTester$TestBuilder.apply(ChiselScalatestTester.scala:33)
```
Going through the code I'm not sure why conversion to hex string is even necessary? As far as I can see the hash is written and read in a file, so why not use a binary file?

kammoh:
What do you about adding JDK 11 (at least) to the CI tests?

kammoh:
Which tests check effectiveness of `CachingAnnotation`? #309 fixes the issue I had with Java 11, but caching doesn't seem to be working: `verilated` folder is apparently deleted each time `test` is run and everything is compiled from scratch. At the same time all tests are passing. Is it because this PR is not compatible with #303 or am I doing something wrong here?

PR #276, opened 2021-03-19
ci: install verilator instead of using a container

ekiwi:
This gives us a more up to date version of verilator and will make things easier to reproduce for people using the latest Ubuntu LTS.


ekiwi:
This also speeds up CI by ~1min 20sec

jackkoenig (approved):

PR #280, opened 2021-03-24
Add Support for using Vec literals

chick:
Depends on chipsalliance/chisel3#1834

Extends literal handling API with
- package object chiseltest
  - implicit class testableVec[T <: Vec[_]](x: T)
    - def pokePartial(value: T): Unit
    - def expectPartial(value: T): Unit

Include VecLiteralsSpec for tests

### DIFF
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -5,6 +5,7 @@ import chiseltest.internal._
 import chisel3._
 import chisel3.experimental.{DataMirror, Direction, EnumType, FixedPoint, Interval}
 import chisel3.experimental.BundleLiterals._
+import chisel3.experimental.VecLiterals._
 import chisel3.util._
 
 /** Basic interfaces and implicit conversions for testers2
@@ -52,7 +53,53 @@ package object chiseltest {
       }.foreach { case (k, v) =>
         v match {
           case record: Record => record.expectPartial(value.elements(k).asInstanceOf[Record])
-          case data:   Data   => data.expect(value.elements(k))
+          case data: Data => data.expect(value.elements(k))
+        }
+      }
+    }
+  }
+
+  implicit class testableVec[T <: Vec[_]](x: T) {
+
+    /** Poke the given signal with a [[Vec.litValue()]]
+      * Literals of this Record can be instantiated with
+      * {{{
+      *   someVec.Lit(_.elements("foo") -> 4.U)
+      * }}}
+      * `pokePartial` will only poke [[Input]] signals of `x`,
+      * and elements of `x` which contain no literal will be ignored.
+      */
+    def pokePartial(value: T): Unit = {
+      require(DataMirror.checkTypeEquivalence(x, value), s"Vec type mismatch")
+      x.getElements.zipWithIndex.filter { case (v, index) =>
+        DataMirror.directionOf(v) != ActualDirection.Output && {
+          value.getElements(index) match {
+            case _: T => true
+            case data: Data => data.isLit()
+          }
+        }
+      }.foreach { case (v, index) =>
+        v match {
+          case vec: T => vec.pokePartial(value.getElements(index).asInstanceOf[T])
+          case data: Data => data.poke(value.getElements(index))
+        }
+      }
+    }
+
+    /** Check the given signal with a [[Vec.litValue()]];
+      * elements of `x` which contain no literal will be ignored.
+      */
+    def expectPartial(value: T): Unit = {
+      require(DataMirror.checkTypeEquivalence(x, value), s"Vec type mismatch")
+      x.getElements.zipWithIndex.filter { case (v, index) =>
+        value.getElements(index) match {
+          case _: T => true
+          case d: Data => d.isLit()
+        }
+      }.foreach { case (v, index) =>
+        v match {
+          case vec: T => vec.expectPartial(value.getElements(index).asInstanceOf[T])
+          case data: Data => data.expect(value.getElements(index))
         }
       }
     }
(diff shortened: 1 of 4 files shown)


src/main/scala/chiseltest/package.scala
```diff
@@ -52,7 +53,53 @@ package object chiseltest {
...
+      x.getElements.zipWithIndex.filter { case (v, index) =>
+        DataMirror.directionOf(v) != ActualDirection.Output && {
+          value.getElements(index) match {
+            case _: T => true
+            case data: Data => data.isLit()
+          }
+        }
+      }.foreach { case (v, index) =>
+        v match {
+          case vec: T => vec.pokePartial(value.getElements(index).asInstanceOf[T])
```
ducky64, line 83: Does it make sense to match on T, as opposed to Vec?
Also, would this support Vecs inside Records and vice versa?

src/main/scala/chiseltest/package.scala
```diff
@@ -114,6 +161,12 @@ package object chiseltest {
           x.poke(value)
         }
       }
+      case (x: Vec[_], value: Vec[_]) => {
+        require(DataMirror.checkTypeEquivalence(x, value), s"Vec[_ type mismatch")
```
ducky64, line 165: Something seems off with that string...

src/main/scala/chiseltest/package.scala
```diff
@@ -143,6 +196,11 @@ package object chiseltest {
         }.toSeq
         chiselTypeOf(x).Lit(elementValueFns: _*).asInstanceOf[T]
       }
+      case (x: Vec[_]) =>
+        val elementValueFns = x.getElements.zipWithIndex.map { case (elt: Data, index: Int) =>
```
ducky64, line 200: Does this discard the index? At least, do you want to add an assertion, if not propagate the index through?

src/main/scala/chiseltest/package.scala
```diff
@@ -175,6 +233,12 @@ package object chiseltest {
           x.expectWithStale(value, message, stale)
         }
       }
+      case (x: Vec[_], value: Vec[_]) => {
+        require(DataMirror.checkTypeEquivalence(x, value), s"Vec[_ type mismatch")
```
ducky64, line 237: String also looks funny

src/test/scala/chiseltest/tests/VecLiteralsSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+package chiseltest.tests
+
+import chisel3._
+import chisel3.experimental.VecLiterals._
+import chiseltest._
+import org.scalatest._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class VecLiteralsSpec extends AnyFreeSpec with ChiselScalatestTester with Matchers {
```
ducky64, line 12: Do you also want to add a Record-in-Vec or vice versa test case?

ducky64 (changes requested):
Looks mostly fine, with a few concerns, inline in review.

src/main/scala/chiseltest/package.scala
chick: Because of the parameterization of `Vec`, this seems to be the easiest correct syntactic construction
It does support rec in vec and vice versa

src/main/scala/chiseltest/package.scala
chick: Fixed typo

src/main/scala/chiseltest/package.scala
chick: Removed unneeded `zipWithIndex`

src/main/scala/chiseltest/package.scala
chick: Fixed message typo

src/test/scala/chiseltest/tests/VecLiteralsSpec.scala
chick: Added
- `Vec` of `Vec`
- `Bundle` of `Vec`
- `Vec` of `Bundle`
- `Bundle` of `Bundle` (this one in BundleLiteralsSpec)

chick:
I believe all requests have been addressed, This will not pass CI until the 
chipsalliance/chisel3#1834 is merged

src/main/scala/chiseltest/package.scala
ducky64: But will this recursively pokePartial with an inner Bundle? It looks like it will do a normal strict poke.
Which is fine for now, to allow a recursive pokePartial later would be strictly an API expansion I think,

ducky64 (approved):
lgtm with one comment that does not need to be addressed.

Sorry about the delay...

ekiwi:
@chick Does this need updates in order to work with the version of VecLiterals that got merged with chisel?

chick:
@ekiwiI was going to say I don't think so but I restarted and got some errors still, I was hoping chisel3 3.5-SNAPSHOT containing vec lit support would be publish on merge and it would fix the failures here. I'll check it out locally, hopefully it's something simple

chick:
@ekiwi The chisel3 with vec literals has been published in 3.5-SNAPSHOT and it works when I use that locally to run chiseltest's test suite, but somehow on CI here it does not see the Vec literals API stuff. I've restarted several times to no avail. Same problem is breaking the verilator regressions here. I'll keep poking at it. If you have a chance to take a look maybe you will see something I am missing

ekiwi:
@chick maybe we are doing some mis-guided caching. Let me see.

PR #274, opened 2021-03-18
Chisel Coverage Support

ekiwi:
_Note_: This is a technology preview.

**Major Features**
- chiseltest backends now return a `AnnotationSeq` from the `test(...){ ... }` call which contains information that can be used to generate a coverage report
- the Verilator backend now automatically extracts coverage counts for all `cover` statements in the firrtl and returns them as a `TestCoverage` annotation
- the tradle backend now provides coverage information from treadle through a `TestCoverage` annotation
- the `LineCoverage.scala` file shows how one could implement line coverage on top of this infrastructure:
  - the `LineCoveragePass` instruments the high firrtl
  - the `LineCoverage.processCoverage` function turn the generic `TestCoverage` annotation into a struct containing the line coverage
  - the `LineCoverage.textReport` function generates a textual report from that information
- the `CoverageScanChainPass` replaces each cover statements with a saturating counter, all counters are connected in a scan chain
  - this will be the start of `firesim` support for universal Chisel Coverage
 
**Known TODOs**
- [ ] make sure that the Verilator backend coverage does not break when a Verilog Black box has its own coverage statements
- [ ] add `ConditionCoverage` object and instrumentation pass
- [ ] add `FsmCoverage` object and instrumentation pass
- [ ] experiment with a mux path coverage pass
- [ ] multi-clock support (will require a better Clock/Reset inference pass in firrtl)
- [ ] Reset/Clock coverage support (how often has each clock been toggled? how often has each reset been release?)

**Known Limitations**
- no support for merging coverage across design changes (Adam had some ideas on how to do this reliably and it should be possible to implement them on top of the feature.)

**Example Line Coverage Report**

```
Line | Cnt | Source
----------------------------------------------------------------------------------
   1 |     | // SPDX-License-Identifier: Apache-2.0
   2 |     | 
   3 |     | package chiseltest.coverage
   4 |     | 
   5 |     | import chisel3._
   6 |     | 
   7 |     | class Test1Module(withSubmodules: Boolean = false) extends Module {
   8 |     |   val a = IO(Input(UInt(3.W)))
   9 |     |   val b = IO(Output(UInt(3.W)))
  10 |     | 
  11 |   6 |   b := 0.U // line 5
  12 |     | 
  13 |   6 |   when(a === 4.U) {
  14 |   0 |     b := 1.U
  15 |     |   }
  16 |     | 
  17 |   6 |   when(5.U < a) {
  18 |   1 |     b := 2.U
  19 |     |   }.otherwise {
[...]

PR #281, opened 2021-03-25
Report TestCoverage from Treadle and Verilator backends

ekiwi:
This PR changes the `test(...){...}` method to return a firrtl `AnnotationSeq`.
When using Treadle or Verilator as backends, this sequence will contain a `TestCoverage` annotation with coverage counts for every `cover` statement in the circuit.

This PR contains about 50% of the changes from https://github.com/ucb-bar/chisel-testers2/pull/274.
I tried to isolate the minimal changes necessary in `chiseltest` to support automated coverage, without adding any actual coverage passes just yet.


src/main/scala/chiseltest/legacy/backends/verilator/VerilatorBackend.scala
```diff
@@ -246,5 +251,15 @@ class VerilatorBackend[T <: Module](
 
       simApiInterface.finish() // Do this to close down the communication
     }
+    collectCoverageAnnotations()
```
chick, line 254: Nit: wondering if this should be `convertCoverageAnnotations` or similar to make clear it's distinction from other `collectCoverageAnnotations` methods

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
```diff
@@ -0,0 +1,85 @@
...
+  val x = IO(Input(SInt(8.W)))
+  val y = IO(Input(SInt(8.W)))
+  val inCircle = IO(Output(Bool()))
+  val inRectangle = IO(Output(Bool()))
+  val radius = 100
+  inCircle := (x * x + y * y) <= (radius * radius).S
+  inRectangle := (x <= radius.S && x >= -radius.S) && (y <= radius.S && y >= -radius.S)
+  chisel3.experimental.verification.cover(inCircle)
+  chisel3.experimental.verification.cover(inRectangle)
+}
```
chick, line 85: missing newline

chick (approved):
Looks great to me. I have some comments in there, not any show stoppers. Probably good to get a cursory sign off from @ducky64

src/main/scala/chiseltest/coverage/ModuleInstancesPass.scala
```diff
@@ -0,0 +1,54 @@
...
+    val children = InstanceKeyGraph(state.circuit).getChildInstances.toMap
+    val topInstance = InstanceKey("", state.circuit.main)
+    val topChildren = children(topInstance.module)
+    val instances = topInstance +: topChildren.flatMap(onInstance("", _, children))
+    val instanceToModule = instances.toList.map(i => i.name -> i.module)
+    val anno = ModuleInstancesAnnotation(instanceToModule)
+    state.copy(annotations = anno +: state.annotations)
+  }
+
+  /** expands the instance name to its complete path (module the main module) */
```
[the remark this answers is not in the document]
ekiwi: Good catch, I will try to fix this.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -118,7 +119,9 @@ trait BackendInstance[T <: Module] extends BackendInterface {
     * Takes the test function, which takes the module used as the testing interface.
     * TesterContext setup is done externally.
     *
+    * @return coverage annotations
```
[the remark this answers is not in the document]
ekiwi: Yes. The idea here is that we want to reduce the number of annotations that are kept around during the test in order to consume less main memory. Like if we just retained all annotations without any filtering, there could be a huge custom annotation that is kept alive and eats up all your RAM.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorBackend.scala
ekiwi: OK. I will rename it.

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
ekiwi: I was confused for a bit why `scalafmt` didn't fix this. Turns out it does not run on the `test` directory, only on `main`.

src/main/scala/chiseltest/coverage/ModuleInstancesPass.scala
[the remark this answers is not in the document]
ekiwi: Turns out what I meant to say was _modul**o** the main module_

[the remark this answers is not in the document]
ekiwi: But I will try to rephrase it to be less confusing

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorBackend.scala
ekiwi: fixed

src/main/scala/chiseltest/coverage/ModuleInstancesPass.scala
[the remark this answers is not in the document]
ekiwi: fixed

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
ekiwi: fixed

src/main/scala/chiseltest/backends/treadle/TreadleBackend.scala
```diff
@@ -14,7 +16,8 @@ class TreadleBackend[T <: Module](
   val dut:                T,
   val dataNames:          Map[Data, String],
   val combinationalPaths: Map[Data, Set[Data]],
-  tester:                 TreadleTester)
+  tester:                 TreadleTester,
+  coverageAnnotations:    AnnotationSeq)
```
ducky64, line 20: Does it make sense for something more general here? (eg, pass in all annotations)

src/main/scala/chiseltest/coverage/Coverage.scala
```diff
@@ -0,0 +1,20 @@
+package chiseltest.coverage
+
+import firrtl._
+import firrtl.annotations.{Annotation, NoTargetAnnotation}
+
+/** Coverage counts returned from the simulator interface. */
+case class TestCoverage(counts: List[(String, Long)]) extends NoTargetAnnotation
```
ducky64, line 7: Can you define (in a comment) what the strings mean?

src/test/scala/chiseltest/coverage/CompilerTest.scala
```diff
@@ -0,0 +1,36 @@
...
+import firrtl.options.TargetDirAnnotation
+import org.scalatest.TestSuite
+
+import java.io.File
+
+/** Base trait for tests that need to compile a circuit and inspect the resulting firrtl / Verilog  */
+trait CompilerTest extends ChiselScalatestTester { this: TestSuite =>
+  protected def annos: AnnotationSeq = Seq()
+
+  protected def compile[M <: Module](gen: => M, target: String, a: AnnotationSeq = List(), ll: String = "warn"): (String, AnnotationSeq) = {
```
ducky64, line 19: Can you write a comment to define the return string-type? It appears it's any of FIRRTL or Verilog?

src/test/scala/chiseltest/coverage/Test1Module.scala
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
```
ducky64, line 1: It doesn't appear this is being used other than in SimulatorCoverageTest, can it be put in that same file?

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    val cov = getCoverage(r)
+
+    // the cover point in the main module are not prefixed, but the one in the child module are
+    assert(cov.keys.toList.sorted == List("c0.cover_0", "c1.cover_0", "cover_0"))
+
+    // since we executed one step (+ the implicit reset) and all inputs are zero by default, we expect the count to be 3
+    assert(cov("cover_0") == 3)
+  }
+
+  it should "allow us to estimate pi" in {
```
ducky64, line 47: A short blurb about the intent of this test could be helpful. (what is this test testing? or is this just a catch-all integration test - still useful, but clarity is nice)

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    val r = test(new Test1Module(withSubmodules = true)).withAnnotations(noAutoCov) { dut =>
+      dut.clock.step()
+    }
+    val cov = getCoverage(r)
+
+    // the cover point in the main module are not prefixed, but the one in the child module are
+    assert(cov.keys.toList.sorted == List("c0.cover_0", "c1.cover_0", "cover_0"))
+
+    // since we executed one step (+ the implicit reset) and all inputs are zero by default, we expect the count to be 3
+    assert(cov("cover_0") == 3)
```
ducky64, line 44: Do you also want to check the value for the child modules too?

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
```diff
@@ -23,9 +23,8 @@ class VerilatorCoverageTests extends AnyFlatSpec with ChiselScalatestTester with
     }
     val output = outputStream.toString
     coverage.exists() should be(true)
-    output.contains("--coverage-toggle") should be(true)
-    output.contains("--coverage-line") should be(false)
-    output.contains("--coverage-user") should be(false)
+    assert(output.contains("--coverage-toggle"))
```
ducky64, line 26: Can you do something like `output should contain(...)`?
(and similarly below)

ducky64 (commented):
Overall looks sane, with one concern about committing to an AnnotationSeq return type for test invocations. Also stylistic suggestions.

I didn't look over the code in the legacy directory.

src/main/scala/chiseltest/backends/treadle/TreadleBackend.scala
ekiwi: My assumption was that this is a semi-private API, i.e., most users aren't really expected to manually instantiate it. Thus changing this to be more broad in the future shouldn't be too big of a problem. For now this only contains the coverage annotations, so the name is very descriptive wrt. the current implementation.

src/main/scala/chiseltest/coverage/Coverage.scala
ekiwi: Will do.

src/test/scala/chiseltest/coverage/CompilerTest.scala
ekiwi: Yes.

src/test/scala/chiseltest/coverage/Test1Module.scala
ekiwi: I would rather not, since once I add the line coverage, I will depend on exact line number in this file for my tests.
So yes, in isolation it would make sense to move this into its own file, but considering the other additions I want to make, it wouldn't be a good idea.

The one thing I can offer is to make a separate package, something along the lines of `chiseltest.coverage.circuits` or `chiseltest.coverage.examples` or `chiseltest.coverage.benchmarks` or similar.

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
ekiwi: I am personally not a fan of the matchers and the original code was obviously using them wrong which meant that the error message I got from scalatest was not helpful at all. If you want me to, I can try to change it back to using matchers _correctly_, however I am not 100% certain that I know how to.

src/main/scala/chiseltest/ChiselScalatestTester.scala
```diff
@@ -22,7 +22,7 @@ trait ChiselScalatestTester extends Assertions with TestSuiteMixin with TestEnvI
       sanitizeFileName(scalaTestContext.value.get.name)
     }
 
-    def apply(testFn: T => Unit): Unit = {
+    def apply(testFn: T => Unit): AnnotationSeq = {
```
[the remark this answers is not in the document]
ekiwi: I was trying to make the minimal changes, so just directly returning an `AnnotationSeq` was the most straight forward thing to do. I can add a wraper object with a `.getAnnotationSeq` method if you would like me to.

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
ekiwi: Good catch!

src/test/scala/chiseltest/coverage/SimulatorCoverageTest.scala
ekiwi: added a comment

src/main/scala/chiseltest/backends/treadle/TreadleBackend.scala
ducky64: Yep, that's fine.

src/test/scala/chiseltest/coverage/Test1Module.scala
ducky64: I see. The main benefit of having it in the same file would be marginally more convenient readability. But a comment describing the line stability concerns would be great.

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
ducky64: No strong preference as long as we're somewhat locally consistent.

src/main/scala/chiseltest/ChiselScalatestTester.scala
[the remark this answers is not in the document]
ducky64: I think having a wrapper class would be preferred to keep flexibility.

src/test/scala/chiseltest/experimental/tests/VerilatorCoverageTests.scala
ekiwi: I converted it to matchers.
The right matcher is `output should include(...)` since `contain` only seems to work on sequences, but not on a String.

src/main/scala/chiseltest/ChiselScalatestTester.scala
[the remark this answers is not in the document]
ekiwi: I am creating a `TestResults` wrapper class. Do you have a good idea where it should live?
I was going to put it into the `Testers2.scala`, but then I realized that it is in the `internal` package and
we would probably want `TestResults to just live in `chiseltest.`.

src/test/scala/chiseltest/coverage/Test1Module.scala
ekiwi: added a comment

src/main/scala/chiseltest/TestResult.scala
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest
+
+import firrtl.AnnotationSeq
+
+class TestResult(annotations: AnnotationSeq) {
```
ekiwi, line 7: @ducky64 Is there a good file to put the `TestResult` class in? (Instead of creating a new file)

ducky64: I think having its own file is fine? It is (or I guess could be) a significant API, even though the implementation is super short?

ekiwi: I am happy with that.
Any good ideas for a non-obvious scala doc?
I was trying, but all I could come up with was:
> Contains the results of a test run.

ducky64: Perhaps also mention that this is meant for future expansion of the test return API, though there's not much in it right now.

ducky64: (I could totally see someone asking "wait why does this even exist")

ekiwi: I will add a plain comment, since this is mostly developer and not user facing.

ekiwi: added the comment

ducky64 (approved):
lgtm!

PR #289, opened 2021-04-06
Better Verilator CI

ekiwi:


src/test/scala/chiseltest/backends/verilator/VerilatorCoverageTest.scala
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest.backends.verilator
+
+import chiseltest.coverage.SimulatorCoverageTest
+import chiseltest.internal.VerilatorBackendAnnotation
+
+class VerilatorCoverageTest extends SimulatorCoverageTest("Verilator", VerilatorBackendAnnotation) {}
```
ducky64, line 8: Should this really be in its own file?
(or can you refactor it so that the treadle coverage tests are grammatically parallel, eg in its own file with the base class in a separate file?)

src/test/scala/chiseltest/backends/verilator/VerilogBlackBoxTest.scala
```diff
@@ -0,0 +1,45 @@
...
+}
+
+class VerilogBlackBoxTest extends AnyFlatSpec with ChiselScalatestTester {
+  behavior of "Verilator Backend"
+
+  val annos = Seq(VerilatorBackendAnnotation)
+
+  it should "support Verilog black boxes" in {
+    test(new BlackBoxAdderWrapper).withAnnotations(annos) { dut =>
+      dut.io.a.poke(1.U)
```
ducky64, line 42: are you intentionally not checking the outputs?

ducky64 (commented):
two comments, otherwise looks fine.

src/test/scala/chiseltest/backends/verilator/VerilogBlackBoxTest.scala
ekiwi: Yes. This just checks that the backend does not crash.

src/test/scala/chiseltest/backends/verilator/VerilatorCoverageTest.scala
ekiwi: Treadle is currently our default simulator, i.e. all tests are run with Treadle by default, so keeping the treadle test under the "default" package made sense to me. I wanted to move this test into the `verilator` package so that it would be caught by the `sbt testOnly ...` command I use to get all the Verilator tests.

If, in the future, we start changing tests so that we can run them with various backends, we could add a test option for that and keep them in a single file. For now, however, this is a one off test.

src/test/scala/chiseltest/backends/verilator/VerilogBlackBoxTest.scala
ducky64: I'd prefer it either checks the output, removes the functional lines in the blackbox, or has a comment to indicate that this is only a basic non-crash sanity check. Otherwise, the dead code could be a source of confusion?

ekiwi: I will add the output check. Thanks for the feedback!

ducky64 (approved):
lgtm

Issue #282, opened 2021-03-30
Verilator tests failing


hansemandse:
Some Verilator tests of mine are failing after updating to 0.5-SNAPSHOT. The tests in question are `neuroproc.unittests.TrueDualPortMemoryTester` and `neuroproc.unittests.TrueDualPortFIFOTester` available [here](https://github.com/hansemandse/KWSonSNN/tree/master/src/test/scala/neuroproc/unittests). Both tests are working in 0.3.2.

The error message is:
```[info]   java.lang.RuntimeException: Failed to find line `VL_COVER_INSERT(count32p,  "filename",filenamep,  "lineno",lineno,  "column",column,` in test_run_dir/True_dual_port_memory_should_work_in_both_Chisel_and_Verilog/VTrueDualPortMemoryVerilog.cpp.```

I am using:
- chisel3 3.5-SNAPSHOT ([7b8fbbe](https://github.com/chipsalliance/chisel3/commit/7b8fbbe8913a2eb92f50948dc2be50dc482fb58d))
- chisel-testers2 0.5-SNAPSHOT (5d5071a)

`verilator --version` => Verilator 3.916 2017-11-25 rev verilator_3_914-65-g0478dbd

ekiwi:
Thanks for the report.

To support the new cover points feature that will be introduced in the x.5 release, we will release we will be requiring at least verilator `4.026`.
The immediate cover statements that the firrtl compiler may produce now are only supported starting with [Verilator 4.026](https://www.veripool.org/news/266-Verilator-Verilator-4-026-Released).

However, I will try to see if there is a way to get older verilator versions supported as long as you are not using any cover points.

ekiwi:
Kevin's TODO:
- [ ] do not try to patch coverage when there are no cover points in the design
- [ ] detect verilator version and show a warning if it is too old

ekiwi:
> I'm seeing this issue with the latest Verilator from stable branch (Verilator 4.200 2021-03-12 rev v4.200-16-gf0d66453) as well.

Thanks for pointing that out. Do you happen to have a small test case that we could add to CI?
Our CI is currently using `Verilator 4.028 2020-02-06 rev v4.026-92-g890cecc1` and unfortunately did not catch this regression.

kammoh:
Sure, but I'm guessing the issue is not really related to Verilator version. 
Shouldn't coverage pass and data analysis only be evoked if some coverage annotations are actually included in the elaborated design?
[...]

PR #290, opened 2021-04-06
verilator: support coverage for version 4.028 - 4.108

ekiwi:
This PR is intended to fix https://github.com/ucb-bar/chisel-testers2/issues/282


ekiwi:
Supporting Verilator `4.2xx` will have to wait for a little because that might require changes to the CPP harness generator in order to accommodate the new context object: https://github.com/verilator/verilator/commit/2cad22a22af92c9fb7748cf7062dab798102b16a

README.md
```diff
@@ -130,6 +130,31 @@ To dump VCDs (into the test_run_dir subfolder) using sbt:
 testOnly chiseltest.tests.BasicTest -- -DwriteVcd=1
 ```
 
+## Verilator Backend
+
+To run your tests using the [verilator](https://www.veripool.org/wiki/verilator) simulator:
+
+1. Ensure that you have a _supported version_ of verilator installed.
+   You can [consult the CI script](https://github.com/ucb-bar/chisel-testers2/blob/master/.github/workflows/test.yml#L38)
```
ducky64, line 138: Is it worth adding how to set up a supported version of Verilator? This could be linking to the Chisel readme if it's there.

Also, instead of linking the CI scripts, perhaps just link the Chisel readme section with the version, since that's more readable and perhaps more authoritative?

README.md
```diff
@@ -130,6 +130,31 @@ To dump VCDs (into the test_run_dir subfolder) using sbt:
...
+   to find out which versions you can use.
+2. Enable the Verilator Backend by adding the right annotation to your tests:
```
ducky64, line 140: Perhaps clarify that this is on a per-`test()` basis (as opposed to project-wide or file-wide)?

README.md
```diff
@@ -130,6 +130,31 @@ To dump VCDs (into the test_run_dir subfolder) using sbt:
...
+
+```.scala
+// [...]
+import chiseltest.experimental.TestOptionBuilder._
+import chiseltest.internal.VerilatorBackendAnnotation
```
ducky64, line 145: Can this be moved (or duplicated) to experimental, so people aren't importing internal items?

ducky64 (approved):
Some suggestions but otherwise looks good

README.md
ekiwi: I was afraid of having two sources of truth, but I think you are correct wrt. copying it into the readme being much more readable.

README.md
ekiwi: like "enable the Verilator backend for your individual tests by adding the right annotation:" ?

ducky64: Yeah, that's more clear

README.md
ekiwi: I would be happy to. I am not sure what the story around the Verilator backend is supposed to be.
To be honest, I feel like `withAnnotation`, `VerilatorBackendAnnotation` and `TreadleBackendAnnotation` should just be available when you `import chiseltest._`. What do you think?

README.md
ducky64: Yeah, understand the two sources of truth; I'm fine with the version not being duplicated in this document, but I think a link to the Chisel readme section containing the version would be preferable to a link to a CI script, even if the versions are pretty obvious in the CI script. Bonus: Chisel readme actually has setup directions.

README.md
ducky64: Is `withAnnotation` experimental?
But I'm happy with `VerilatorBackendAnnotation` and friends coming with `withAnnotation`.

The story around the Verilator backend is that it lives in legacy because it's largely an imported codebase, that hasn't been reviewed for code quality or style. But as `VerilatorBackendAnnotation` is more an interface class that doesn't have much in the way of implementation, I think that can be closer to mainline.

Not sure why `VerilatorBackendAnnotation` is internal - maybe we forgot about it? Or was there another public-facing API intended for making tests run on Verilator? @chick ?

ekiwi: I think I am going to remove the example from this PR for now. Changing the package of important annotations should be a PR on its own.

PR #303, opened 2021-05-14
Verilator generated artifacts is a subfolder

kammoh:
All binaries, object files, C++ files, Makefiles, etc generated by Verilator are kept in a "verilated" subfolder and deleted before execution of a new test.


ekiwi:
Does this fix the issue you were having?

kammoh:
> Does this fix the issue you were having?

It's hard to say for sure because the issue not happening all the time. I haven't seen it happening so far!

PR #306, opened 2021-05-20
Attempt to provide test of backpressure in fork

chick:
I can't figure it out but `src/test/scala/chiseltest/tests/BackPressureTest.scala` fails it's first
test. I am probably just messing up the fork code, but I am not sure where to go next.
Assistance appreciated

### DIFF
--- a/src/main/scala/chiseltest/DecoupledDriver.scala
+++ b/src/main/scala/chiseltest/DecoupledDriver.scala
@@ -96,6 +96,18 @@ class DecoupledDriver[T <: Data](x: ReadyValidIO[T]) {
       .joinAndStep(getSinkClock)
   }
 
+  /** Use to create back pressure on a Decoupled output
+    */
+  def notReady(): Unit = timescope {
+    x.ready.poke(false.B)
+  }
+
+  /** Use to create back pressure on a Decoupled output
+    */
+  def notValid(): Unit = timescope {
+    x.valid.poke(false.B)
+  }
+
   def expectDequeueNow(data: T): Unit = timescope {
     // TODO: check for init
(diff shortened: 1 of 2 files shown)


kammoh:
I think moving the back-pressure logic to a `timescore` solves the issue. `ready` should also be probably set to `0` at the end of the loop. Here's how I got back-pressure working: https://gist.github.com/kammoh/c078b50193dbbd4e796da9a779c75111

src/test/scala/chiseltest/tests/BackPressureTest.scala
```diff
@@ -0,0 +1,138 @@
...
+        new GcdOutputBundle(16).Lit(_.value1 -> x.U, _.value2 -> y.U, _.gcd -> BigInt(x).gcd(BigInt(y)).U)
+      }
+
+      fork {
+        dut.input.enqueueSeq(inputSeq)
+        println("all data enqueued")
+      }.fork {
+        for(expected <- resultSeq) {
+          dut.output.expectDequeue(expected)
+          dut.output.ready.poke(false.B)
```
ducky64, line 96: I think the issue is the poke here, you have a poke without a gating step right after the inner timescope ends. It's weird?

In any case, the explicit poke should also be unnecessary, because with `initSink`, ready defaults to false when not otherwise driven (for example, in a `expectDequeue`) - one of the design goals of chiseltest was to avoid state errors from needing to manually revert signals, so the `timescope` construct handles it for you.

ducky64 (changes requested):

chick:
Closed. Better possible solution found with PR #326

PR #307, opened 2021-05-21
move verilator caching test to verilator specific directory

ekiwi:
This ensures that it will be included in the verilator regression tests that run on all supported versions.


chick (approved):
lgtm

PR #309, opened 2021-05-22
Fix #267 regression on JVM > 8 

kammoh:
Use of `CachingAnnotation` (introduced by #267) fails on JVM versions newer than 8 as `javax.xml.bind.DatatypeConverter` is no longer available:
```
*** ABORTED ***
  java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
  at chiseltest.legacy.backends.verilator.VerilatorExecutive$.getCachedSim(VerilatorExecutive.scala:73)
  at chiseltest.legacy.backends.verilator.VerilatorExecutive$.start(VerilatorExecutive.scala:200)
  at chiseltest.defaults.package$.createDefaultTester(defaults.scala:24)
  at chiseltest.ChiselScalatestTester$TestBuilder.apply(ChiselScalatestTester.scala:33)
```
This PR:
- Eliminate all usages of `javax.xml.bind.DatatypeConverter`
- Remove redundant conversion of hash digest to string
- Port some usages of `java.io.File` readers/writers to `os.read`/`write`
- Some cleanup, trying more idiomatic Scala
- Fix exception thrown on failure of stale verilated directory


kammoh:
Unfortunately I ended up addressing multiple issues here, but I feel the changes are clear, rather small, and _almost_ related :) Please feel free to break up into multiple PRs if you feel that's a must. Otherwise, ready for review.

src/main/scala/chiseltest/legacy/backends/verilator/Utils.scala
```diff
@@ -419,7 +419,7 @@ private[chiseltest] object verilogToVerilator extends EditableBuildCSimulatorCom
     if (staleObjDri.exists()) {
       println(s"Deleting stale Verilator object directory: $staleObjDri")
       if (!firrtl.FileUtils.deleteDirectoryHierarchy(staleObjDri)) {
-        throw BackendException(s"Deleting $staleObjDri failed!")
+        throw new RuntimeException(s"Deleting $staleObjDri failed!")
```
ekiwi, line 422: Why did you change the exception type here?

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -1,27 +1,26 @@
...
 import firrtl.util.BackendCompilationUtilities
-import java.security.MessageDigest
-import javax.xml.bind.DatatypeConverter
 
 import org.json4s._
 import org.json4s.jackson.Serialization
-import org.json4s.jackson.Serialization.{read, write}
 
 object VerilatorExecutive extends BackendExecutive {
-  import firrtl._
```
ekiwi, line 24: :+1:

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -42,140 +41,115 @@ object VerilatorExecutive extends BackendExecutive {
     dut:            T,
     circuit:        Circuit,
     elaboratedAnno: AnnotationSeq,
-    targetDir:      String
-  ): Option[BackendInstance[T]] = {
-    if (elaboratedAnno.contains(CachingAnnotation)) {
```
ekiwi, line 47: I honestly prefer the `if(....contains(...))` over doing a `collectFirst` with side-effects.

ekiwi (changes requested):
Most changes are great, besides that one block that you had to move because you switched from a `if` to a `collectFirst`. That creates a big diff that is really hard to review and I don't think that change does anything to fix your bug.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
ekiwi: Please revert back to the old structure and try to keep the diff a bit smaller.

src/main/scala/chiseltest/legacy/backends/verilator/Utils.scala
ekiwi: If there is no good reason for doing so, then please revert back.

kammoh: `BackendException` adds the message "Unknown backend: xxx. Backend should be firrtl, verilator, ivl, vcs, or glsim" which is misleading/irrelevant to this situation (directory deletion failed). I had mistakenly used it in #303 and only discovered my mistake while testing this PR.

ekiwi: OK. No need to change this then.

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
kammoh: > I honestly prefer the `if(....contains(...))` over doing a `collectFirst` with side-effects.
Sure I can change it back, but I didn't get the part about side-effect? I think they're both the same in terms of side effects?
I had a hard time figuring out what [ this lonely `None` at #L152](https://github.com/ucb-bar/chisel-testers2/pull/309/files/557c2b27160f3fce0df2c6f60bf72cd2d6b3158e#diff-0af24921e1f6b9e31fdad8acf63750b78887fc046853645e9643fa3e3ba22597L152) (original code) was doing there.

ekiwi: I think functional APIs should be used when you have a side effect free function, when you are doing things that are mostly about their side effects, I prefer to stick to the imperative features in scala.
So I would use `map` for pure functions, `foreach` when you are doing side-effects.

ekiwi: The `None` is there because of the use of `return` in the code. Would probably be nice to do this more like:

```
if(....) {
} else { None }
```

kammoh: `Option.when()` looks neat in these situations, but is only available in 2.13+ (maybe also backported to `scala-collection-compat`?).

kammoh: > ```
>  if(....) {
>    ....
>   Some(...)
> } else { None }
> ```

There's a nested condition (matching of hashes) which makes it difficult to do this cleanly without a `return`.

ekiwi (approved):
LGTM

ekiwi:
Thanks a lot @kammoh

Issue #311, opened 2021-05-22
Problems with --coverage-user and chisel-testers2 master


oharboe:
Is there a test-case 

I've tried this with Verilator 4.028 and 4.202.

This command line creates a single .cpp file that has includes all the modules in the test.

```
/usr/bin/perl /home/oyvind/ascenium/verilator2/install/share/verilator/bin/verilator_includer ... VDoohickeyTestBench__ALL.cpp
```

```
// DESCRIPTION: Generated by verilator_includer via makefile
#define VL_INCLUDE_OPT include
#include "VDoohickeyTestBench.cpp"
#include "VDoohickey_Foo.cpp"
#include "VDoohickey_Bar.cpp"
```

In Verilator 4.028 this creates problems because static insertp() function is duplicated in each of the files above, causing error when __ALLcls.cpp is built.

In Verilator 4.202 there are different problems:

```
VDoohicekyTestBench__Slow.cpp: In function ‘void chisel_insertp(const char*, const char*, const char*, const char*, const char*, int, const char*, int, const char*, const string&, const char*, const char*, const char*, const char*, const char*, const char*)’:
VCacheInterfaceUnitTestBench__Slow.cpp:52:19: error: ‘_insertp’ is not a member of ‘VerilatedCov’
   52 |     VerilatedCov::_insertp(
[lots more errors deleted]
```

oharboe:
There are various problems with Ubuntu 20.04, so I had to create the pull request below to get chisel-testers2 going on Ubuntu 20.04 and with Verilator 4.202:

To reproduce:

1. Checkout https://github.com/ucb-bar/chisel-testers2/pull/313
2. Build Verilator 4.202
3. Run the command below on Ubuntu 20.04:

```
sbt test
```

Snippet of result:

```
[deleted]
In file included from VVerilatorBasicTests_Anon__ALL.cpp:4:
VVerilatorBasicTests_Anon__Slow.cpp: In function ‘void chisel_insertp(const char*, const char*, const char*, const char*, const char*, int, const char*, int, const char*, const string&, const char*, const char*, const char*, const char*, const char*, const char*)’:eltest.tests.ThreadSafetyTest 10s
VVerilatorBasicTests_Anon__Slow.cpp:49:19: error: ‘_insertp’ is not a member of ‘VerilatedCov’
   49 |     VerilatedCov::_insertp(
      |                   ^~~~~~~~
In file included from ./VVerilatorBasicTests_Anon.h:12,
[...]

PR #313, opened 2021-05-22
Verilator 4 202 problems on ubuntu 20 04

oharboe:


ekiwi:
Thanks for trying to reproduce your problem, @oharboe !

It would be great if you could rebase this on top of the latest `master` since we just merged a PR from @kammoh that fixes the OpenJDK11 issue.

@ryan-lund it would be great if you could have a look at this.

oharboe:
verilator 4.202 still fails after rebase on master.

kammoh:
Is this PR adding compatibility with verilator `4.202` (which is currently not supported)? I see `Text file busy` error in the CI sbt test logs, which suggests verilated model was running at the same time another test was building the same executable file, i.e. a race condition. 
Can you try disabling parallel runs? I guess adding `Test / parallelExecution := false` should do.

oharboe:
> Is this PR adding compatibility with verilator `4.202` (which is currently not supported)?
> I see `Text file busy` error in the CI sbt test logs, which suggests verilated model was running at the same time another test was building the same executable file, i.e. a race condition.
> Can you try disabling parallel runs? I guess adding `Test / parallelExecution := false` should do.

No. I previously had to work around limitations that broke Java 11. That's no longer necessary, so I guess this should now be a github issue and not a pull request: Verilator 4.202 doesn't work on master.

ekiwi:
> Verilator 4.202 doesn't work on master.

Oh OK. So Verilator `4.202` is currently not supported: https://github.com/ucb-bar/chisel-testers2#verilator-backend
I welcome PRs to add support, but currently I do not have time to add support myself.

oharboe:
> > Verilator 4.202 doesn't work on master.
> 
> Oh OK. So Verilator `4.202` is currently not supported: https://github.com/ucb-bar/chisel-testers2#verilator-backend
> I welcome PRs to add support, but currently I do not have time to add support myself.

See #311 there is a problem with 4.028 as wella s 4.202, I suspect the same problem, different error messages. 

Now that master works with jdk 11, this branch has served it's purpose. Closing.

PR #315, opened 2021-05-26
tests: fails to build with verilator

oharboe:
Demonstrates Verilator build failure with master.


ekiwi:
I can confirm that this test still fails. Investigating.

ekiwi:
This issue is being addressed in https://github.com/ucb-bar/chisel-testers2/pull/363

Thanks for providing a working example!

PR #317, opened 2021-05-26
Adding support to Verilator 202

kammoh:
Trying to add support for Verilator 4.202 (and hopefully onwards). 

Related issue: #311

### DIFF
--- a/src/main/scala/chiseltest/legacy/backends/verilator/PatchCoverageCpp.scala
+++ b/src/main/scala/chiseltest/legacy/backends/verilator/PatchCoverageCpp.scala
@@ -24,14 +24,21 @@ object PatchCoverageCpp {
   }
 
   private def replaceCoverage(cppFile: Path, lines: Array[String]): Unit = {
-    // we add our code at the beginning of the coverage section
-    val coverageStart = findLine(CoverageStartNeedle, cppFile, lines)
-    lines(coverageStart) += "\n" + CustomCoverInsertCode + "\n"
+    val (verilatorMajor, verilatorMinor) = VerilatorCppHarnessGenerator.verilatorVersion
+    require(verilatorMajor == 4)
 
-    // then we replace the call
-    val call = findLine(CallNeedle, cppFile, lines)
-    val callLine = lines(call).replaceAllLiterally(CallNeedle, CallReplacement)
-    lines(call) = callLine
+    if (verilatorMinor < 200) {
+      // we add our code at the beginning of the coverage section
+      val coverageStart = findLine(CoverageStartNeedle, cppFile, lines)
+      lines(coverageStart) += "\n" + CustomCoverInsertCode + "\n"
+
+      // then we replace the call
+      val call = findLine(CallNeedle, cppFile, lines)
+      val callLine = lines(call).replaceAllLiterally(CallNeedle, CallReplacement)
+      lines(call) = callLine
+    } else {
+      // ???
+    }
   }
 
   private def loadFiles(dir: String): Seq[(Path, Array[String])] = {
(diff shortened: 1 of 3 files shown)


src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
```diff
@@ -40,10 +47,15 @@ object VerilatorCppHarnessGenerator {
...
     val dutVerilatorClassName = "V" + dutName
-    // example version string: Verilator 4.038 2020-07-11 rev v4.038
-    // this will probably break when Verilator hits version 10+, but hopefully we can drop compatibility by then
-    val verilatorVersion = "verilator --version".!!.split(' ')(1).stripLineEnd
-    val verilatorRunFlushCallback = if (verilatorVersion >= "4.038") {
+    val (verilatorMajor, verilatorMinor) = verilatorVersion
+
+    val coverageInit =
+      if (verilatorMajor >= 4 && verilatorMinor >= 202)
+        """|Verilated::defaultContextp()->coveragep()->forcePerInstance(true);
```
ekiwi, line 54: it's awesome that you found that!

src/main/scala/chiseltest/legacy/backends/verilator/PatchCoverageCpp.scala
```diff
@@ -24,14 +24,21 @@ object PatchCoverageCpp {
...
+      // we add our code at the beginning of the coverage section
+      val coverageStart = findLine(CoverageStartNeedle, cppFile, lines)
+      lines(coverageStart) += "\n" + CustomCoverInsertCode + "\n"
+
+      // then we replace the call
+      val call = findLine(CallNeedle, cppFile, lines)
+      val callLine = lines(call).replaceAllLiterally(CallNeedle, CallReplacement)
+      lines(call) = callLine
+    } else {
+      // ???
```
ekiwi, line 40: Could you please add a comment here that states that starting with Verilator 4.2 the whole patch coverage hack is no longer necessary?

ekiwi (changes requested):
Once you add that comment I think this should be ready to merge.

Thanks a lot for your excellent work!

src/main/scala/chiseltest/legacy/backends/verilator/PatchCoverageCpp.scala
kammoh: Sure, that's a good idea. I was thinking of figuring out the right patch for 4.200, but I don't think I have the time for it now.
What do you think about removing `else` and moving the version condition for patch all the way up before [VerilatorExecutive.scala#L249](https://github.com/ucb-bar/chisel-testers2/blob/b07c3650403a4d9e7091c7bde03bae3aa247d2d8/src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala#L249)?

kammoh:
> Once you add that comment I think this should be ready to merge.
> 
> Thanks a lot for your excellent work!

You're welcome! I have to admit I'm not using the coverage feature right now, and solely depend on the available tests to assume nothing is broken. In other words, I have no idea what I'm doing :))

src/main/scala/chiseltest/legacy/backends/verilator/VerilatorCppHarnessGenerator.scala
kammoh: It's on `4.202` release notes with (literally) your name on it! :D

ekiwi:
> You're welcome! I have to admit I'm not using the coverage feature right now, and solely depend on the available tests to assume nothing is broken. In other words, I have no idea what I'm doing :))

I trust my coverage tests, so that you don't have to :)

src/main/scala/chiseltest/legacy/backends/verilator/PatchCoverageCpp.scala
ekiwi: That sounds good to me. In 1-2 years, once there is an Ubuntu LTS release that has Verilator `4.202+` we can the just delete the whole `PatchCoverageCpp.scala`.

ekiwi (approved):
:shipit:

PR #320, opened 2021-05-28
Fix comparison of hashes for `CachingAnnotation`

kammoh:
I had made a stupid mistake in #308 and used `equals` for comparing hashes, which is object comparison not the content, and will always fail on two separate arrays, even if their contents are the same. The original code in #267 also used `==` which I believe is an alias for `equals`, and therefore never worked either.
That probably why I never got `CachingAnnotation` to work :)
It seems there are currently no tests checking for effectiveness of `CachingAnnotation`, or if they are they don't seem to be working.


ekiwi (approved):
I guess we need better tests. Thanks for catching this.

ekiwi:
Please rerun scala format

PR #325, opened 2021-06-09
Defer creation of caching hash file

kammoh:
The caching hash files were creating in the beginning of checks. In case `CachingAnnotation` was used and there were any issues during Verilator cpp generation/build (e.g. problem with Verilator installation or version), verilator generation would never run again (after fixing the installation issue). The user needed to manually delete the previous directory or hash files.
Caching hash is consolidated into a single `caching.hash` file. Should potentially make the cached run a little faster.
Also avoids needless overwrites.


ekiwi (approved):
LGTM. Thanks! :ship:

PR #326, opened 2021-06-10
Demonstrate back-pressure in decoupled interface testing

chick:
Adds a test that illustrates back-pressure input stalling by occasionally inserting
extra steps in enqueue and dequeue blocks

Based on [BootCamp PR 153](https://github.com/freechipsproject/chisel-bootcamp/pull/153)

### DIFF
--- a/src/test/scala/chiseltest/tests/DecoupledBackPressureInputStallTest.scala
+++ b/src/test/scala/chiseltest/tests/DecoupledBackPressureInputStallTest.scala
@@ -1,0 +1,171 @@
+package chiseltest.tests
+
+import chisel3._
+import chisel3.experimental.BundleLiterals.AddBundleLiteralConstructor
+import chisel3.util.{Counter, Decoupled}
+import chiseltest.experimental.TestOptionBuilder.ChiselScalatestOptionBuilder
+import chiseltest.internal.WriteVcdAnnotation
+import chiseltest._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers.be
+import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
+
+import scala.util.Random
+
+class GcdInputBundle(val w: Int) extends Bundle {
+  val value1 = UInt(w.W)
+  val value2 = UInt(w.W)
+}
+
+class GcdOutputBundle(val w: Int) extends Bundle {
+  val value1 = UInt(w.W)
+  val value2 = UInt(w.W)
+  val gcd    = UInt(w.W)
+}
+/**
+  * Compute GCD using subtraction method.
+  * Subtracts the smaller of registers x and y from the larger until register y is zero.
+  * value input register x is then the Gcd
+  * returns a packet of information with the two input values and their GCD
+  */
+class DecoupledGcd(width: Int) extends Module {
+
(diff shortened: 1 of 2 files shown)


chick:
This replaces #306 with a cleaner interface and a
test with decent examples.

ducky64:
Are you sure these are necessary? The current implementation of [`initSource` ](https://github.com/ucb-bar/chisel-testers2/blob/master/src/main/scala/chiseltest/DecoupledDriver.scala#L12) and [`initSink`](https://github.com/ucb-bar/chisel-testers2/blob/master/src/main/scala/chiseltest/DecoupledDriver.scala#L62) set valid and ready to false, which is intended to be the default state unless temporarily overridden by something like `enqueue` or `expectDequeue`.

Also, are you sure about the implementation? The idea of a timescope is that the pokes inside disappear (and revert) at the end of a timescope. As written, it seems you open a timescope, do a poke, and immediately close the timescope - so it effectively shouldn't do anything. Typically the idea with a timescope is that it should do some kind of operation that advances time, like stepping on a block.

Note that this PR seems to follows the common latching pokes style of writing tests, where you would have to remember to poke (for example) valid = false after asserting valid = true for one transaction. In contrast, the chiseltest style provides a default and a time-scoped temporary override, with the system handling the revert for you.

chick:
@ducky64 Several points:
- I could not get this to work without wrapping using `timescope`s in `notReady` and `notValid`
  - I was unable to get the tests in `DecoupledBackPressureInputStallTest` to work using `initSource` and `initSink`
  - Maybe you could provide an example of how you see the API on this.
- Motivation for this was @wunderabt trying to construct a decoupled back-pressure example for the bootcamp
  - They tried using a slower clock in the module to create the back-pressure but that approach has a lot of problems.
- I'm a big believer in TBUINU "the best UI is no UI", it follows
  - Back-pressure should be built-in to the decoupled tools in some way that is clear and painless
  - Maybe `enqueue`, `dequeue` family could have additional arguments, with good defaults.
  - Maybe @ekiwi has some ideas here.
    - We should make it hard to write tests of decoupled interfaces that don't handle back-pressure
- If this turns out to be really complicated I think we should make some small step that will let us get a bootcamp example

chick:
These errors should be fixed by PR #329

ekiwi (approved):
LGTM

src/test/scala/chiseltest/tests/DecoupledBackPressureInputStallTest.scala
```diff
@@ -0,0 +1,153 @@
...
+      x := bundle.value1
+      y := bundle.value2
+      xInitial := bundle.value1
+      yInitial := bundle.value2
+      busy := true.B
+      output.valid := false.B
+    }
+  }
+
+  val cycles = IO(Output(UInt(32.W)))
```
ducky64, line 75: In a proper design, this wouldn't be part of the DUT right? This is pure test instrumentation? Can this be moved into the test harness, or if a cycle count isn't available yet (is it?) can we have a TODO for it?

ducky64 (changes requested):
Mostly looks good, some stylistic issues.

chick:
Closing as this is more of an example than a test. This was prompted by [Bootcamp PR 153](https://github.com/freechipsproject/chisel-bootcamp/pull/153). 153 has been closed as I believe the DecoupledGCD code demonstrates back-pressure handling already

PR #334, opened 2021-07-16
Replace old backend abstraction with new Simulator interface

ekiwi:
This PR removes the old backend abstraction and replaces it with the new simulator interface from: https://github.com/ucb-bar/chisel-testers2/pull/337

Things I want to fix before this is merged:

- [x] add back simulator binary caching

The user facing API should stay mostly the same, with the biggest difference so far being that the `Verilator`/`Treadle` backend annotations and the `WriteVcdAnnotation` have been exposed through the `chiseltest._` toplevel package. The import from `chiseltest.internal._` will no longer work.


ducky64:
+1506, -2201 definitely has a bit of sticker shock...

Is this something that could be feasibly be broken down into more review-able chunks? Otherwise maybe higher-level docs or a synchronous meeting might be helpful?

src/main/scala/chiseltest/internal/BackendExecutive.scala
```diff
@@ -1,24 +1,56 @@
...
+import chiseltest.coverage.Coverage
+import chiseltest.simulator.Simulator
 import firrtl.AnnotationSeq
 import firrtl.annotations.ReferenceTarget
-import firrtl.transforms.CombinationalPath
+import firrtl.transforms.{CheckCombLoops, CombinationalPath}
 
-trait BackendExecutive {
-  def getTopModule(circuit: Circuit): BaseModule = {
+object BackendExecutive {
```
ekiwi, line 14: @ducky64 I changed this from a `trait` to an `object` since it is generic for all `Simulator` backends now. OK?

src/main/scala/chiseltest/internal/BackendExecutive.scala
```diff
@@ -1,24 +1,56 @@
...
+
+  private def getTopModule(circuit: Circuit): BaseModule = {
     (circuit.components.find(_.name == circuit.name)).get.id
   }
 
   /** Returns a Seq of (data reference, fully qualified element names) for the input.
     * name is the name of data
     */
-  def getDataNames(name: String, data: Data): Seq[(Data, String)] = Seq(data -> name) ++ (data match {
+  private def getDataNames(name: String, data: Data): Seq[(Data, String)] = Seq(data -> name) ++ (data match {
```
ekiwi, line 53: Is it ok to make this method private?

src/main/resources/simulator/sim_api.h
```diff
@@ -174,7 +173,7 @@ template <class T> class sim_api_t {
       size_t cmd;
       while(!recv_cmd(cmd));
       switch ((SIM_CMD) cmd) {
-        case RESET: reset(); is_reset = true; is_exit = true; break;
+        case RESET: /** no longer supported! */ break;
```
ekiwi, line 176: The reason why this is no longer needed is that the new interface treats reset like any other input pin.

.github/workflows/test.yml
```diff
@@ -25,30 +26,9 @@ jobs:
       - name: Setup Scala
         uses: olafurpg/setup-scala@v10
         with:
-          java-version: adopt@1.8
```
ekiwi, line 28: I merged the two checks into one by making the Java version part of the test matrix.

src/main/scala/chiseltest/package.scala
```diff
@@ -323,4 +323,10 @@ package object chiseltest {
   implicit def decoupledToDriver[T <: Data](x: ReadyValidIO[T]) = new DecoupledDriver(x)
 
   implicit def validToDriver[T <: Data](x: ValidIO[T]) = new ValidDriver(x)
+
+  // expose public flags
```
ekiwi, line 327: New annotations that we expose as part of chiseltest.

src/main/scala/chiseltest/simulator/Simulator.scala
```diff
@@ -0,0 +1,169 @@
...
+    */
+  def createContext(state: CircuitState): SimulatorContext
+}
+
+/** Defines a simulator backend that should be used. */
+trait SimulatorAnnotation extends NoTargetAnnotation {
+  def getSimulator: Simulator
+}
+
+trait WriteWaveformAnnotation extends NoTargetAnnotation {
```
ekiwi, line 69: This trait allows us to write a generic `getWavformFormat` function below

src/main/scala/chiseltest/internal/BackendExecutive.scala
ducky64: Huh? I'm confused.

Don't the methods in BackendExecutive depend on circuit state? Doesn't an object mean you can only have one of these (singleton)? What gain are we hoping for here?

src/main/scala/chiseltest/internal/BackendExecutive.scala
ducky64: If you can make it work, sure. It probably should have been `protected` in the first place, since this really isn't a public API. Not sure if `private` will work, but it depends on how you structure everything else?

src/main/scala/chiseltest/internal/BackendExecutive.scala
ekiwi: > Don't the methods in BackendExecutive depend on circuit state?

Only the `BackendInstance` depends on the circuit state. The `BackendExecutive` just contains a function that lets you create a `BackendInstance` from a Chisel module. Since it only contains functions, making it an `object` makes sense to me.

src/test/scala/chiseltest/backends/verilator/VerilatorCachingTests.scala
```diff
@@ -16,82 +17,58 @@ class VerilatorCachingTests extends AnyFlatSpec with ChiselScalatestTester with
...
-    // warmup to improve measurement accuracy
-    (0 until 2).foreach { _ => runTest(42, default) }
     startWithEmptyTestDir()
-    // here we test out hypothesis that without caching, the tet execution times will all be about the same
-    val times = (0 until 3).map { _ => runTest(42, default) }
-    val delta = avgRelativeDelta(times)
-    assert(delta < nonCachingMax)
-    println(times)
-    println(delta)
+    // by default no re-use should occure
```
ducky64, line 29: typo

src/test/scala/chiseltest/backends/verilator/VerilatorClockPokeTest.scala
```diff
@@ -51,7 +50,7 @@ class VerilatorClockPokeTest extends AnyFlatSpec with ChiselScalatestTester {
     }
   }
 
-  it should "clock-as-bool-verilator-clock-poke" in {
+  it should "clock-as-bool-verilator-clock-poke" ignore { // TODO: re-add multi clock support
```
ducky64, line 53: Is clock poking as a bool just not supported at all? Is this a limitation of our API or the underlying simulator?

ducky64 (commented):
It also looks like all the old legacy backends have been deleted from here, were they added back in #337? Were they rewritten or why are they being moved?

Otherwise looks fine skimming through it.

src/test/scala/chiseltest/backends/verilator/VerilatorClockPokeTest.scala
ekiwi: Clock poking as a bool only makes sense if there is a way to advance the simulation time. In this example that is done by advancing the "main" clock. However, I am not a fan of having such an "asymmetrical" API, where one clock needs to be stepped in order to advance time for the other clock. Since no-one seems to use this so far, I would like to punt the decision on how to support multi-clock to the future.

Wrt. simulator support: poking top-level clock inputs should be fine for verilator and treadle and probably also for VCS. Essent won't support it, but then it also just doesn't support multi-clock at all (Scott said that adding multi-clock support is an eventual goal).

src/test/scala/chiseltest/backends/verilator/VerilatorCachingTests.scala
ekiwi: fixed

chick (approved):
Looks good to me.
I have tested on my mac and linux box.

src/main/scala/chiseltest/simulator/TreadleSimulator.scala
```diff
@@ -16,7 +16,7 @@ case object TreadleBackendAnnotation extends SimulatorAnnotation {
 private object TreadleSimulator extends Simulator {
   override def name:        String = "treadle"
   override def isAvailable: Boolean = true
-  override def findVersions(): Unit = {
+  def findVersions(): Unit = {
```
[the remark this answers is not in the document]
ekiwi: I essentially demoted it to an internal API (by removing it from the public interface). So we can remove it any time we want.

src/main/scala/chiseltest/simulator/Simulator.scala
```diff
@@ -0,0 +1,142 @@
...
+    */
+  def getCoverage(): List[(String, Long)]
+
+  /** Resets all coverage counters to zero.
+    * @note Not supported by all simulators. Must result in a [[NotImplementedError]] if not supported.
+    */
+  def resetCoverage(): Unit
+}
+
+/** Throws by [[SimulatorContext.step]] if the circuit has no clock input */
```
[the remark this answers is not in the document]
ekiwi: fixed

src/main/scala/chiseltest/internal/BackendExecutive.scala
```diff
@@ -1,24 +1,56 @@
...
+    val tester = sim.createContext(lowFirrtl)
+
+    new GenericBackend(dut, portNames, pathsAsData, tester, coverageAnnotations)
+  }
+
+  private def componentToName(component: ReferenceTarget): String = {
+    component.name
+  }
+
+  private def getTopModule(circuit: Circuit): BaseModule = {
```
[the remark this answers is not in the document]
ekiwi: removed

src/main/scala/chiseltest/simulator/Caching.scala
```diff
@@ -0,0 +1,87 @@
...
+    Seq(System.getProperty("java.version"), scala.util.Properties.versionNumberString)
+
+  private def annotationHash(annos: AnnotationSeq): String = {
+    val relevant = annos.filterNot(ignoreAnno)
+    // we serialize every annotation individually so that we can later sort them
+    val values = relevant.map(a => JsonProtocol.serialize(Seq(a))).sorted
+    hashStrings(values)
+  }
+
+  private def ignoreAnno(a: Annotation): Boolean = {
```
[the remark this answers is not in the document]
ekiwi: I would like to leave this in as a reminder that we might have to eventually ignore certain annotations.
For now my goal was to get good enough caching to restore the old functionality that was available before. However, caching is far from solved yet.

PR #335, opened 2021-07-21
test: actually test caching

ekiwi:
During my work on https://github.com/ucb-bar/chisel-testers2/pull/334 I realized that the current tests keep on working, even if caching is removed.

The new tests try to determine whether something is cached by measuring the relative time it takes to execute a test/. I have observed that cached tests normally execute about 2.5 times faster than uncached tests, thus I am hopeful that this won't be too brittle.


chick (approved):
I think this looks fine. I have a couple of comments.
Relying on time seems likely to fail now and then for reasons that will not be obvious
I think it might be nice if there were a for of `test` that provided some controlled access to internals like working directory, cache hit, etc.

PR #336, opened 2021-07-21
remove simulator specific Verilog coverage flags

ekiwi:
If the user wants to rely on the simulator built-in Verilog coverage, they can directly supply the relevant flags to the simulator.

I talked to @vighneshiyer and he is OK removing this as it isn't a very clean API.

I improved the Verilator tests to actually check that an appropriate coverage file is generated.


chick (approved):
Looks good.
Passes all tests on a8 using vcs

PR #338, opened 2021-07-22
ci: test on mac

ekiwi:


ekiwi:
This should be merged only **after** https://github.com/ucb-bar/chisel-testers2/pull/339

chick (approved):
Looks ok. 
Tests are not that long but the matrix will now run the tests 8 times

PR #341, opened 2021-07-23
vcs does not support coverage

ekiwi:


src/main/scala/chiseltest/legacy/backends/verilator/VerilatorExecutive.scala
```diff
@@ -112,7 +112,15 @@ object VerilatorExecutive extends BackendExecutive {
...
         return Some(
-          new VerilatorBackend(dut, portNames, pathsAsData, command, targetDirPath.toString(), coverageAnnotations)
+          new VerilatorBackend(
+            dut,
+            portNames,
+            pathsAsData,
+            command,
+            targetDirPath.toString(),
+            coverageAnnotations,
+            true
```
chick, line 122:
```suggestion
            supportsCoverage = true
```

src/main/scala/chiseltest/legacy/backends/vcs/VcsBackend.scala
```diff
@@ -21,4 +21,4 @@ class VcsBackend[T <: Module](
   command:             Seq[String],
   targetDir:           String,
   coverageAnnotations: AnnotationSeq)
-    extends VerilatorBackend(dut, dataNames, combinationalPaths, command, targetDir, coverageAnnotations)
+    extends VerilatorBackend(dut, dataNames, combinationalPaths, command, targetDir, coverageAnnotations, false)
```
chick, line 24:
```suggestion
    extends VerilatorBackend(dut, dataNames, combinationalPaths, command, targetDir, coverageAnnotations, supportsCoverage = false)
```

chick (approved):
Looks fine to me.
Note suggested changes, I don't feel strongly about it, but believe it to be a good practice to not pass naked true or false as parameter

PR #342, opened 2021-07-23
make caching tests more stable

ekiwi:
1) small warmup compilation that isn't timed
2) average differences instead of checking all of them


chick (approved):
LGTM

PR #349, opened 2021-07-30
Verilator JNI

ekiwi:
TODOs:

- [ ] think about replacing the compiler library
- [ ] try to unload native libs when done: https://stackoverflow.com/questions/11939468/how-to-make-jni-reload-shared-library-on-calling-native-function


ekiwi:
I am giving up on JNI for now since it is really hard to load more than one simulation library in parallel. The trick with generating Java adapter classes dynamically works from IntelliJ, but it fails to work with `sbt`'s custom class loader architecture and I would rather not tell users to use inferior `sbt` test settings. For now we should settle for the JNA based approach which gives around a 8x speedup on the peek/poke heavy GCD benchmark (compared to ~40x with JNI).

PR #350, opened 2021-07-30
simulator: explicitly convert argument to bytes

ekiwi:
Otherwise there might be errors trying to resolve the overloaded put method.

This is addressing the issue observed by @sequencer


sequencer:
Pending CI in https://github.com/sequencer/chipsalliance-playground/actions/runs/1084162727
I'm on the flight, I think this can be approved after my ci pass.

sequencer (approved):

PR #353, opened 2021-08-05
experiment to get faster coverage feedback from verilator

ekiwi:


ekiwi:
closed for now, might be revived later

PR #354, opened 2021-08-06
simulator: remove old IPC option for verilator

ekiwi:
It is too much overhead to maintain both and
with this PR, we support caching the shared
library.


chick (approved):
This looks good to me.
Tests run fine

PR #356, opened 2021-08-11
Minimal Formal Verification Support

ekiwi:
This PR adds some fundamental formal verification support.

A user needs to:
1) add `with Formal` to their `ChiselScalatestTester` derived testing class
2) call `verify` with a verification operation

So for a simple bounded check the syntax is:
```.scala
verify(new Dut, Seq(BoundedCheck(kMax = 1)))
```

Public Annotations included in this PR:
- `Z3EngineAnnotation`: use Z3 for model checking (default)
- `CVC4EngineAnnotation`: use CVC4 for model checking
- `BoundedCheck`: the only verification op so far, but could be extended with something like `KInduction`
- `ResetOption`: by default the formal backend will drive the reset pin high for at least one cycle in the beginning of the check; `ResetOption(0)` will remove any constraints from the reset pin, `ResetOption(3)` will enforce a reset period of at least three cycles

When the check fails, the counter example is automatically replayed with treadle, thus the generated VCD should have the same quality as if it was generated from a unit test.

Things that are missing from this MVP:
- support for more conservative modelling of undefined behaviors
- support for stop statements as assertions
- proofs (using k-induction or other backend specific techniques)
- library support for `past`
- coverage support

### DIFF
--- a/src/main/scala/chiseltest/formal/AddResetAssumptionPass.scala
+++ b/src/main/scala/chiseltest/formal/AddResetAssumptionPass.scala
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest.formal
+
+import chisel3.util.log2Ceil
+import firrtl._
+import firrtl.annotations._
+import firrtl.options.Dependency
+import firrtl.transforms._
+
+/** adds an assumption to the toplevel module that all resets are active in the first cycle */
+private object AddResetAssumptionPass extends Transform with DependencyAPIMigration {
+  // run on lowered firrtl
+  override def prerequisites = Seq(
+    Dependency(firrtl.passes.ExpandWhens),
+    Dependency(firrtl.passes.LowerTypes),
+    Dependency(firrtl.transforms.RemoveReset),
+    // try to work around dead code elimination removing our registers
+    Dependency[firrtl.transforms.DeadCodeElimination]
+  )
+  override def invalidates(a: Transform) = false
+  // since we generate PresetRegAnnotations, we need to run after preset propagation
+  override def optionalPrerequisites = Seq(Dependency[PropagatePresetAnnotations])
+  // we want to run before the actual Verilog is emitted
+  override def optionalPrerequisiteOf = firrtl.stage.Forms.BackendEmitters
+
+  override def execute(state: CircuitState): CircuitState = {
+    val resetLength = getResetLength(state.annotations)
+    if (resetLength == 0 || resetIsPreset(state.circuit.main, state.annotations)) return state
+
+    val main = state.circuit.modules.collectFirst { case m: ir.Module if m.name == state.circuit.main => m }.get
+    val (clock, reset) = findClockAndReset(main)
+    val namespace = Namespace(main)
+
+    // add a port for the preset init
+    val preset = ir.Port(ir.NoInfo, namespace.newName("_preset"), ir.Input, ir.AsyncResetType)
+    val presetAnno = PresetAnnotation(CircuitTarget(main.name).module(main.name).ref(preset.name))
+
+    // add reset count register
+    val resetCountType = ir.UIntType(ir.IntWidth(List(1, log2Ceil(resetLength + 1)).max))
+    val resetCount = ir.DefRegister(
+      ir.NoInfo,
+      namespace.newName("_resetCount"),
+      resetCountType,
+      clock,
+      reset = ir.Reference(preset).copy(flow = SourceFlow),
+      init = Utils.getGroundZero(resetCountType)
+    )
+    val resetCountSource = ir.Reference(resetCount).copy(flow = SourceFlow)
+
+    // we are in the reset phase iff the counter is less than the reset length
+    val lessThanCycles = ir.DoPrim(
+      PrimOps.Lt,
+      List(resetCountSource, ir.UIntLiteral(resetLength, resetCountType.width)),
+      List(),
+      Utils.BoolType
+    )
+    val resetPhase = ir.DefNode(ir.NoInfo, namespace.newName("_resetPhase"), lessThanCycles)
+    val resetPhaseRef = ir.Reference(resetPhase).copy(flow = SourceFlow)
+
+    // we increment the counter as long as we are in the reset phase, after that it just retains its value
+    val nextValue = Utils.mux(resetPhaseRef, plusOne(resetCountSource), resetCountSource)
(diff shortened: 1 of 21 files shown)


sequencer:
I'll review this this weekend.

sequencer:
I have no nitpick on this, This is what I want! I never image SMT can write VCD. Maltese is really amazing!

src/main/scala/chiseltest/formal/AddResetAssumptionPass.scala
```diff
@@ -0,0 +1,123 @@
...
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named clock!")
+      )
+    val reset = m.ports
+      .find(_.name == "reset")
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named reset!")
+      )
+    (ir.Reference(clock).copy(flow = SourceFlow), ir.Reference(reset).copy(flow = SourceFlow))
+  }
```
sequencer, line 115: This limit the dut to be `Module` with clock and reset, In the future, maybe we can add another pass to find all real clock and reset for a `RawModule`.

src/main/scala/chiseltest/formal/Formal.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/** Adds the `verify` command for formal checks to a ChiselScalatestTester */
+trait Formal { this: ChiselScalatestTester =>
+  def verify[T <: Module](dutGen: => T, annos: AnnotationSeq): Unit = {
+    val withTargetDir = addDefaultTargetDir(getTestName, annos)
+    Formal.verify(dutGen, withTargetDir)
+  }
+  private def getTestName: String = sanitizeFileName(scalaTestContext.value.get.name)
+}
```
sequencer, line 37: I personally think `Formal.verify` should not be only expose to Scalatest, maybe Formal can be a standalone trait?
IIRC `ChiselScalatestTester` only provides the default directory creating utils.

src/main/scala/chiseltest/formal/Formal.scala
```diff
@@ -0,0 +1,76 @@
...
+
+private object Formal {
```
sequencer, line 39: Or we just public this object, while making most of non-user functions private.

src/main/scala/chiseltest/formal/backends/Trace.scala
```diff
@@ -0,0 +1,36 @@
...
+  inputs:  Seq[Seq[(String, BigInt)]],
+  regInit: Seq[(String, BigInt)],
+  memInit: Seq[(String, Seq[BigInt])])
+
+private object Trace {
+  def replayOnTreadle(trace: Trace, circuit: ir.Circuit, annos: AnnotationSeq): Unit = {
+    val dut = TreadleBackendAnnotation.getSimulator.createContext(CircuitState(circuit, annos))
+
+    // initialize
+    trace.regInit.foreach { case (name, value) => dut.poke(name, value) }
```
sequencer, line 19: So you have implemented the internal signal poke?

sequencer (approved):
Merge as you wish, I really need this API. Only nitpick is I think we don't wanna formal depends on scalatest.

sequencer:
I just come up with this idea:
Maybe an `Aspect` based API is good for inserting `assume` and `assert` to add different test cases?

src/main/scala/chiseltest/formal/backends/Trace.scala
ekiwi: This only works for treadle. Poking registers in Verilator is very difficult as it turns out and will probably never be implemented.

ekiwi:
> Maybe an `Aspect` based API is good for inserting `assume` and `assert` to add different test cases?

I was playing around with that here: https://github.com/ekiwi/kiwi-formal/blob/main/src/kiwi/formal/VerificationAspect.scala

However, I don't have an API that I like yet.

sequencer:
@ekiwi I added a734e2e for mill build-from-source support. But I don't know how did I manage to push to your branch.

src/main/scala/chiseltest/formal/backends/Trace.scala
sequencer: I think here exist some problem:
I only see Data -> Name map expose via `DataMirror` rather than a FIRRTL pass, which make `reg` you poke not found in the map.

sequencer: I think this can happen via:
1. add a pass to extract `Data -> Verilog Name` via RenameMap, by default it can be port only, and let user add expose annotations.
2. generate a Verilator [config file](https://verilator.org/guide/latest/exe_verilator.html#cmdoption-arg-public) to expose those signals.
3. improve JNA/JNI interface to this.

ekiwi: You could always try to prototype it, but from my experiments and looking at the Verilator generated code, the problem is that on a clock edge, Verilator not only updates the register values, it also calculated the next register values. Thus if you change a register after the clock edge, it won't affect the next value that was already calculated and thus your change is going to be overwritten at the next clock edge (unless the register isn't updated by the hardware).

ekiwi: That is what I discovered when trying to use registers for coverage counter: https://github.com/ucb-bar/chisel-testers2/pull/353

sequencer: maybe [cocotb](https://github.com/cocotb/cocotb/blob/fba7b854b00375500552ff8804757f468152b111/cocotb/handle.py#L694-L698)  seems to be better to refer to?
It seems to use `Force` and `Release` to achieve this.

src/main/scala/chiseltest/formal/Formal.scala
ekiwi: The problem is that how you get a test name is completely different for `utest`.

ekiwi: I changed it so that `Formal` only depends on having a `getTestName` function. Maybe you can figure out how to get that working with `utest`.

tdb-alcorn:
:tada: just wanted to say thanks for implementing this, it is awesome! can't believe we finally got here :)

marvintau:
Any plan for implementing k-induction? 😉

ekiwi:
> Any plan for implementing k-induction? wink

I would be happy to review PRs for k-induction support. Unfortunately I do not have time to work on this myself.

PR #359, opened 2021-08-12
test: move utils into their own package

ekiwi:
pulled out from https://github.com/ucb-bar/chisel-testers2/pull/356


chick (approved):
Looks great. I like the way target dir is handled

PR #360, opened 2021-08-12
simulator: extends the number of annotations that should be removed

ekiwi:
This ensures more isolation between different phases in the compilation process.
E.g. when we set the log level for compiling to LoFirrtl it should not leak into the compilation
to Verilog.


chick (approved):
lgtm

PR #363, opened 2021-08-13
Fix Verilator Coverage Regression

ekiwi:
This PR fixes the regression reported by @oharboe in https://github.com/ucb-bar/chisel-testers2/pull/315


chick (approved):
I'm not that familiar with this memory stuff, but code looks good, tests seem good.

PR #364, opened 2021-08-14
scalatest: public annotations, no flags

ekiwi:
This PR modifies the `ChiselScalatestTester`:

- it adds the option to pass annotations through the regular `test` method (similar to how the Raw and the utest tester already do it)
- it makes the `withAnnotation` methods parts of the `TestBuilder` object and thus available without `experimental` import
- it removes the deprecated legacy `TesterOptions`
- removes any support for passing flags directly since this was never well tested or supported


ducky64 (commented):
I think overall this looks fine, and the cleanup of removing the not-really-used and not-really-supported flags is appreciated.

One question: now it looks like we can specify annotations to both `test(...)` and as `withAnnotations(...)`. Can we standardize on one, or is there a reason to keep both around? If Chisel is doing something similar, we should be consistent, otherwise I'd prefer `withAnnotations(...)` because it's composable and explicit.

ekiwi:
> One question: now it looks like we can specify annotations to both `test(...)` and as `withAnnotations(...)`. Can we standardize on one, or is there a reason to keep both around? If Chisel is doing something similar, we should be consistent, otherwise I'd prefer `withAnnotations(...)` because it's composable and explicit.

I personally would prefer just supplying annotations to `test(...)`, however, we should keep the `withAnnotations` in order not to break old code too much. As it stands, people will have to change imports, but the main code should be source compatible.
We can always deprecate `withAnnotations` if we want to get rid of it eventually.
What do you think?

ducky64:
I think we need to only support one and deprecate the other. I'd lean towards supporting `withAnnotations` unless there's an argument for having it be part of the `test` function, because `withAnnotations` is composable.

ekiwi:
> I think we need to only support one and deprecate the other. I'd lean towards supporting `withAnnotations` unless there's an argument for having it be part of the `test` function, because `withAnnotations` is composable.

The arguments against `withAnnotations` are:
1. It is a lot more noisy then just adding a `Seq(VerilatorBackendAnnotation)` to the call to `test`
2. I have never seen two `withAnnotations` calls used
3. The RawTester and the utest based tester already accept annotations as an optional second argument to test

ducky64:
For (2), it's more of future-proofing to support other options later. I don't expect multiple chained `withAnnotation` calls.
For (1), I'd argue that withAnnotation is more clear than just a bare `Seq` dumped in `test`?

ekiwi:
OK. I still like the idea of passing annotations with `test(....)` directly, however, we should not hold up this PR over it. I am going to remove the option for now.

ducky64 (approved):
lgtm. It looks like annos has still been added as an argument to test() (which is now consistent with the other backends?) but we should at some point resolve this by standardizing on one and deprecating the other.

ekiwi:
> lgtm. It looks like annos has still been added as an argument to test()

This should be removed in the latest push (which came in around the same time as your comment..)

ducky64:
Yep, looks good now!

ekiwi:
@schoeberl With this PR merged, `withAnnotation` and all simulator + waveform dump annotations are now stable and available through `import chiseltest._`. I hope that helps!

mwachs5:
by the way the scaladoc still references `.withFlags`

jackkoenig:
Also once nice thing about CLI arguments is that you can have `--help`, the same doesn't exist for the annotations.

ekiwi:
> Also once nice thing about CLI arguments is that you can have `--help`, the same doesn't exist for the annotations.

We just had no one using or maintaining the argument parser, so that most of the annotations weren't actually properly accessible. So we decided to remove this option for now. The documentation bug that Megan noticed needs to be fixed.

PR #365, opened 2021-08-14
Single threaded by default

ekiwi:
I am trying to see what the impact of using a single threaded backend by default would be. This could be an alternative to adapting a `PeekPokeTester` compatibility API since performance for single threaded tests seems to be the major concern that keeps users from switching.

So far, on the GCD benchmark the numbers (with verilator) are:
- plain SimulatorContext: 8.6s
- chiseltest w/ the ThreadedBackend: 282s
- chiseltest w/ the SingleThreadBackend: 12s

### DIFF
--- a/src/main/scala/chiseltest/internal/BackendExecutive.scala
+++ b/src/main/scala/chiseltest/internal/BackendExecutive.scala
@@ -13,7 +13,7 @@ import firrtl.transforms.{CheckCombLoops, CombinationalPath}
 
 object BackendExecutive {
 
-  def start[T <: Module](dutGen: () => T, testersAnnotationSeq: AnnotationSeq): GenericBackend[T] = {
+  def start[T <: Module](dutGen: () => T, testersAnnotationSeq: AnnotationSeq): BackendInstance[T] = {
 
     // elaborate the design
     val (highFirrtl, dut: T) = Compiler.elaborate[T](dutGen, testersAnnotationSeq)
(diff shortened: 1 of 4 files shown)


sequencer:
That’s great!
I used to have these thoughts:
1. IPC from Java to C consumes too much, step based API seems a little embarrassing.
my idea is creating a simulator then let it free to run, using annotation to create a trigger, when this trigger is true, simulator interrupt Scala via DPI or VPI, then peek or poke in Scala, this reduced the IPC overhead. 

2. Thread at Scala was the worst design I have ever seen. 
as for threading, I think we can consider send fork command down to simulator, using fork syscall to really fork in the system. That’s a tough job, but I think that’s the only possible solution to gain some performance. 

tldr,
Scala should have a as lean as possible design, features need to be implemented inside simulator, exposing in Scala is really a bad design for performance.

src/main/scala/chiseltest/internal/Testers2.scala
```diff
@@ -41,6 +41,16 @@ case object CachingAnnotation extends TestOptionObject {
   )
 }
 
+case object UseThreads extends TestOptionObject {
```
ekiwi, line 44: Here is the `UseThreads` annotation that I was talking about at the dev meeting today.

src/main/scala/chiseltest/internal/BackendExecutive.scala
```diff
@@ -36,7 +36,12 @@ object BackendExecutive {
     val sim = Simulator.getSimulator(testersAnnotationSeq)
     val tester = sim.createContext(lowFirrtl)
 
-    new GenericBackend(dut, portNames, pathsAsData, tester, coverageAnnotations)
+    val useThreads = testersAnnotationSeq.contains(UseThreads)
```
ekiwi, line 39: This is an example of how we can use annotations in `chiseltest`

PR #371, opened 2021-08-15
Single thread

sequencer:
I implemented this to subsume #365,
The reason I reimplement this is because I'm going to implement a new API to let simulator interrupt Scala to speed up.

This PR introduce a breaking API that if user wanna use threading API, they should explicitly add `UseThreadingAnnotation`.

No doubt, a slim backend is really necessary for high performance simulator to reduce the overhead introduced by ThreadedBackend. So this PR introduced `SingleThreadBackend` to speed up.
slipping this to user will gain a lot performance improvement without any code modifications, however this will break codes if user use `fork/join` etc APIs, it will throw a `ThreadOrderDependentException`. I personally prefer this, and add some release note to introduce a breaking change.

Another option is to remove `UseThreadingAnnotation` and use `DontUseThreadingAnnotation`, this will explicitly ask user to add this Annotation to their code to gain that performance.


src/main/scala/chiseltest/internal/BackendExecutive.scala
```diff
@@ -24,19 +24,23 @@ object BackendExecutive {
...
-    new GenericBackend(dut, portNames, pathsAsData, tester, coverageAnnotations)
+    val useThreading = testersAnnotationSeq.contains(UseThreadingAnnotation)
+
+    if (useThreading) {
+      // extract combinatorial loops from the LoFirrtl circuit
+      val pathAnnotations = (new CheckCombLoops).execute(lowFirrtl).annotations
+      val paths = pathAnnotations.collect { case c: CombinationalPath => c }
+      val pathsAsData = combinationalPathsToData(dut, paths, portNames, componentToName)
+      new ThreadedBackend(dut, portNames, pathsAsData, tester, coverageAnnotations)
+    } else
```
sequencer, line 42: Only `ThreadedBackend` new combinational Loop detection.

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -59,43 +42,80 @@ trait BackendInterface {
...
+    throw new ClockTimeoutNotImplementedException(s"setTimeout not implemented in ${getClass.getName}")
+  }
 
-  /** Advances the target clock by one cycle.
-    */
-  def step(signal: Clock, cycles: Int): Unit
+  // Threading API
+  def doFork(runnable: () => Unit, name: Option[String], region: Option[Region]): AbstractTesterThread = {
+    throw new ThreadNotImplementedException(s"fork/join not implemented in ${getClass.getName}")
+  }
```
sequencer, line 81: By default `ThreadNotImplementedException` will be thrown if not using threading API, while user use `fork/join`

src/main/scala/chiseltest/internal/GenericBackend.scala
```diff
@@ -1,217 +1,47 @@
...
-/** Chiseltest threaded backend using the generic SimulatorContext abstraction from [[chiseltest.simulator]] */
-class GenericBackend[T <: Module](
-  val dut:                T,
-  val dataNames:          Map[Data, String],
-  val combinationalPaths: Map[Data, Set[Data]],
-  tester:                 SimulatorContext,
-  coverageAnnotations:    AnnotationSeq)
-    extends BackendInstance[T]
-    with ThreadedBackend[T] {
+abstract class GenericBackend[T <: Module] extends BackendInterface with LazyLogging {
```
sequencer, line 6: A minimal GenericBackend only contains 
1. basic API(peek poke step(I still not so sure step is basic)) implementation
2. dataNames to map Data -> String(I personally this should be a full hierarchy path in the future.)

src/main/scala/chiseltest/internal/SingleThreadBackend.scala
```diff
@@ -0,0 +1,45 @@
+package chiseltest.internal
+
+import chisel3.{Data, Module}
+import chiseltest.coverage.TestCoverage
+import chiseltest.simulator.SimulatorContext
+import chiseltest.testableClock
+import firrtl.AnnotationSeq
+
+class SingleThreadBackend[T <: Module](
```
sequencer, line 9: The slim API

src/main/scala/chiseltest/internal/SingleThreadBackend.scala
```diff
@@ -0,0 +1,45 @@
...
+    with BackendInstance[T] {
+  override def expectBits(
+    signal:  Data,
+    value:   BigInt,
+    message: Option[String],
+    decode:  Option[BigInt => String],
+    stale:   Boolean
+  ): Unit = {
+    super.expectBits(signal, value, message, decode, stale)
+    Context().env.checkpoint()
```
sequencer, line 24: checkpoint at once, this is a little different then `ThreadedBackend`, but I think this is an improvement.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -36,8 +61,162 @@ case class ForkBuilder(name: Option[String], region: Option[Region], threads: Se
...
+    tester.poke(dataNames(signal), intValue)
+    logger.debug(s"${resolveName(signal)} <- $intValue")
+  }
+
+  override def peekClock(signal: Clock): Boolean = {
+    doPeek(signal, new Throwable)
+    val a = tester.peek(dataNames(signal))
+    logger.debug(s"${resolveName(signal)} -> $a")
+    a > 0
+  }
```
sequencer, line 98: override default API to insert threading recording.

sequencer:
Ask @ekiwi for review, I think this is done.

ekiwi:
> Ask @ekiwi for review, I think this is done.

Richard is back and knows a lot more about this part of the code. I added him as a reviewer.

In general I don't think we should make the SingleThreaded backend the default. The user needs to explicitly ask for it. By default, `fork` should be supported without any special user annotation. This is required for backwards compatibility and we also don't want to make it harder to use some of the more exciting features of chiseltest.

src/main/scala/chiseltest/internal/SingleThreadBackend.scala
ekiwi: Could you write a test that makes sure that the errors reported are the same no matter whether threading is enabled or not?

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -2,13 +2,37 @@
...
+import chiseltest.coverage.TestCoverage
+import chiseltest.simulator.SimulatorContext
+import firrtl.AnnotationSeq
+import firrtl.annotations.NoTargetAnnotation
 
 import scala.collection.mutable
 
+trait AbstractTesterThread
+
+class TesterThreadList(protected val elts: Seq[AbstractTesterThread]) {
```
ekiwi, line 18: This should be private

src/main/scala/chiseltest/internal/ThreadedBackend.scala
```diff
@@ -2,13 +2,37 @@
...
+import java.util.concurrent.{ConcurrentLinkedQueue, Semaphore}
+import chiseltest.{Region, TemporalParadox, ThreadOrderDependentException, TimeoutException}
+import chiseltest.coverage.TestCoverage
+import chiseltest.simulator.SimulatorContext
+import firrtl.AnnotationSeq
+import firrtl.annotations.NoTargetAnnotation
 
 import scala.collection.mutable
 
+trait AbstractTesterThread
```
ekiwi, line 16: should be private

src/main/scala/chiseltest/internal/SingleThreadBackend.scala
sequencer: This won’t affect threading code,, this is just for single thead api. and it’s reasonable to checkpoint at schedule for threading api

ekiwi: Basically what I would like to ensure is that if you have the exact same test run with the two different backends, the errors reported should be 100% the same. Switching the backends should not affect any behavior besides the fact that `fork` won't work in a non-threaded context.

src/main/scala/chiseltest/internal/ThreadedBackend.scala
sequencer: both of this two was public, I think cleanup can be a separate PR?

src/main/scala/chiseltest/internal/ThreadedBackend.scala
sequencer: both of this two was public, I think cleanup can be a separate PR?

src/main/scala/chiseltest/internal/SingleThreadBackend.scala
sequencer: you mean "two different backends" is for both threading and non-threading?
I prefer achieving this in the future, after refactoring entire chiseltest into the Stage/Phases as the last time I did.

sequencer: > the errors reported should be 100% the same

I think this is for different "simulator backends" like Verilator, VCS, NCSIM etc. But for different API, I think this restriction can be relaxed, in this case, the only difference is whether report asserting failed in this cycle or next cycle.
`ThreadedBackend` need to report in the next cycle to gather all exception among the thread.
However `SingleThreadAPI` only contains one thread, and can throw exception at his cycle.

ekiwi: > `ThreadedBackend` need to report in the next cycle to gather all exception among the thread.
> However `SingleThreadAPI` only contains one thread, and can throw exception at his cycle.

In order for the threading option to only be an optimization hint, both SingleThreaded and ThreadedBackend need to behave 100% the same on tests that do not use fork.

sequencer: > In order for the threading option to only be an optimization hint, both SingleThreaded and ThreadedBackend need to behave 100% the same on tests that do not use fork.

I see, so you mean, for entire tests, we should ran both SingleThreaded and Threaded successfully?

ekiwi: It is enough if you craft 2-3 example tests that make sure both threaded and single threaded behave the same.

sequencer: Got it :)

ducky64:
So a few thoughts here:
- I agree that simulator performance is important, and that threading / the scheduler is probably the top contributor to the performance bug right now.
- But, a major part of the bet with ChiselTest is the emphasis on expressiveness and re-usable components (like DecoupledDriver). I'm not sure there's a good way to write those without threading (or coroutines which aren't supported in Java; or writing explicit state machines like iotesters which is 🗑🔥). Because of this, I think that having a highly restricted, even if highly performant, API by default is not the right trade-off. There might be a case to be made for a secondary opt-in API that disallows threading, but I'm also iffy on that.
- I think there's low-hanging fruit to improve scheduler performance by preventing unnecessary context switches (loosely defined as where the thread being switched to doesn't actually do any useful work in its timeslot). If there's only one thread, it shouldn't need a context switch, and if the target thread is blocking on some IO state (eg, wait for low, wait for high), that also shouldn't need a context switch. My guess is that is probably the major cause of performance bugs right now. The one caveat with this approach is that there's nothing preventing the user from using threading features, so it's eas(ier) to write subtle performance bugs, though hopefully the impact will be more limited.

ekiwi:
Just to summarize what I think the current situation with this PR is:

- I believe that @ducky64 and I agree that _by default_ all threading features should be available (this PR currently disables threading by default so it won't be accepted)
- I think it would be valuable to provide a `NoThreadingAnnotation` to allow users that do not want to use threading to opt-in for a faster single threaded backend
- I think @ducky64 would prefer that we optimize the existing threaded backend for when the user isn't actually forking which would provide everyone with immediate benefits without having to know about any special annotations, but also seems quite challenging to implement

sequencer:
Sorry for the delay, I'm working on this, too busy since last Friday. :(

ekiwi:
> Sorry for the delay, I'm working on this, too busy since last Friday. :(

No worries. Just wanted to summarize the discussion.

sequencer:
> I think it would be valuable to provide a NoThreadingAnnotation to allow users that do not want to use threading to opt-in for a faster single threaded backend

I'd like to provide this API instead: `BackendInterfaceAnnotation(someInterface)` to make it possible to provide other interface implementations in the future. What do you think?

ekiwi:
> What do you think?

No. The whole `BackendInterface` is an internal API right now and we do not want to expose that to the users. There is too much in flux to expose it right now without a good use-case. None of the stuff in `chiseltest.internal` should leak into our user facing API.

sequencer:
So, I think I need to modify this API: 
https://github.com/ucb-bar/chisel-testers2/blob/df21b65814d32d6848f02d2f43a7abb5e0ed216a/src/main/scala/chiseltest/ChiselScalatestTester.scala#L121-L123

 to `def test[T <: Module](dutGen: => T, annos: Seq[Annotation] = Seq(UseThreadingAnnotation)): TestBuilder[T]` to provide the ability to let use override this default, since we currently have the API: `withAnnotations` but doesn't have the `withoutAnnotations` or `filterAnnotation.`

Do you think this is doable?

ekiwi:
> Do you think this is doable?

Why not add a `NoThreadingAnnotation`? The default should not need an annotation.
Essentially - they way I think of this - the user provides a _performance hint_ to the runtime that they do not require threading and the runtime can make use of that.

sequencer:
> Why not add a NoThreadingAnnotation? The default should not need an annotation.
Essentially - they way I think of this - the user provides a performance hint to the runtime that they do not require threading and the runtime can make use of that.

OK, I'll add `NoThreadingAnnotation` to select `SingleThreadedBackend`.

ekiwi:
> OK, I'll add `NoThreadingAnnotation` to select `SingleThreadedBackend`.

Please note, that while I would be happy with that small API extension, I am not sure if @ducky64 agrees.

sequencer:
I switch to `UseSingleThreadingAnnotation` for more concrete API naming.

ducky64:
I think from a user-facing perspective, an _experimental_ `NoThreadingAnnotation` could be useful - in the short term to select the single-threaded backend, and in the long term (when we optimize the scheduler to have the single-threaded case be fast) this can be used to enforce that fast path is taken. I emphasize experimental because I think there's still significant uncertainty about this API, but there's also an evolutionary path forward.

I haven't looked at the code changes here recently, but I'd prefer that this short term solution minimize modifications to the existing codebase (with threading), because I think scheduler improvements are the right solution and may obsolete this solution. Like, feel free to argue your case if you think deeper changes are necessary, but there's a bar to clear.

sequencer:
Sure. So you are suggesting changing `UseSingleThreadingAnnotation` to `NoThreadingAnnotation`?

> the single-threaded backend, and in the long term (when we optimize the scheduler to have the single-threaded case be fast) this can be used to enforce that fast path is taken.

Generally, I think the performance bottleneck is chiseltest almost pull everything from simulation to Scala. This of cause provide a smooth experience to users. However, this significantly increases the IPC overhead.

Personally, I don't think beside adding `fork` features to Verilator/treadle and other simulator backends, there exists any possible solutions from Scala(JVM): you still need to check threading conflict in each cycle with `peek`. Curious, how to make this "fastpath" happen?

In my mind, I'm trying to create a new interface: using AOP to inject `trigger` API, when `trigger` is true, then Verilator stops and send an interrupt to JVM. This asynchronous API make simulator free to run until encounter a trigger exception. This can reduce most of IPC problem.

ekiwi:
> In my mind, I'm trying to create a new interface: using AOP to inject `trigger` signal, when `trigger` is true, then Verilator stops and send an interrupt to JVM. This asynchronous API makes simulator free to run until encounter a trigger exception. This can reduce most of IPC problem.

I would be curious to see your speed-up number on the GCD benchmark: https://github.com/ekiwi/chisel-testers2/blob/0eb16106794d2c9b68fd91cb46b06ae35d4d7bbc/src/test/scala/chiseltest/benchmark/ChiseltestBenchmark.scala

schoeberl:
Wouldn't it just be a flag to skip synchronization code per default and switch it to use on the first call of fork?

sequencer:
> Wouldn't it just be a flag to skip synchronization code per default and switch it to use on the first call of fork?

I think this can be an optimization to `ThreadedBackend`, but a simple `SingleThreadBackend` is also necessary for users who doesn't need any `Threaded` API.

PR #372, opened 2021-08-16
Use ScalaTest 3.1 for derpication warnings

schoeberl:
not yet to merge, needs to be tested

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -70,7 +70,7 @@ scalacOptions ++= Seq(
 libraryDependencies ++= Seq(
   "edu.berkeley.cs" %% "chisel3"       % defaultVersions("chisel3"),
   "edu.berkeley.cs" %% "treadle"       % defaultVersions("treadle"),
-  "org.scalatest"   %% "scalatest"     % "3.2.9",
+  "org.scalatest"   %% "scalatest"     % "3.1.2",
   "com.lihaoyi"     %% "utest"         % "0.7.9",
   "net.java.dev.jna" % "jna" % "5.8.0",
   "org.scala-lang"   % "scala-reflect" % scalaVersion.value,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -70,7 +70,7 @@ scalacOptions ++= Seq(
 libraryDependencies ++= Seq(
   "edu.berkeley.cs" %% "chisel3"       % defaultVersions("chisel3"),
   "edu.berkeley.cs" %% "treadle"       % defaultVersions("treadle"),
-  "org.scalatest"   %% "scalatest"     % "3.2.9",
+  "org.scalatest"   %% "scalatest"     % "3.1.2",
```
ekiwi, line 73: `3.1.4` seems to be the latest `3.1.x` release. Could you try changing it to that?

ekiwi (changes requested):
Thanks for paying attention and looking into this @schoeberl !
I agree that making users switch from 3.0 to 3.2 directly without the deprecation warnings would have been a bad idea.

build.sbt
schoeberl: 3.1.4 works as well, did do the change

ekiwi (approved):
Thanks! :ship:

sequencer:
Basically, It is troublesome to force ChiselTest depending on ScalaTest, this will force user codes depends on the same version as ChiselTest does, user cannot change or bump their ScalaTest version.
I think in the future, better way might be spliting ScalaTest and UTest dependency out from ChiselTest codebase to decouple this dependency, Currently ScalaTest API only creates test folder and wrapper the exception. maybe it might be better to split them out from ChiselTest codebase.

ekiwi:
I agree with splitting out utest. However, for non-expert users, I think it is better to provide a default solution that works with a wide range of IDEs + sbt.

However, one thing we could consider is to actually inherit a specific spec.
`scalatest` actually recommends that you do not extend `AnyFlatSpec` for every test you write. Instead you are supposed to make one project specific spec and then use that everywhere. If we had done that, we wouldn't have to change things in quite so many places. Intel did the right thing for VTA: https://github.com/apache/tvm-vta/blob/main/hardware/chisel/src/test/scala/unittest/Generic.scala#L30

So one way to help newcomers while sheltering them from `scalatest` changes would be to define our own `ChiselSpec` that directly inherits from `AnyFlatSpec` and then change the docs to tell people to always use the `ChiselSpec`.

Issue #366, opened 2021-08-14
New unit-test added: Unsatisfied link error with Verilator when using asserts in tests


oharboe:
This was with 0.5-SNAPSHOT as of writing.

I got this error when I tried 0.5-SNAPSHOT today.

Ubuntu 20.04

```
[info]   java.lang.UnsatisfiedLinkError: Unable to load library '.../verilated/VDoohickeyTestBench_123':
[info] .../verilated/VDoohickeyTestBench_123': undefined symbol: _Z13sc_time_stampv
```

oharboe:
Reproducible with pull request #367

ekiwi:
Thanks for all the reports @oharboe !

ekiwi:
Patch just got in and will be published in ~10-15min. Keep these reports coming and I will try to fix them. Thanks you so much!

PR #373, opened 2021-08-16
Fix verilator linking error

ekiwi:
This regression was introduced when switching to the JNA based harness. Thanks to @oharboe for reporting this bug.


ekiwi:
fixes https://github.com/ucb-bar/chisel-testers2/issues/366 and https://github.com/ucb-bar/chisel-testers2/pull/367

chick (approved):
lgtm

PR #376, opened 2021-08-17
ci: reorganize into more independent jobs

ekiwi:
This also cuts down on the number of `mac` jobs we run because those machines seem to be harder to get.


chick (approved):
Looks like a clean implementation

PR #378, opened 2021-08-18
verilator: add re-use cache test

oharboe:


ekiwi:
Yeah, I don't think it ever did re-use binaries across tests. I think Ryan's use case was to re-use a binary that was created _the last time_ he ran a particular test. Re-use across tests is more tricky since we need to make sure to properly synchronize things between multiple threads.

oharboe:
> Yeah, I don't think it ever did re-use binaries across tests. I think Ryan's use case was to re-use a binary that was created _the last time_ he ran a particular test. Re-use across tests is more tricky since we need to make sure to properly synchronize things between multiple threads.

But it is needed to make dent into CI build times...

ekiwi:
> But it is needed to make dent into CI build times...

OK. Just wanted to make sure that I understood correctly that this is a feature request and not a regression.

oharboe:
> > But it is needed to make dent into CI build times...
> 
> OK. Just wanted to make sure that I understood correctly that this is a feature request and not a regression.

Yes... it's a feature, but a pretty basic one.

Currently caching will *never* happen on CI servers as no test is run twice. At least I can't imagine why. 

The only use-case where caching will happen is when one is iterating on the test-code for a *single* test in a local edit, compile, debug workflow 

For C++ and Scala tests, there is a very clear separation of concerns between *building* and *running*.

This separation of concerns is lacking in Chisel and it's causing headaches...

The lack of separation of concerns forces me to throttle CI testing speed even more, by trying to running only some of our tests in parallel:

The problem is CI servers and memory consumption. Verilator can be extremely memory hungry when running integration tests, whereas for unit-tests, Verilator memory consumption is no big deal.

If I can have control over which tests I pre-build Verilator for, then I can pre-build Verilator binaries that use tens of gigabytes in serial and do the rest in parallel.

ekiwi:
> Yes... it's a feature, but a pretty basic one.

> If I can have control over which tests I pre-build Verilator for, then I can pre-build Verilator binaries that use tens of gigabytes in serial and do the rest in parallel.

Sounds like you have a very specific use case in mind here. Instead of just making sure that there is no unnecessary compilation of duplicate simulators for the same, you want to be able to pre-build Verilator simulations.
All in all, I think caching is a great feature, but it will also be tricky to get right. (Apparently the first attempt which took a Master's student several months wasn't all that successful).

So to manage expectations: I don't think I will have time to implement a more advanced caching mechanism that what we have right now before the next release. If you want to give it a shot yourself, I should have time now and then to look at a PR.

oharboe:
> > Yes... it's a feature, but a pretty basic one.
> 
> > If I can have control over which tests I pre-build Verilator for, then I can pre-build Verilator binaries that use tens of gigabytes in serial and do the rest in parallel.
> 
> Sounds like you have a very specific use case in mind here. Instead of just making sure that there is no unnecessary compilation of duplicate simulators for the same, you want to be able to pre-build Verilator simulations.
> All in all, I think caching is a great feature, but it will also be tricky to get right. (Apparently the first attempt which took a Master's student several months wasn't all that successful).
> 
> So to manage expectations: I don't think I will have time to implement a more advanced caching mechanism that what we have right now before the next release. If you want to give it a shot yourself, I should have time now and then to look at a PR.

I did go on, didn't I? :-) Yes, there's more to continous integration and making this feature scale.

However, the basic feature that is missing is that caching will not have any effect in a Continuous Integration (build server) scenario, as only tests run more than once will be cached. That doesn't happen in Continous Integration.

I don't think Continous Integration can be classified as anything but a base use-case these days.

But... I'm glad the infrastructure is there. It does have value to put it into the release in that it will enable more people to try it out and help write PRs.

ekiwi:
This is probably not going to be merged any-time soon, unless there is a volunteer to implement the requested feature,

PR #383, opened 2021-08-19
remove old compatibility package

ekiwi:
The rename to chiseltest was done two years and three minor versions ago in PR #93.


ekiwi:
Is this something we can do?

chick:
There are usages in the wild of `import chisel3.tester`. Most that I dug deeper into are a repos that have been updated in a year and half or more. There's also a few gists laying around with it in them.
Probably the correct thing to do is deprecate everything in that package and remove it later
@ducky64 what do you think?

ekiwi:
>  Probably the correct thing to do is deprecate everything in that package and remove it later

Sounds good. By the new Chisel deprecation guidelines, we would also be aloud to backport the deprecation and to remove it from the next release. At least that is my understanding.

ducky64:
We should definitely have a deprecation phase before just removing it outright.

ekiwi:
Just for reference, the deprecation policy can be found here: https://github.com/chipsalliance/chisel3/discussions/2055#discussioncomment-1067786

ekiwi:
deprecation PR is up: https://github.com/ucb-bar/chiseltest/pull/472

PR #384, opened 2021-08-22
step() example

Martoni:
The simple example need step() no ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,21 @@ You can also build ChiselTest locally with `publishLocal`.
 ### Writing a Test
 ChiselTest integrates with the [ScalaTest](http://scalatest.org) framework, which provides a framework for detection and execution of unit tests.
 
-Assuming a typical Chisel project, create a new file in `src/test/scala/`, for example, `BasicTest.scala`.
+Assuming a typical Chisel project with `MyModule` defined in `src/main/scala/MyModule.scala`:
+
+```scala
+class MyModule extend Module {
+    val io = IO(new Bundle {
+        val in = Input(UInt(16.W))
+        val out = Output(UInt(16.W))
+    })
+
+    io.out := io.in
+}
+```
+
+Create a new file in `src/test/scala/`, for example, `BasicTest.scala`.
+
 
 In this file:
 1.  Add the necessary imports:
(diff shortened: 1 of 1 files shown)


edwardcwang:
Actually, where is the definition of `MyModule`? If `MyModule` is combinational, then there is actually no need for a `step()`.

Martoni:
> If `MyModule` is combinational, then there is actually no need for a `step()`.

Of course, but there is no example using step() in the readme. It can be good place to show step() a call no ?

edwardcwang:
Maybe we need to show an example implementation of `MyModule` (which is currently left to the imagination)?

Martoni:
Yes it can be a good Idea. We can also add an example of peek() call, which is not really complicated but could be good for beginners.

README.md
```diff
@@ -36,7 +36,21 @@ You can also build ChiselTest locally with `publishLocal`.
...
+Assuming a typical Chisel project with `MyModule` defined in `src/main/scala/MyModule.scala`:
+
+```scala
+class MyModule extend Module {
+    val io = IO(new Bundle {
+        val in = Input(UInt(16.W))
+        val out = Output(UInt(16.W))
+    })
+
+    io.out := io.in
```
ekiwi, line 48: could you add a register here and adjust the test sequence accordingly?

```
  io.out := RegNext(io.in)
```

README.md
```diff
@@ -72,10 +86,13 @@ In this file:
     The argument to the test stimulus block (`c` in this case) is a handle to the module under test.
 5.  In the test body, use `poke`, `step`, and `expect` operations to write the test:
     ```scala
-    c.in.poke(0.U)
-    c.out.expect(0.U)
-    c.in.poke(42.U)
-    c.out.expect(42.U)
+    c.io.in.poke(0.U)
```
ekiwi, line 89: Once we add the register this needs to be changed to:

```.scala
    c.io.in.poke(0.U)
    c.clock.step()
    c.io.out.expect(0.U)
    c.io.in.poke(42.U)
    c.clock.step()
    c.io.out.expect(42.U)
```

Note that the canonical version for a single step is: `.step()` and not `.step(1)`.

ekiwi (changes requested):
Looks good, but adding a register would make it clear why the `step` is needed.

README.md
Martoni: > Note that the canonical version for a single step is: `.step()` and not `.step(1)`.

This is typically something that should be included in the documentation :)

README.md
```diff
@@ -72,10 +86,13 @@ In this file:
...
+    c.clock.step()
+    c.io.out.expect(0.U)
+    c.io.in.poke(42.U)
+    c.clock.step()
+    c.io.out.expect(42.U)
+    println("Last output value :" + c.io.out.peek())
```
ekiwi, line 95: Sorry that I did not catch this last time, but ideally this should be `c.io.out.peek().litValue()` in order to demonstrate to users how to get a scala value.

ekiwi (changes requested):

README.md
Martoni: `c.io.out.peek().litValue()` or `c.io.out.peek().litValue` ?

ekiwi (approved):
LGTM Thanks!

PR #386, opened 2021-08-23
formal: add a past(..., N) function for assert/assume statements

ekiwi:
This allows users to write assertions/assumptions about past values without having to manually keep track of whether they are valid or not. An automated analysis pass figures out the maximum `Past` delay and delays the assertion for the right amount of cycles after reset.

### DIFF
--- a/src/main/scala/chiseltest/formal/FirrtlUtils.scala
+++ b/src/main/scala/chiseltest/formal/FirrtlUtils.scala
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+package chiseltest.formal
+
+import chisel3.util.log2Ceil
+import firrtl._
+
+/** firrtl utility functions used in our passes.
+  * At some point, these could be moved into the firrtl repo.
+  */
+private object FirrtlUtils {
+  def plusOne(e: ir.Expression): ir.Expression = {
+    val width = e.tpe.asInstanceOf[ir.UIntType].width.asInstanceOf[ir.IntWidth].width
+    val addTpe = ir.UIntType(ir.IntWidth(width + 1))
+    val add = ir.DoPrim(PrimOps.Add, List(e, ir.UIntLiteral(1, ir.IntWidth(width))), List(), addTpe)
+    ir.DoPrim(PrimOps.Bits, List(add), List(width - 1, 0), e.tpe)
+  }
+
+  def findClockAndReset(m: ir.Module): (ir.Reference, ir.Reference) = {
+    val clock = m.ports
+      .find(_.name == "clock")
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named clock!")
+      )
+    val reset = m.ports
+      .find(_.name == "reset")
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named reset!")
+      )
+    (ir.Reference(clock).copy(flow = SourceFlow), ir.Reference(reset).copy(flow = SourceFlow))
+  }
+
+  /** Creates a register in canonical form from LoFirrtl expressions. */
+  def makeRegister(
+    info:  ir.Info,
+    name:  String,
+    clock: ir.Expression,
+    reset: ir.Expression,
+    next:  Option[ir.Expression],
+    init:  Option[ir.Expression]
+  ): (ir.Reference, ir.DefRegister, ir.Statement) = {
+    require(clock.tpe == ir.ClockType, s"Invalid clock expression: ${clock.serialize} : ${clock.tpe.serialize}")
+    val hasAsyncReset = reset.tpe match {
+      case ir.AsyncResetType => true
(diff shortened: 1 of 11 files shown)


ekiwi:
I talked to the author of https://github.com/danielkasza/dank-formal/ and he pointed out to me that at least `||` needs to get a special handling. Essentially if our assertion/assumption is a disjunction, we might only want to disable individual terms and not the complete assertion/assumption.

src/main/scala/chiseltest/formal/past.scala
```diff
@@ -0,0 +1,242 @@
...
+  * by the time necessary for all past values to become available.
+  * E.g.:
+  * @example {{{
+  *   assert(past(out) == past(past(in)))
+  * }}}
+  * is equivalent to:
+  * @example {{{
+  *   when(cyclesSinceReset >= 2.U) {
+  *     assert(RegNext(out) == RegNext(RegNext(in)))
+  *   }
```
jackkoenig, line 26:
```suggestion
  *   assert(past(out) === past(past(in)))
  * }}}
  * is equivalent to:
  * @example {{{
  *   when(cyclesSinceReset >= 2.U) {
  *     assert(RegNext(out) === RegNext(RegNext(in)))
  *   }
```

schoeberl:
BTW, we worked on something similar within the ChiselVeryfy library: https://github.com/chiselverify/chiselverify/wiki/Timed-Assertions

ekiwi:
> BTW, we worked on something similar within the ChiselVeryfy library: https://github.com/chiselverify/chiselverify/wiki/Timed-Assertions

Thanks for pointing this out, @schoeberl !

For anyone following along, @schoeberl 's timed assertion are a great solution if you want to add assertions to your (concrete) chiseltest test. This PR is working on timed assertions for formal verification (but they should also work in simulation).

src/main/scala/chiseltest/formal/past.scala
ekiwi: fixed

ekiwi:
I have grown fairly confident that the `past` function in the PR is a good way forward.
Most of the lines added in this PR are the numerous examples that I wrote to try out the `past` function and they convince me that we should merge this. Do you agree @jackkoenig ?

src/main/scala/chiseltest/formal/FirrtlUtils.scala
```diff
@@ -0,0 +1,101 @@
...
+      .find(_.name == "clock")
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named clock!")
+      )
+    val reset = m.ports
+      .find(_.name == "reset")
+      .getOrElse(
+        throw new RuntimeException(s"[${m.name}] Expected module to have a port named reset!")
+      )
+    (ir.Reference(clock).copy(flow = SourceFlow), ir.Reference(reset).copy(flow = SourceFlow))
```
jackkoenig, line 29: I don't love this, I'd rather use analysis to find the clock and reset based on what registers use but I'll accept it for this initial version.

jackkoenig (approved):
This is fantastic. I'm not 100% on all of the code but the tests are some of the most thorough I've ever seen and really illustrate the utility here.

src/main/scala/chiseltest/formal/FirrtlUtils.scala
ekiwi: I was thinking about that, but since most of `chiseltest` does not really support multi-clock or multi-reset circuits, including our formal verification infrastructure, I don't think it makes sense to solve this problem in this PR.
Once we move `past` to Chisel, we can think a bit more about how to support multi-clock scenarios. However, I do think we need a comprehensive clock domain analysis pass before we can make this a reality, since we would want to be able to catch errors where the `past` delay registers are on a different clock domain then the assertion. That would require a global analysis since two clock inputs could be connected to the same or different clock domain outside of the current module. I do have a [PR for the prototype of a global clock and reset analysis](https://github.com/chipsalliance/firrtl/pull/2252). But to solve the problem, I think we would first have to tighten up our type system a bit and maybe restrict the an analysis to clock signals and _async_ resets.

sequencer:
@ekiwi any idea to port this feature to chisel3? I think this API is really useful.

ekiwi:
> @ekiwi any idea to port this feature to chisel3? I think this API is really useful.

There is no reason why we couldn't just copy the implementation over to Chisel besides it being a new-ish API and maybe some people being skeptical. So if you want to have it in chisel, feel free to make a PR with the code copied over from here. Should only be one file (+ the tests).

Btw. I now have a workshop paper that includes a description of the "safe" `past` functionality: https://woset-workshop.github.io/WOSET2021.html#article-3

ekiwi:
Oh, just remembered: The one problem is that in its current form it only works for Chisel modules that do not use `withClock`, i.e., it assumes one default clock domain in the module. We could try to work on fixing that in the PR. The easiest solution might just be to warn people not to use it when they have more than one clock domain.

PR #390, opened 2021-09-04
formal: fix trace replay with arbitrary values

ekiwi:
Previously, the registers we used to represent DefRandom nodes in treadle were constant propagated through which would result in an incorrect VCD when the bug involved arbitrary values chosen by the formal engine.


chick (changes requested):
Looks like 2.13 Seq issues need to be fixed.
Otherwise it looks like a nice clean change

chick (approved):
LGTM

PR #391, opened 2021-09-08
internal: add PrintPeekPoke annotation for debugging

ekiwi:
Some convenient functionality that I recently used to track down a bug. Might come in handy to others and it is part of the `internal` package, so it won't increase our public API surface.

### DIFF
--- a/src/main/scala/chiseltest/simulator/DebugPrintWrapper.scala
+++ b/src/main/scala/chiseltest/simulator/DebugPrintWrapper.scala
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+package chiseltest.simulator
+
+/** Wraps a [[SimulatorContext]] and prints the result of all operations performed on
+  * it to stdout
+  */
+class DebugPrintWrapper(inner: SimulatorContext) extends SimulatorContext {
+  private var cycle: Long = 0
+  override def sim = inner.sim
+  override def step(n: Int): Unit = {
+    val start = cycle
+    cycle += n
+    inner.step(n)
+    println(s"step $start -> $cycle")
+  }
+  override def peek(signal: String) = {
+    val res = inner.peek(signal)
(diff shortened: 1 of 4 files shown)


chick (approved):
Looks clean, and useful.

src/main/scala/chiseltest/simulator/DebugPrintWrapper.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+  override def peekMemory(signal: String, index: Long) = {
+    val res = inner.peekMemory(signal, index)
+    println(s"$signal[$index] -> $res")
+    res
+  }
+
+  override def pokeMemory(signal: String, index: Long, value: BigInt): Unit = {
+    println(s"$signal[$index] <- $value")
+    inner.pokeMemory(signal, index, value)
```
ducky64, line 33: Do you want to be consistent in whether the print happens before or after the pokes / other actions?

src/main/scala/chiseltest/simulator/DebugPrintWrapper.scala
```diff
@@ -0,0 +1,49 @@
...
+    println("finish()")
+  }
+  override def resetCoverage(): Unit = {
+    inner.resetCoverage()
+    println("resetCoverage()")
+  }
+  override def getCoverage() = {
+    val res = inner.getCoverage()
+    println("getCoverage()")
+    // println(res.mkString("\n"))
```
ducky64, line 46: dead code

ducky64 (approved):
printfs were also useful for me in development, and this is a pretty architecturally clean way to do it!

Some nits in the comments though.

src/main/scala/chiseltest/simulator/DebugPrintWrapper.scala
ekiwi: good catch

PR #399, opened 2021-09-14
Remove final step from generic backend main loop.

chick:
Fixes problem where assert fires due to exiting of default timescope
Fixed couple of nits to make code clean.

### DIFF
--- a/src/test/scala/chiseltest/experimental/tests/AssertTest.scala
+++ b/src/test/scala/chiseltest/experimental/tests/AssertTest.scala
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest.experimental.tests
+
+import chisel3._
+import chisel3.tester._
+import org.scalatest.freespec.AnyFreeSpec
+
+class AssertFail extends Module {
+  val io = IO(new Bundle {
+    val input = Input(UInt(2.W))
+  })
+
+  assert(io.input === 3.U) // This should not be triggered
+}
+
+/*
+Prior to this fix this test would fail because
+the GenericBackend would reset the inputs to their initial values
+and then call a final step and the assert would fire.
+Generic backend no longer calls the final step
+ */
+class AssertTest extends AnyFreeSpec with ChiselScalatestTester {
+  "Internally test resets input but it should not call a final step, as that incorrectly triggers the assert" in {
+    test(new AssertFail).withAnnotations(Seq()) { dut =>
+      dut.io.input.poke(3.U)
+      dut.clock.step()
+    }
+  }
+}
--- a/src/main/scala/chiseltest/internal/GenericBackend.scala
+++ b/src/main/scala/chiseltest/internal/GenericBackend.scala
@@ -4,7 +4,7 @@ package chiseltest.internal
 
 import chiseltest.{ClockResolutionException, Region, TimeoutException}
 import chisel3._
-import chiseltest.coverage.{Coverage, TestCoverage}
+import chiseltest.coverage.TestCoverage
 import chiseltest.simulator.SimulatorContext
 import firrtl.AnnotationSeq
 
@@ -147,7 +147,7 @@ class GenericBackend[T <: Module](
     val mainThread = new TesterThread(
       () => {
         tester.poke("reset", 1)
-        tester.step(1)
+        tester.step()
         tester.poke("reset", 0)
 
         testFn(dut)
@@ -191,7 +191,9 @@ class GenericBackend[T <: Module](
           }
         }
 
-        tester.step(1)
+        if (!mainThread.done) {
+          tester.step()
+        }
       }
     } finally {
       rootTimescope = None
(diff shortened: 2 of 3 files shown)


src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
...
+}
+
+class AssertTest extends AnyFreeSpec with ChiselScalatestTester {
+  "break" in {
+    test(new AssertFail).withAnnotations(Seq()) { dut =>
+      dut.io.input.poke(3.U)
+      dut.clock.step()
+    }
+  }
+}
```
ducky64, line 27: Need newline at end of file

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
...
+import chisel3.tester._
+import chiseltest.VerilatorBackendAnnotation
+import org.scalatest.freespec.AnyFreeSpec
+
+class AssertFail extends Module {
+  val io = IO(new Bundle {
+    val input = Input(UInt(2.W))
+  })
+
+  printf("io.input is %d\n", io.input) // prints 3 and then 0
```
ducky64, line 15: Unit tests should not print

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  assert(io.input === 3.U) // fails, apparently on the second cycle
```
ducky64, line 17: Well, it shouldn't fail with this fix?

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
...
+}
+
+class AssertTest extends AnyFreeSpec with ChiselScalatestTester {
+  "break" in {
```
ducky64, line 21: It shouldn't break?

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest.experimental.tests
+
+import chisel3._
+import chisel3.tester._
+import chiseltest.VerilatorBackendAnnotation
```
ducky64, line 7: Does this need Verilator?

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,27 @@
...
+  val io = IO(new Bundle {
+    val input = Input(UInt(2.W))
+  })
+
+  printf("io.input is %d\n", io.input) // prints 3 and then 0
+
+  assert(io.input === 3.U) // fails, apparently on the second cycle
+}
+
+class AssertTest extends AnyFreeSpec with ChiselScalatestTester {
```
ducky64, line 20: I think this test case needs a comment, because the behavior it's testing is non-obvious. Something along the lines of that because pokes go out of scope at the end of the main test body, there should be no trailing clock edge (and no asserts evaluated) after the test ends.

ducky64 (changes requested):
Mostly looks good, a few style nits though. Most important is that the test case needs a comment, because the bug tested is not obvious.

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
chick: Oops, fixed

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
chick: Changed comment to say it should not fail

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
chick: Agreed. Fixed!

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
chick: It does not, import removed

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
chick: Good point, better comment added

ekiwi:
The coverage test needs to be fixed. Since we have one fewer `step` now, that means that some of the counts will be off by one.

chick:
I just came to the same conclusion

ekiwi (approved):
LGTM

chick:
@ducky64 I think I got all your change requests

src/test/scala/chiseltest/experimental/tests/AssertTest.scala
```diff
@@ -0,0 +1,30 @@
...
+class AssertFail extends Module {
+  val io = IO(new Bundle {
+    val input = Input(UInt(2.W))
+  })
+
+  assert(io.input === 3.U) // This should not be triggered
+}
+
+/*
+Prior to the the fix that this is test would fail because
```
ducky64, line 18:
```suggestion
Prior to the the fix that this is the test would fail because
```

ducky64 (approved):
One typo, otherwise looks good.

PR #407, opened 2021-10-12
Add Observer trait and function for Boringtools submodule drilling

carlosedp:
This PR adds the Observer trait that can be used on wrapper modules for looking into submodules on chiseltest.

The wrapper makes sure the observed objects are only created in the object instantiation during the test and not Verilog generation (creating the observed wires).

## Usage

Create a Chisel module that instantiates a submodule as usual:

```scala
import chisel3._

// This is the module and submodule to be tested
class SubModule extends Module {
  val reg  = Reg(UInt(6.W))
  reg := 42.U
}

// Top module which instantiates the submodule
class Top extends Module {
  val submodule = Module(new SubModule)
}

// Generate the main project Verilog
object Proj extends App {
 (new chisel3.stage.ChiselStage).emitVerilog(
   new Top()
)}
```

In the test, wrap the top module adding the `Observer` trait and inside it use the `observe` method pointing to submodule objects. Use this wrapper as the DUT and the observed objects will be available for testing:

```scala
import chisel3._
import chiseltest._
import org.scalatest._

import flatspec._
import matchers.should._
import chiseltest.experimental.Observer

// Here we create a wrapper extending the Top module to add the `Observer` trait
class TopWrapper extends Top with Observer {
  // Observe the submodule "reg" Register
  val observed_reg  = observe(submodule.reg)
}

it should "observe a submodule Reg by using BoringUtils" in {
  test(new TopWrapper) { c =>
    c.observed_reg.expect(42.U)
  }
}
```

### DIFF
--- a/src/main/scala/chiseltest/experimental/Observe.scala
+++ b/src/main/scala/chiseltest/experimental/Observe.scala
@@ -55,7 +55,7 @@ object expose {
     * import chiseltest.experimental._
     *
     * // Here we create a wrapper extending the Top module adding the exposed signals
-    * class TopWrapper extends Top with Observer {
+    * class TopWrapper extends Top {
     *   // Expose the submodule "reg" Register
     *   val exposed_reg  = expose(submodule.reg)
     * }
--- a/src/main/scala/chiseltest/experimental/Observe.scala
+++ b/src/main/scala/chiseltest/experimental/Observe.scala
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package chiseltest.experimental
+
+import chisel3._
+import chisel3.util.experimental._
+
+/** Add this trait to your wrapper module to observe on submodule signals using the `observe` method. This avoids
+  * cluttering the original module with signals that would be used only by the test harness.
+  */
+trait Observer { this: Module =>
+
+  /** The method `observe` allows one to bring to the wrapped module (top), signals from instantiated sub-modules so
+    * they can be tested by peek/poke chiseltest.
+    *
+    * Usage:
+    *
+    * {{{
+    * import chiseltest.experimental.Observer
+    *
+    * // This is the module to be tested
+    * class SubModule extends Module {
+    *     val reg  = Reg(UInt(6.W))
+    *     reg := 42.U
+    *   }
+    *
+    *   // Top module which instantiates the submodule
+    *   class Top extends Module {
+    *     val submodule = Module(new SubModule)
+    *   }
+    *
+    *   // Here we create a wrapper extending the Top module adding the observer
+    *   class TopWrapper extends Top with Observer {
+    *     val observed_reg  = observe(submodule.reg)
+    *   }
+    *
+    * }}}
+    *
+    * The `Top` module can be tested with `chiseltest` and `scalatest` like:
+    * {{{
+    *   it should "observe a submodule Reg by using BoringUtils" in {
+    *     test(new TopWrapper) { c =>
+    *       c.observed_reg.expect(42.U)
+    *     }
+    *   }
+    * }}}
+    *
+    * @param signal
+    *   the signal to be observed
+    * @return
+    *   a signal with the same format to be tested on Top module's spec.
+    */
+  protected def observe[T <: Data](signal: T): T = {
+    val ob = IO(Output(chiselTypeOf(signal)))
+    ob := 0.U.asTypeOf(chiselTypeOf(signal))
+    BoringUtils.bore(signal, Seq(ob))
+    ob
+  }
+}
--- a/src/main/scala/chiseltest/experimental/Observe.scala
+++ b/src/main/scala/chiseltest/experimental/Observe.scala
@@ -16,33 +16,48 @@ trait Observer { this: Module =>
     * Usage:
     *
     * {{{
-    * import chiseltest.experimental.Observer
+    * import chisel3._
     *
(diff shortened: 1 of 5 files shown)


sequencer:
I personally suggest using VPI and chipsalliance/chisel3#2077 to achieve this feature.

ekiwi:
Short status update: @carlosedp I think that you are correct in that we should not make this more complicated than we have to.

My thinking right now:
- do we actually need this to be a trait? We could also just make `observe` a free function that people can call as a somewhat nicer interface to BoringUtils
- I would prefer if `observe` did not create an I/O by default. Sometimes you just need to observe an internal signal in order to write a property about it: https://github.com/ucb-bar/chisel-testers2/blob/b990df45adaa4e69d9c49029ae705ddef4ddd8fe/src/test/scala/chiseltest/formal/examples/VGBComparisonOfFormalAndSimulation.scala#L163
- maybe we can also add an `expose` function which calls `observe` and adds an output port

carlosedp:
Do you think this is a "go"? :)

ekiwi:
> Do you think this is a "go"? :)

Oh, sorry @carlosedp, my research is keeping my busy so I forgot about this PR.

I would prefer if you could create a `observe` and `expose` function, where `observe` creates a `Wire` and only `expose` will create an `IO` port. You can create the function with an `object` and the `apply` method like:
```.scala
object observe {
  def apply[T <: Data](signal: T): T = {
    val wire = WireInit(0.U.asTypeOf(chiselTypeOf(signal)))
    BoringUtils.bore(signal, Seq(wire))
    wire
  }
}

object expose {
  def apply[T <: Data](signal: T): T = {
    val ob = IO(Output(chiselTypeOf(signal)))
    ob := 0.U.asTypeOf(chiselTypeOf(signal))
    BoringUtils.bore(signal, Seq(ob))
    ob
  }
}
```

carlosedp:
I think this is way better. Not needing to use the Observer trait makes things clearer.

carlosedp:
Would you give me a heads-up on the usage of the `obseve` function? I'd add a use case for it too.

ekiwi:
> Would you give me a heads-up on the usage of the `obseve` function? I'd add a use case for it too.

That would be awesome. A good starting point would be to:
- delete the `Observer` trait here: https://github.com/ucb-bar/chisel-testers2/blob/b990df45adaa4e69d9c49029ae705ddef4ddd8fe/src/test/scala/chiseltest/formal/examples/VGBComparisonOfFormalAndSimulation.scala#L162
- you could use this as an example of how to use `observe`: https://github.com/ucb-bar/chisel-testers2/blob/b990df45adaa4e69d9c49029ae705ddef4ddd8fe/src/test/scala/chiseltest/formal/examples/VGBComparisonOfFormalAndSimulation.scala#L35

carlosedp:
Done, I've replaced the previous `Observer` trait in two places.

src/main/scala/chiseltest/experimental/Observe.scala
```diff
@@ -0,0 +1,81 @@
...
+    *
+    * // Generate the main project Verilog
+    * object Proj extends App {
+    *  (new chisel3.stage.ChiselStage).emitVerilog(
+    *    new Top()
+    *  )}
+    * }}}
+    *
+    * Then in your spec, test the `Top` module with `chiseltest` and `scalatest` while being
+    *      able to observe the submodule:
```
ekiwi, line 47: should be something like: "while exposing the signal from the submodule to the testbench"

src/main/scala/chiseltest/experimental/Observe.scala
```diff
@@ -0,0 +1,81 @@
...
+    * import matchers.should._
+    * import chiseltest.experimental._
+    *
+    * // Here we create a wrapper extending the Top module adding the exposed signals
+    * class TopWrapper extends Top {
+    *   // Expose the submodule "reg" Register
+    *   val exposed_reg  = expose(submodule.reg)
+    * }
+    *
+    * it should "observe a submodule Reg by using BoringUtils" in {
```
ekiwi, line 63: "expose"

src/main/scala/chiseltest/experimental/Observe.scala
```diff
@@ -0,0 +1,81 @@
...
+    *   test(new TopWrapper) { c =>
+    *     c.exposed_reg.expect(42.U)
+    *   }
+    * }
+    * }}}
+    *
+    * @param signal
+    *   the signal to be observed
```
ekiwi, line 71: "exposed"

src/test/scala/chiseltest/experimental/tests/ObserveTest.scala
```diff
@@ -0,0 +1,58 @@
...
+import chiseltest.experimental.{expose}
+
+import chisel3._
+import chiseltest._
+import org.scalatest._
+
+import flatspec._
+import matchers.should._
+
+class ObserverTest extends AnyFlatSpec with ChiselScalatestTester with Matchers {
```
ekiwi, line 14: Please rename to `ObserveExposeTests` (class and filename)

ekiwi (changes requested):
Four small changes and then we can get this in!

ekiwi (approved):
LGTM :shipit:

carlosedp:
Something went wonky with the master tests...
https://github.com/ucb-bar/chisel-testers2/runs/4072541570?check_suite_focus=true

PR #410, opened 2021-10-17
verilator: reduce logging spam

oharboe:


ekiwi:
Thanks @oharboe ! We will have a more complete fix for this issue in: https://github.com/ucb-bar/chisel-testers2/pull/413

Issue #408, opened 2021-10-15
Width inference error for SInt when using verilator backend


SharzyL:
When running the following test using testers2:
```scala
import chisel3._
import chisel3.tester.{ChiselUtestTester, testableClock, testableData}
import chiseltest.simulator.{VerilatorBackendAnnotation, WriteVcdAnnotation}
import utest._

class Foo extends Module {
  val a: SInt = IO(Output(SInt(3.W)))
  a := 7.U(3.W).asSInt
}

object WhateverTest extends TestSuite with ChiselUtestTester {
  val tests: Tests = Tests {
    test("whatever") {
      testCircuit(new Foo, Seq(VerilatorBackendAnnotation)) { dut =>
        dut.a.peek()
      }
    }
  }
}
```

It emits the following error:
```
requirement failed: The literal value 7 was elaborated with a specified width of 3 bits, but at least 4 bits are required.
java.lang.IllegalArgumentException: requirement failed: The literal value 7 was elaborated with a specified width of 3 bits, but at least 4 bits are required.
	at scala.Predef$.require(Predef.scala:281)
	at chisel3.internal.firrtl.LitArg.<init>(IR.scala:111)
	at chisel3.internal.firrtl.SLit.<init>(IR.scala:130)
	at chisel3.SIntFactory.Lit(SIntFactory.scala:20)
	at chisel3.SIntFactory.Lit$(SIntFactory.scala:19)
	at chisel3.package$SInt$.Lit(package.scala:180)
	at chisel3.package$fromBigIntToLiteral.asSInt(package.scala:70)
	at chiseltest.package$testableData.peekWithStale(package.scala:186)
	at chiseltest.package$testableData.peek(package.scala:210)
	at multiplier.WhateverTest$.$anonfun$tests$10(WhateverTester.scala:36)
	at multiplier.WhateverTest$.$anonfun$tests$10$adapted(WhateverTester.scala:35)
	at chiseltest.internal.GenericBackend.$anonfun$run$1(GenericBackend.scala:166)
	at chiseltest.internal.ThreadedBackend$TesterThread$$anon$1.$anonfun$run$1(ThreadedBackend.scala:495)
	at chiseltest.internal.GenericBackend.doTimescope(GenericBackend.scala:111)
	at chiseltest.internal.ThreadedBackend$TesterThread$$anon$1.run(ThreadedBackend.scala:495)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

[...]

PR #415, opened 2021-10-20
simulator: fix reading/writing signed ints on verilator + iverilog

ekiwi:
This should fix: https://github.com/ucb-bar/chisel-testers2/issues/408


chick (approved):
Looks good, nice work!

PR #421, opened 2021-10-22
Change in behavior in assertion firing

jackkoenig:
In chiseltest 0.3.x the following test would fail due to the assertion firing. On master, it does not fail. I'm not sure that this is a bug, but it certainly is a change in behavior worth noting.


jackkoenig:
Yeah our code is (well was) definitely wrong. The weird bit is that it _used_ to fire. Unclear why, but perhaps that was the bug all along.

PR #424, opened 2021-10-22
Add support for plusargs to Verilator backend

jackkoenig:
This isn't super thoroughly tested and I haven't yet added support to Icarus Verilog, but I wanted to get the ball rolling as this is a badly needed feature by SiFive.


jackkoenig:
> This looks good. Could I ask you to also add support to the treadle backend?

Done

> Most of the work would be adding a test for treadle

Accurate prediction 🙂

PR #427, opened 2021-10-28
Assert behavior tests

chick:
chiseltest's behavior on asserts
with treadle vs verilator
@jackkoenig tests in src/test/scala/chiseltest/experimental/tests/AssertTest.scala
all pass. Can you give me a hint as to what case I am missing


ekiwi:
Seems like all your tests pass (once they are properly tagged with `RequiresVerilator`). What were you trying to demonstrate?
@jackkoenig @chick

PR #431, opened 2021-10-29
vcs: run binary from os.pwd

ekiwi:
The path to the simulator binary and the waveform was updated to work when running the simulation from os.pwd.

@jiegec, could you please check if this fixes your issue? This PR should address the wrong binary path as well as the misplaced wave dump.


chick (approved):
lgtm

PR #435, opened 2021-11-04
formal: run firrtl optimizations before generating STMLib

ekiwi:
For the examples that we have in the repo, the SMT solving isn't really getting any faster, but the SMT files we generate are somewhat smaller. For a bigger example (ShapeProcessorProps.smt2) it goes from 114K to 57K.

This PR also adds an escape hatch annotation in case the optimizations turn out not to be a good idea, but I think they should be on by default.


chick (approved):
LGTM

PR #441, opened 2021-11-10
ci: test btormc on both linux and mac

ekiwi:


ekiwi:
Looks like the official action to install oss-cad-suite only supports linux :(

jackkoenig (approved):

PR #444, opened 2021-11-16
Implements Hardware testing allowing test suite to detect exceptions

chick:
- Create new low level `runReturningExceptions` that can return `ChiselAssertionError`s
- Optimize step calling by using step parameter and tester.cycleCount
- Add `runReturningExceptions` to `ChiselScalatestTester`
- Add tests for new function with emphasis on precedence of simultaneous outcomes
- Fix to scaladoc Unknown tag
- Called step without specifying default argument


ekiwi (changes requested):
Hi @chick, thanks for your contribution!

Could you please remove the changes to `TreadleSimulator.scala` from this PR?
I think there is a bug and it would be easier to debug that if this change was in a separate PR.

ekiwi:
I am going to close this for now since https://github.com/ucb-bar/chisel-testers2/pull/445 was merged.
If you want to add your change to the Treadle interface, then please open another PR for that.

PR #448, opened 2021-11-20
Update Readme to reflect necessary imports and use of AnyFlatSpec

michael-etzkorn:
When I first started using chiseltest, I ran into some issues with `FlatSpec` and imports. Hopefully these imports and using `AnyFlatSpec` instead will help with any import issues people might encounter as they start using chiseltest.


README.md
```diff
@@ -55,18 +55,20 @@ Create a new file in `src/test/scala/`, for example, `BasicTest.scala`.
...
 1.  Add the necessary imports:
     ```scala
-    import org.scalatest._
-    import chiseltest._
-    import chisel3._
+   import org.scalatest._
+   import org.scalatest.flatspec._ 
+   import matchers.should._
+   import chiseltest._
+   import chisel3._
```
carlosedp, line 62: Scalafix organize-imports usually formats this as:

```
import chisel3._
import chiseltest._
import org.scalatest._

import flatspec._
import matchers._
```

I believe it's more clear.

README.md
```diff
@@ -55,18 +55,20 @@ Create a new file in `src/test/scala/`, for example, `BasicTest.scala`.
...
     ```
 2.  Create a test class:
     ```scala
-    class BasicTest extends FlatSpec with ChiselScalatestTester with Matchers {
+    class BasicTest extends AnyFlatSpec with ChiselScalatestTester with Matchers {
```
carlosedp, line 66: If previous review is accepted, this should be changed to:
`class BasicTest extends AnyFlatSpec with ChiselScalatestTester with should.Matchers {`

README.md
ekiwi: Personally I would prefer just:
```
import chisel3._
import chiseltest._
import org.scalatest.flatspec.AnyFlatSpec
```

michael-etzkorn: I like the minimal imports, and what about `Matchers`?

ekiwi: The whole `Matchers` thing is mostly a matter of taste...

If you do not have `Matchers`, you can still use regular asserts to perform checks in your tests + the `expect` function from chiseltest also provides you with a built-in assert. I personally believe that for new comers writing asserts is less confusing and if people want to they can always import `Matchers`, I just prefer not to include it by default.

In case this was confusing:
```.scala
// without matchers
assert(foo() == 42)
// with matchers
foo() should be (42)
```

README.md
michael-etzkorn: I'm going to use Kevin's suggested imports. I'll add something to the `README` in this pr about `Matchers` being an optional addition.

ekiwi (approved):
Thanks for cleaning this up!

Awesome first contribution.

:ship:

PR #453, opened 2021-12-02
allow poke and expect to accept BigInt and Boolean values

ekiwi:
I am curious what people think about allowing `UInt`, `SInt` and `Bool` ports to be poked / expected with a `BigInt` / `Boolean` value directly. I feel like that might make chiseltest a little easier to use.
If we decide to include this, I might look into allowing `Vec(_, UInt/SInt/Bool)` to be poked / expected with a Seq[BigInt].

### DIFF
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -136,6 +136,39 @@ package object chiseltest {
       Context().backend.pokeBits(signal, maskedValue)
     }
 
+    // Throw an exception if the value cannot fit into the signal.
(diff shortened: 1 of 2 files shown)


jackkoenig:
In general I think this is fine, I just wish we had a good way to deal with Bundles, and especially things like Vec of Bundles. We need to do some work in Chisel itself to make literals work better.

schoeberl:
I think this is already nice, independent of Bundles and Vecs. You can call expect on the individual fields.

ekiwi:
replaced by https://github.com/ucb-bar/chiseltest/pull/480

PR #454, opened 2021-12-02
Add peek<Type> as peek method that returns Scala types

carlosedp:
This PR adds a peek method that returns Scala types instead of Chisel types.

This makes a cleaner syntax avoiding the use of `peek().litValue` simplifying the use of scalatest or other test framework matchers.

The method name is a suggestion.

### DIFF
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -205,38 +205,57 @@ package object chiseltest {
       case x => throw new LiteralTypeException(s"don't know how to peek $x")
     }
 
-    protected def peekWithStaleNative(stale: Boolean): Any = x match {
+    protected def peekWithStaleBool(stale: Boolean): Boolean = x match {
       case (x: Bool) =>
         Context().backend.peekBits(x, stale) match {
           case x: BigInt if x == 0 => false
           case x: BigInt if x == 1 => true
           case x => throw new LiteralTypeException(s"peeked Bool with value $x not 0 or 1")
         }
+    }
+
+    protected def peekWithStaleInt(stale: Boolean): BigInt = x match {
       case (x: UInt) => Context().backend.peekBits(x, stale)
       case (x: SInt) => Context().backend.peekBits(x, stale)
+    }
+
+    protected def peekWithStaleDec(stale: Boolean): BigDecimal = x match {
       case (x: FixedPoint) => {
         val multiplier = BigDecimal(2).pow(x.binaryPoint.get)
         BigDecimal(Context().backend.peekBits(x, stale)) / multiplier
       }
-      case x: Interval =>
-        throw new LiteralTypeException(s"peeking literal Interval type is not yet supported")
-      case (x: Record) => {
-        throw new LiteralTypeException(s"peeking literal Record type is not yet supported")
-      }
-      case (x: Vec[_]) =>
-        val elementValueFns = x.getElements.map { case elt: Data =>
-          elt.peekWithStaleNative(stale)
-        }
-        Seq(elementValueFns: _*)
-      case (x: EnumType) => {
-        throw new NotImplementedError(s"peeking enums ($x) not yet supported, need programmatic enum construction")
-      }
-      case x => throw new LiteralTypeException(s"don't know how to peek $x")
     }
 
-    def peek(): T = peekWithStale(false)
+    protected def peekWithStaleVecInt(stale: Boolean): Seq[BigInt] = x match {
+      case (x: Vec[UInt]) => {
+        val elementValueFns = x.getElements.map { case elt: Data =>
+          elt.peekWithStaleInt(stale)
+        }
+        Seq(elementValueFns: _*)
+      }
+      case (x: Vec[SInt]) => {
+        val elementValueFns = x.getElements.map { case elt: Data =>
+          elt.peekWithStaleInt(stale)
+        }
+        Seq(elementValueFns: _*)
+      }
+    }
 
-    def asNative(): Any = peekWithStaleNative(false)
+    protected def peekWithStaleVecBool(stale: Boolean): Seq[Boolean] = x match {
+      case (x: Vec[Bool]) => {
+        val elementValueFns = x.getElements.map { case elt: Data =>
+          elt.peekWithStaleBool(stale)
+        }
+        Seq(elementValueFns: _*)
+      }
+    }
+
+    def peek():           T = peekWithStale(false)
+    def peekBool():       Boolean = peekWithStaleBool(false)
+    def peekInt():        BigInt = peekWithStaleInt(false)
+    def peekDecimal():    BigDecimal = peekWithStaleDec(false)
+    def peekVectorInt():  Seq[BigInt] = peekWithStaleVecInt(false)
+    def peekVectorBool(): Seq[Boolean] = peekWithStaleVecBool(false)
 
     protected def expectWithStale(value: T, message: Option[String], stale: Boolean): Unit = (x, value) match {
       case (x: Bool, value: Bool) =>
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -214,7 +214,7 @@ package object chiseltest {
         }
     }
 
-    protected def peekWithStaleInt(stale: Boolean): BigInt = x match {
+    protected def peekWithStaleBigInt(stale: Boolean): BigInt = x match {
       case (x: UInt) => Context().backend.peekBits(x, stale)
       case (x: SInt) => Context().backend.peekBits(x, stale)
     }
@@ -229,13 +229,13 @@ package object chiseltest {
     protected def peekWithStaleVecInt(stale: Boolean): Seq[BigInt] = x match {
       case (x: Vec[UInt]) => {
         val elementValueFns = x.getElements.map { case elt: Data =>
-          elt.peekWithStaleInt(stale)
+          elt.peekWithStaleBigInt(stale)
         }
         Seq(elementValueFns: _*)
       }
       case (x: Vec[SInt]) => {
         val elementValueFns = x.getElements.map { case elt: Data =>
-          elt.peekWithStaleInt(stale)
+          elt.peekWithStaleBigInt(stale)
         }
         Seq(elementValueFns: _*)
       }
@@ -250,12 +250,13 @@ package object chiseltest {
       }
     }
 
-    def peek():           T = peekWithStale(false)
-    def peekBool():       Boolean = peekWithStaleBool(false)
-    def peekInt():        BigInt = peekWithStaleInt(false)
-    def peekDecimal():    BigDecimal = peekWithStaleDec(false)
-    def peekVectorInt():  Seq[BigInt] = peekWithStaleVecInt(false)
-    def peekVectorBool(): Seq[Boolean] = peekWithStaleVecBool(false)
+    def peek():          T = peekWithStale(false)
+    def peekBool():      Boolean = peekWithStaleBool(false)
+    def peekBigInt():    BigInt = peekWithStaleBigInt(false)
+    def peekInt():       Integer = peekWithStaleBigInt(false).toInt
+    def peekDecimal():   BigDecimal = peekWithStaleDec(false)
+    def peekVecBigInt(): Seq[BigInt] = peekWithStaleVecInt(false)
+    def peekVecBool():   Seq[Boolean] = peekWithStaleVecBool(false)
 
     protected def expectWithStale(value: T, message: Option[String], stale: Boolean): Unit = (x, value) match {
       case (x: Bool, value: Bool) =>
--- a/src/main/scala/chiseltest/package.scala
+++ b/src/main/scala/chiseltest/package.scala
@@ -250,13 +250,13 @@ package object chiseltest {
       }
     }
 
-    def peek():          T = peekWithStale(false)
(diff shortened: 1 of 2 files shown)


ekiwi:
This is perfect timing! I am also playing around with allowing simple types to be poked directly without conversion.
https://github.com/ucb-bar/chisel-testers2/pull/453

carlosedp:
I found a weird case comparing the Booleans with `should be`.
Scalatest complains:

```
[info] compiling 1 Scala source to /Users/cdepaula/repos/chisel/chisel-testers2/target/scala-2.12/test-classes ...
[error] /Users/cdepaula/repos/chisel/chisel-testers2/src/test/scala/chiseltest/tests/LiteralTest.scala:59:23: overloaded method value should with alternatives:
[error]   (endWithWord: org.scalatest.matchers.dsl.EndWithWord)(implicit ev: Any <:< String)LiteralTest.this.ResultOfEndWithWordForString <and>
[error]   (startWithWord: org.scalatest.matchers.dsl.StartWithWord)(implicit ev: Any <:< String)LiteralTest.this.ResultOfStartWithWordForString <and>
[error]   (includeWord: org.scalatest.matchers.dsl.IncludeWord)(implicit ev: Any <:< String)LiteralTest.this.ResultOfIncludeWordForString <and>
[error]   (notExist: org.scalatest.matchers.dsl.ResultOfNotExist)(implicit existence: org.scalatest.enablers.Existence[Any])org.scalatest.Assertion <and>
[error]   (existWord: org.scalatest.matchers.dsl.ExistWord)(implicit existence: org.scalatest.enablers.Existence[Any])org.scalatest.Assertion <and>
[error]   (containWord: org.scalatest.matchers.dsl.ContainWord)org.scalatest.matchers.dsl.ResultOfContainWord[Any] <and>
[error]   (haveWord: org.scalatest.matchers.dsl.HaveWord)LiteralTest.this.ResultOfHaveWordForExtent[Any] <and>
[error]   (beWord: org.scalatest.matchers.dsl.BeWord)LiteralTest.this.ResultOfBeWordForAny[Any] <and>
[error]   (inv: org.scalactic.TripleEqualsSupport.TripleEqualsInvocationOnSpread[Any])(implicit ev: Numeric[Any])org.scalatest.Assertion <and>
[error]   [U](inv: org.scalactic.TripleEqualsSupport.TripleEqualsInvocation[U])(implicit constraint: org.scalactic.CanEqual[Any,U])org.scalatest.Assertion <and>
[error]   (notWord: org.scalatest.matchers.dsl.NotWord)org.scalatest.matchers.dsl.ResultOfNotWordForAny[Any] <and>
[error]   [TYPECLASS1[_], TYPECLASS2[_]](rightMatcherFactory2: org.scalatest.matchers.dsl.MatcherFactory2[Any,TYPECLASS1,TYPECLASS2])(implicit typeClass1: TYPECLASS1[Any], implicit typeClass2: TYPECLASS2[Any])org.scalatest.Assertion <and>
[error]   [TYPECLASS1[_]](rightMatcherFactory1: org.scalatest.matchers.dsl.MatcherFactory1[Any,TYPECLASS1])(implicit typeClass1: TYPECLASS1[Any])org.scalatest.Assertion <and>
[error]   (rightMatcherX1: org.scalatest.matchers.Matcher[Any])org.scalatest.Assertion
[error]  cannot be applied to (org.scalatest.matchers.Matcher[Boolean])
[error]       c.out.litPeek() should be (true)
``` 

I tried comparing the types, and the `litPeek()` method returns a java.lang.Boolean and the testcase has scala.Boolean. I tried forcing return of the same type but it also failed. Might need some help on this.

carlosedp:
This also addresses Martin's PR here: https://github.com/chipsalliance/chisel3/pull/2263

ekiwi:
> This also addresses Martin's PR here: [chipsalliance/chisel3#2263](https://github.com/chipsalliance/chisel3/pull/2263)

Yes. I think fixing this in chiseltest instead of chisel seems to be a good idea to me.

jackkoenig:
I think we should make literals in chisel3 work more as one might expect. In the meantime (and perhaps in addition to that), I think extending chiseltest in this way makes sense.

ekiwi:
> I think we should make literals in chisel3 work more as one might expect.

I have been promising people the magical "chisel constant propagation", but seems like it is too big of a project (especially from the testing perspective) for anyone of us to tackle in the near future.

carlosedp:
Agree. Keeping the language and it's libraries simple is essential for more traction. It's in the small details since Scala is already a quite "complex" language.
Trying to avoid the creation of complex patterns (like multiple conversion steps as `.peek().litValue` and so) helps bringing new users.

carlosedp:
Something is weird in matching the Any function with Boolean. I've opened an issue at scalatest: https://github.com/scalatest/scalatest/issues/2077
Maybe define how to handle `Interval` and `Record`types. Also check how to return `FixedPoint` since I've never used this type before.

ekiwi:
> Something is weird in matching a function returning `Any` with `Boolean`.

I think the way we probably want to try and work around this is to split up the `implicit class testableData[T <: Data](x: T) {` class. Currently it is a generic for all classes that extend Data. Instead, we could (or at least I hope this works ..) try to have one implicit class for every basic type, e.g., `implicit class testableUInt(x: UInt)` and then a generic only for Bundles which the users is actually supposed to inherit from: `implicit class testableBundle[T <: Bundle](x: B)`. That might mean that we will have to repeat some method names, however, I think it might be a lot easier in the end to write the individual methods.

However, even with that approach I am not sure if we will be able to return a Boolean instead of a BigInt. We might just have to rely on the user to do that conversion, e.g., `if(dut.io.a.litPeek() == 1) {...}`

carlosedp:
Actually it works fine the way I defined... it returns the Boolean fine... I believe the error I got might be on Scalatest side with the "should be" matcher.
Even the Vectors work returning a Seq of the enclosed Scala type :)
Separating it all might complicate things.

ekiwi:
> Actually it works fine the way I defined... it returns the Boolean fine...

I am just not sure if I feel comfortable with a method that returns `Any`

schoeberl:
Yes, agree that this should be put into ChiselTest. But, please do *not* have ```lit``` in the function name. That is super misleading and was my original reason why I wanted ```toBigInt```. I also think ```to``` is better than ```as```, because it converts from Chisel land/types *to* Scala/tester land/types.

carlosedp:
What about `.asNative()` ... we can discuss in the upcoming dev-meeting.

carlosedp:
I think this might be more suited with different functions to each type:
```
    def peekBool():       Boolean = peekWithStaleBool(false)
    def peekInt():        BigInt = peekWithStaleInt(false)
    def peekDecimal():    BigDecimal = peekWithStaleDec(false)
    def peekVectorInt():  Seq[BigInt] = peekWithStaleVecInt(false)
    def peekVectorBool(): Seq[Boolean] = peekWithStaleVecBool(false)
```

I just wished I could have a single `peekVector` that depending on the input (`Vec[Bool]` or `Vec[UInt/SInt]`) I had the return as `Seq[Boolean]` or `Seq[BigInt]`.

carlosedp:
Something wonky going on with "formal verification tests (macos-latest, btormc) "...

schoeberl:
why does ```peekInt``` return a ```BigInt``` and not a ```Int```? Do an additional ```peekBigInt```.

Also ```peekVector``` is misleading, as ```Vector``` is a regular Scala type. Use ```peekVec```.

carlosedp:
Thanks for the feedback @schoeberl. What about this:

```scala
    def peek():        T = peekWithStale(false)
    def peekBool():    Boolean = peekWithStaleBool(false)
    def peekBigInt():  BigInt = peekWithStaleBigInt(false)
    def peekInt():     Integer = peekWithStaleBigInt(false).toInt
    def peekDecimal(): BigDecimal = peekWithStaleDec(false)
    def peekVecInt():  Seq[BigInt] = peekWithStaleVecInt(false)
    def peekVecBool(): Seq[Boolean] = peekWithStaleVecBool(false)
```

I also added a couple more tests.

schoeberl:
I would try to stay consistent with the names, not mix Chisel and Scala types: peekBoolean, peekInteger, peekBigInt, peekBigDecimal. peekSeqBigInt... now it becomes a bit lengthy.

To not make the API to big, I would drop the Decimal.

carlosedp:
Better this way:

```scala
    def peek():           T = peekWithStale(false)
    def peekBoolean():    Boolean = peekWithStaleBool(false)
    def peekBigInt():     BigInt = peekWithStaleBigInt(false)
    def peekInteger():    Integer = peekWithStaleBigInt(false).toInt
    def peekBigDecimal(): BigDecimal = peekWithStaleDec(false)
    def peekVecBigInt():  Seq[BigInt] = peekWithStaleVecInt(false)
    def peekVecBoolean(): Seq[Boolean] = peekWithStaleVecBool(false)
```

ekiwi:
I don't think there is any value in distinguishing between `Int` and `BigInt`. They can mostly be used interchangeably and since we internally already use BigInt, casting to `Int` is not a performance improvement.

carlosedp:
> I don't think there is any value in distinguishing between `Int` and `BigInt`. They can mostly be used interchangeably and since we internally already use BigInt, casting to `Int` is not a performance improvement.

I agree, initially I've implemented it only on BigInt but casted to Int for consistency as pointed.

src/main/scala/chiseltest/package.scala
```diff
@@ -205,7 +205,58 @@ package object chiseltest {
...
+          elt.peekWithStaleBool(stale)
+        }
+        Seq(elementValueFns: _*)
+      }
+    }
+
+    def peek():           T = peekWithStale(false)
+    def peekBoolean():    Boolean = peekWithStaleBool(false)
+    def peekBigInt():     BigInt = peekWithStaleBigInt(false)
+    def peekInteger():    Integer = peekWithStaleBigInt(false).toInt
```
jackkoenig, line 256:
```suggestion
    def peekInt():    Int = peekWithStaleBigInt(false).toInt
```
I would stick with `Int` which is the Scala type, not `Integer` which is actually a Java type.

That being said, I agree with Kevin that I don't see value in `Int` over `BigInt`. In my experience, `Int` just leads to grief when accidentally overflowing so it might be best to just stick with BigInt. The implicit conversion from Int to BigInt makes comparisons and stuff work very intuitively.

carlosedp: Awesome, so can I keep the `peekInt` function but returning BigInt?
Also go back to `Bool` instead of `Boolean` for shortness?

carlosedp: Better now:

```scala
    def peek():        T = peekWithStale(false)
    def peekBool():    Boolean = peekWithStaleBool(false)
    def peekInt():     BigInt = peekWithStaleBigInt(false)
    def peekDec():     BigDecimal = peekWithStaleDec(false)
    def peekVecInt():  Seq[BigInt] = peekWithStaleVecInt(false)
    def peekVecBool(): Seq[Boolean] = peekWithStaleVecBool(false)
```

schoeberl: OK, my mistake on ```Integer```. And agree that ```BigInt``` is enough. But then I would still call it ```peekBigInt``. And also ```peekBoolean``` as ```Bool``` is a Chisel type. Furthermore, what does ```peekVec``` return? I guess not a ```Vec```., as that is a Chisel type. Maybe we do not really need the peeking of a Vec, as one can index in the tester and then just use the ```peekBigInt``` on that index. I do not see a big value on getting a ```Seq``` back.

carlosedp: News about this? Should I keep then as they are now?

schoeberl:
You are still mixing Chisel and Scala types in your function names. This is inconsistent. What about just:

```
peekInt()
peekBoolean()
```
and the ```peekInt``` returns a ```BitInt```, which is quite compatible with the Scala ```Int```. No need for decimal, one can always convert from ```BigInt``` when needed. No need for returning ```Seq```'s as one can always index into the ```Vec```. The fewer methods, the nicer the API. Just my opinion.

carlosedp:
There it goes... cleaned up some functions. Renamed Bool one.

schoeberl:
Nice. LGTM.

src/main/scala/chiseltest/package.scala
```diff
@@ -205,7 +205,23 @@ package object chiseltest {
...
+    }
+
+    protected def peekWithStaleBigInt(stale: Boolean): BigInt = x match {
+      case (x: UInt) => Context().backend.peekBits(x, stale)
+      case (x: SInt) => Context().backend.peekBits(x, stale)
+    }
+
+    def peek():        T = peekWithStale(false)
+    def peekBoolean(): Boolean = peekWithStaleBool(false)
+    def peekInt():     BigInt = peekWithStaleBigInt(false)
```
jackkoenig, line 224: It would be best if these weren't available on `Aggregates`, could these be extension methods on `T <: Element` instead of the current `T <: Data`?

carlosedp: Humm.. might need some guidance here :)

ekiwi:
I think: https://github.com/ucb-bar/chiseltest/pull/480 should address the issue to make sure that `peekBoolean` and `peekBigInt` are only available for the right chisel types.

ekiwi:
All functionality in this PR was included with https://github.com/ucb-bar/chiseltest/pull/480

PR #459, opened 2021-12-16
Fix Verilator on Mac

jackkoenig:
CI is failing for Verilator for Mac. There were recent changes to Verilator that make it no longer work with Mac/BSD ar (see https://github.com/verilator/verilator/issues/2999).

This fixes the issue by installing an older version of Verilator via Homebrew.

It's unfortunately a little slow because it builds from source. We badly need to fix CI, we can speed it up in a follow on PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Verilator
         run: |
-          brew install verilator
+          brew extract --version=4.200 verilator homebrew/cask
+          brew install verilator@4.200
           verilator --version
       - name: Setup Scala
         uses: olafurpg/setup-scala@v10


.github/workflows/test.yml
```diff
@@ -77,6 +77,11 @@ jobs:
           export PATH="/usr/local/opt/binutils/bin:$PATH"
```
ekiwi, line 77: try this instead: https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#adding-a-system-path

ekiwi: @jackkoenig

jackkoenig: Thanks!

.github/workflows/test.yml
```diff
@@ -8,182 +8,204 @@ on:
...
+          brew install binutils
+          # Use GNU binutils over built-in binutils
+          echo "ldflags=-L/usr/local/opt/binutils/lib" >> $GITHUB_ENV
+          echo "cppflags=-I/usr/local/opt/binutils/include" >> $GITHUB_ENV
+          echo "/usr/local/opt/binutils/bin:$PATH" >> $GITHUB_PATH
       - name: Install Verilator
         run: |
-          brew install verilator
+          brew install verilator --build-from-source
           verilator --version
```
jackkoenig, line 83: I wouldn't do it this way long term, but yeah we might need to start building verilator from source just like we do for Linux since it seems that you need to use the GNU ar to compile Verilator itself for it to work right...

Anyway just trying out `brew --build-from-source` but assuming this works I'll make this build and cache the same way that the verilator regression do below.

ekiwi: Is this going to be our recommendation for people who want to use chiseltest on Mac? How are you doing this on your own machine and would you mind adding some instructions to our Readme?

jackkoenig: > How are you doing this on your own machine

Honestly, I haven't been using Verilator on my on laptop lately because it's been broken due to the `ar` business. I need to figure out a reasonable way to get it working that I can encourage other people to emulate. It's pretty annoying that the version from package managers (both Homebrew and Macports) just don't actually work. 

Perhaps the real solution is to fix the packages such that they bring in GNU ar but I don't have much experience with that sort of thing...

ekiwi: Is there a way to force brew to pick an older version of Verilator? Are there any older versions that are not broken?

jackkoenig: I think there is, trying it out.

jackkoenig: Okay the solution is:

```
brew extract --version=4.200 verilator homebrew/cask
brew install verilator@4.200
```

This is much better, although ideally the newest version of the package would work. It does build from source though which is a pain so need to figure out how to cache it in Github Actions

jackkoenig:
This has been folded in to https://github.com/ucb-bar/chisel-testers2/pull/463 since there are now 2 separate CI breakages.

PR #467, opened 2021-12-23
Implements a feature that allows DUT's to specify a simulation timeout

chick:
Implements a rocket-chip compatibility feature that allows DUT's to specify a simulation timeout

Rocket-chip's technique of having the DUT carry the timeout simulation conflicts with the late
and very low level evaluation of the DUT generator function. This means that rocket-chip cannot
take advantage of the relatively new `runUntilStop` method. This PR provides a mechanism for
chiseltest to use it.
When the DUT is generated, return the DUT up the chain. If the DUT inherits from `DutSpecifiesTimeout`
then use the `timeout` specified there instead of the `timeout` paramter to `runUntilStop`
A test demonstrating this working is in DutSpecifiesTimeoutSpec

### DIFF
--- a/src/main/scala/chiseltest/internal/HardwareTesterBackend.scala
+++ b/src/main/scala/chiseltest/internal/HardwareTesterBackend.scala
@@ -4,7 +4,15 @@ import chisel3.Module
 import chisel3.testers.BasicTester
 import chiseltest.{defaults, ChiselAssertionError, StopException, TimeoutException}
 import chiseltest.coverage.{Coverage, TestCoverage}
-import chiseltest.simulator.{Compiler, DebugPrintWrapper, DutSpecifiesTimeout, Simulator, SimulatorContext, StepInterrupted, StepOk}
+import chiseltest.simulator.{
+  Compiler,
+  DebugPrintWrapper,
+  DutSpecifiesTimeout,
+  Simulator,
+  SimulatorContext,
+  StepInterrupted,
+  StepOk
+}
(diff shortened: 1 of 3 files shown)


chick:
I am open to other ways of doing this but this is a pretty small change that will allow rocket-chip tests to take advantage of new runUntilStop just by using the DutSpecifiesTimeout. Refactoring on the rocket-chip to allow it to use runUntilStop without this change will involve hundreds of edits in dozens of files.

On a smaller note I am not entirely sure the precedence is right, perhaps runUntilStop's timeout should override this new value. I would have done it that way but using that would have to be done when timeout is > 0 which is less general

ekiwi:
I had a look at the unit tests in rocket chip. Here is my suggestion on how to port them to `chiseltest` with any new changes. Please let me know if this will work for you or if I am missing anything:

I think it should be enough to just disable the timeout when you run the tests, i.e., just do `runUntilStop(timeout=0)`. The way that the tests in rocket chip currently implement the timeout is through a [hardware counter](https://github.com/chipsalliance/rocket-chip/blob/4df857ae7fc70fd4ac270dd18ab80b8c2a75547f/src/main/scala/unittest/UnitTest.scala#L33) and an [assertion](https://github.com/chipsalliance/rocket-chip/blob/4df857ae7fc70fd4ac270dd18ab80b8c2a75547f/src/main/scala/unittest/UnitTest.scala#L34), both of which should continue working with `chiseltest`.

PR #468, opened 2021-12-26
Support FSDB (Verdi) waveform dump

SihaoLiu:
I added support for dumping FSDB waveform used by Synopsys Verdi

Local test passed: `sbt "testOnly chiseltest.simulator.VcsWaveformCompliance"`
gcc version: `gcc version 8.5.0 20210514 (Red Hat 8.5.0-4) (GCC)`
vcs version: `VCS S-2021.09-SP1`
verdi version: `S-2021.09-SP1`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@
 .bloop/
 .metals/
 .vscode/
+
+# Synopsys Verdi
+novas_dump.log
+ucli.key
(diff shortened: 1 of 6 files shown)


sequencer:
Please add a test with `RequiresVcs` tag, this test will be ignored in CI, but we can test locally in our machine with VCS.

ekiwi (approved):
Looks great! Thanks you very much!

PR #470, opened 2021-12-28
Remove VCS warning

SihaoLiu:
Three VCS warnings are removed
1. Remove `-none` when no waveform annotation is given, `-none` is not a valid VCS option
```
Warning-[RT_UO] Unsupported option
  Unsupported option '-none' is ignored
```
2. Replace `debug_pp` per VCS suggested
```
Warning-[DEBUG_DEP] Option will be deprecated
  The option '-debug_pp' will be deprecated in a future release.  
  Please use '-debug_acc+pp+dmptf -debug_region+cell+encrypt' instead.
```
3. Reposition `dumpvars` per IEEE Standard
```
Warning-[TFX-DUMPVARCA] DumpVar called previously
  As $dumpvars was called in previous time step, ignoring this call.$dumpfile at time #1000
  Please refer to section 18.1.2 in the IEEE Verilog Standard 1364-2001 for details on $dumpvars.
```

Local UnitTest Passed: `sbt "testOnly chiseltest.simulator.VcsWaveformCompliance"`
VCS/DVE Version: `S-2021.09-SP1`
Verdi Version: `S-2021.09-SP1`
GCC Version: `gcc version 9.2.1 20191120 (Red Hat 9.2.1-2) (GCC)`

### DIFF
--- a/src/main/scala/chiseltest/simulator/VcsSimulator.scala
+++ b/src/main/scala/chiseltest/simulator/VcsSimulator.scala
@@ -169,20 +169,22 @@ private object VcsSimulator extends Simulator {
   )
 
   private def DefaultFlags(topName: String, cFlags: Seq[String], verdiFlags: Seq[String]) = List(
-    "-full64",
-    "-quiet",
-    "-sverilog",
-    "-timescale=1ns/1ps",
-    "-debug_acc+pp+dmptf",
-    "-debug_region+cell+encrypt",
+    "-full64",                    /* Default using 64-bit VCS*/
+    "-quiet",                     /* Quite mode*/
+    "-sverilog",                  /* Harness is written in SystemVerilog*/
+    "-timescale=1ns/1ps",         /* Clock Period*/
+    "-debug_acc+pp+dmptf",        /* pp: registers and variables, callbacks, driver, and assertion debug capability*/
+                                  /* dmptf: debug ports and internal nodes/memories of tasks/functions*/
+    "-debug_region+cell+encrypt", /* cell: debug both real cell modules and the ports of real cell modules*/
+                                  /* encrypt: debug fully-encrypted instances*/
     s"-Mdir=$topName.csrc",
-    "+v2k",
+    "+v2k",                       /* verilog 2000 standard*/
     "+vpi",
-    "+vcs+lic+wait",
-    "+vcs+initreg+random",
-    "+define+CLOCK_PERIOD=1",
-    "-P",
-    "vpi.tab",
+    "+vcs+lic+wait",              /* wait for VCS license if there isn't one*/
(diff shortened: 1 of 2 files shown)


src/main/scala/chiseltest/simulator/VcsSimulator.scala
```diff
@@ -173,7 +173,8 @@ private object VcsSimulator extends Simulator {
     "-quiet",
     "-sverilog",
     "-timescale=1ns/1ps",
-    "-debug_pp",
+    "-debug_acc+pp+dmptf",
+    "-debug_region+cell+encrypt",
```
ekiwi, line 177: Since you have access to the VCS manual, could you add a small comment explaining why these options are needed? I honestly just copied over the `-debug_pp` from our old code with no idea as to why it is needed.

ekiwi (changes requested):
LGTM.

Only one small suggestion about adding some documentation to the non-obvious VCS option flags.

src/main/scala/chiseltest/simulator/VcsSimulator.scala
SihaoLiu: Sure, for the original option `debug_pp`, here is the explaination from VCS User Guide
-debug_pp
> It enables read/write access and callbacks to design nets, memory callback, assertion debug, VCS DKI, and VPI routine usage.

For the `debug_acc` option: 
`+pp` means:
> This option enables write capability on registers and variables, callbacks for the entire design, driver capability, and assertion debug capability.

`+dmptf` means:
> This option enables dumping of ports and internal nodes/memories of tasks/functions.

For the `debug_region` option:
`+cell` means:
> Applies debug capabilities to both real cell modules and the ports of real cell modules.

`+encrypt` means:
> Applies debug capabilities to the fully-encrypted instances (modules, programs, packages, and interfaces).

I did not find debug_pp in latest VCS documentation, I believe these should be just equivalent options.

ekiwi: Do you know what happens if you just remove these options? Do tests run faster?
We never really access any internal registers or variables via VPI, only the top-level ports.

ekiwi (approved):
:shipit:

src/main/scala/chiseltest/simulator/VcsSimulator.scala
SihaoLiu: I don't think it is only related VPI access. It is also used for waveform dump. If I turned off this option, the VPD and FSDB waveform dump will crash. I added more comments to explain these option

ekiwi: Thanks! The PR will be auto merged once the tests pass.

PR #472, opened 2022-01-04
deprecate chisel3.testers package

ekiwi:
This will allow us to remove the package for good after the next release.


chick (approved):
lgtm

PR #480, opened 2022-01-10
Scala type peek poke

ekiwi:
- [x] clean up peek/poke/expect classes:
  - [x] remove unused parameter,
  - [x] use lazy eval for expect message (similar to scala assert)
  - [x] clean up IntelliJ warnings
- [x] provide scala type peek/poke/expect access for:
  - [x] Bool: Boolean, BigInt (as input only!)
  - [x] UInt: BigInt
  - [x] SInt: BigInt
  - [x] Interval: BigDecimal, Double
  - [x] FixedPoint: BigDecimal, Double
- [x] check input size (since we loosen the type restrictions, we are adding a runtime check to ensure values fit into type)
  - [x] Bool, SInt, UInt

Note: for `Interval` and `FixedPoint` I decided to just make sure that we get all the features from the old PeekPokeTester API.

### DIFF
--- a/src/main/scala/chiseltest/internal/BackendInterface.scala
+++ b/src/main/scala/chiseltest/internal/BackendInterface.scala
@@ -46,17 +46,15 @@ trait BackendInterface {
   def pokeBits(signal: Data, value: BigInt): Unit
 
   /** Returns the current value on a wire.
-    * If stale is true, returns the current combinational value (after previous pokes have taken effect).
-    * If stale is false, returns the value at the beginning of the current cycle.
+    * Returns the current combinational value (after previous pokes have taken effect).
     */
-  def peekBits(signal: Data, stale: Boolean): BigInt
+  def peekBits(signal: Data): BigInt
 
   def expectBits(
     signal:  Data,
     value:   BigInt,
-    message: Option[String],
-    decode:  Option[BigInt => String],
-    stale:   Boolean
+    message: Option[() => String],
+    decode:  Option[BigInt => String]
   ): Unit
 
   /** Sets the timeout of the clock: the number of cycles the clock can advance without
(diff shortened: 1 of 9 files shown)


ekiwi:
@carlosedp  and @schoeberl, I would appreciate your feedback. Would this PR solve your issues wrt. using native Scala types to interact with the DUT?

carlosedp:
Pretty amazing @ekiwi! I just think we could have the `peekBigInt` as `peekInt` like Jack suggested in my PR.

Other than that, the ability to use native and Chisel types for poke and peek/expect is great!
Thanks a lot.

ekiwi:
> I just think we could have the `peekBigInt` as `peekInt` like Jack suggested in my PR.

Changed it.

Thanks!

schoeberl:
+1 Very good! Looking forward to having this available. Sad that it didn't make it for 3.5.0

ekiwi:
> +1 Very good! Looking forward to having this available. Sad that it didn't make it for 3.5.0

We can try to backport it.

jackkoenig (approved):

jackkoenig:
We can have this in 3.5.1/0.5.1!

carlosedp:
Awesome! Thanks all :)

schoeberl:
I do not really understand what backporting means. I thought this means that it will show up in the future 0.4.X version. It will not change 0.5.0, right?

On 0.5.1 and book/teaching timing. I can add this into the book now with the master SNAPSHOT. However, my class starts on February 3rd, and then all code and examples should be in place. Will 0.5.1 be out until then?

ekiwi:
> I do not really understand what backporting means

In this case it will eventually be part of 0.5.1

> On 0.5.1 and book/teaching timing.

I would not count on that. @jackkoenig knows more though.

PR #485, opened 2022-01-12
Update master for 0.5.0 release

jackkoenig:
* Enable CI for 0.5.x branch
* Enable Mergify for 0.5.x branch

I went ahead and added the `mergify_config.yml` a la chisel3 and firrtl so that we can generate the .mergify.yml. I also have it backporting to 0.3.x and 0.5.x but will defer to @ekiwi if he wants more or less branches available for backporting.


ekiwi (approved):
:shipit:

PR #501, opened 2022-01-27
Add FAQ and help for `--full-stacktrace`

jackkoenig:


README.md
```diff
@@ -142,6 +142,33 @@ We currently support the following versions of the [verilator](https://www.verip
 - `v4.108`: [Fedora 34](https://src.fedoraproject.org/rpms/verilator)
 - `v4.202`
 
+## Frequently Asked Questions
+
+### How do I rerun with --full-stacktrace?
+
+Whereas Chisel accepts command-line arguments, chiseltest exposes the underlying annotation interface.
```
ekiwi, line 149: Could you add one sentence here explaining why one would need to use this option?
You are doing a good job explaining how to enable full stack trace printing, but I am not sure if most users would understand what the benefit is or in which situation they would want to use it.

ekiwi (changes requested):

README.md
jackkoenig: Being perfectly honest, I don't think you ever want it unless you're hitting a bug in Chisel itself. I can add that here but 🤷‍♂️

oharboe:
I've tried various remedies to get a full stacktrace without luck:

https://scastie.scala-lang.org/CsGtXgYgTxWTU2K2otoejA

```scala
import chisel3._
import chisel3.stage.PrintFullStackTraceAnnotation

class Example(offset:Int, width:Int) extends Module {
  val bits  = IO(Input(UInt(1024.W)))
  val replace  = IO(Input(UInt(width.W)))
  val out      = IO(Output(UInt()))

  out := bits(bits.getWidth - 1, offset + width) ## replace ## bits(offset - 1, 0)
}

emitVerilog(new Example(1024-24, 24))
//emitVerilog(new Example(1024-24, 24), args=Array("--full-stacktrace"))
//emitVerilog(new Example(1024-24, 24), annotations=Seq(PrintFullStackTraceAnnotation))
//println(getVerilogString(new Example(1024-24, 24)))
```

jackkoenig:
> I've tried various remedies to get a full stacktrace without luck:
> 
> https://scastie.scala-lang.org/CsGtXgYgTxWTU2K2otoejA
> 
> ```scala
> import chisel3._
> import chisel3.stage.PrintFullStackTraceAnnotation
> 
> class Example(offset:Int, width:Int) extends Module {
>   val bits  = IO(Input(UInt(1024.W)))
>   val replace  = IO(Input(UInt(width.W)))
>   val out      = IO(Output(UInt()))
> 
>   out := bits(bits.getWidth - 1, offset + width) ## replace ## bits(offset - 1, 0)
> }
> 
> emitVerilog(new Example(1024-24, 24))
> //emitVerilog(new Example(1024-24, 24), args=Array("--full-stacktrace"))
> //emitVerilog(new Example(1024-24, 24), annotations=Seq(PrintFullStackTraceAnnotation))
> //println(getVerilogString(new Example(1024-24, 24)))
> ```

There is actually an error in that Scastie, but I will grant it is not necessarily communicated the best. If you look in the console it has: 
```
[error] main.scala:11: Invalid bit range (1023,1024) in class Playground$Example$$anonfun$4
[error] There were 1 error(s) during hardware elaboration.
```
It would be nicer if that were somehow included in the stack trace but not how some errors in Chisel work. Any error that is recoverable (so that Chisel can report more than 1 error at a time) does not make it into the stack trace but instead is reported via a println. Scastie does weird things when it inserts stack traces into your source, but the [empty] stack trace has the message "Fatal errors during hardware elaboration. Look above for error list." which should be true when run in a regular command-line.

jackkoenig:
It would be better though to include this and other "recoverable" errors in the exception message though.

oharboe:
> > I've tried various remedies to get a full stacktrace without luck:
> > https://scastie.scala-lang.org/CsGtXgYgTxWTU2K2otoejA
> > ```scala
> > import chisel3._
> > import chisel3.stage.PrintFullStackTraceAnnotation
> > 
> > class Example(offset:Int, width:Int) extends Module {
> >   val bits  = IO(Input(UInt(1024.W)))
> >   val replace  = IO(Input(UInt(width.W)))
> >   val out      = IO(Output(UInt()))
> > 
> >   out := bits(bits.getWidth - 1, offset + width) ## replace ## bits(offset - 1, 0)
> > }
> > 
> > emitVerilog(new Example(1024-24, 24))
> > //emitVerilog(new Example(1024-24, 24), args=Array("--full-stacktrace"))
> > //emitVerilog(new Example(1024-24, 24), annotations=Seq(PrintFullStackTraceAnnotation))
> > //println(getVerilogString(new Example(1024-24, 24)))
> > ```
> 
> There is actually an error in that Scastie, but I will grant it is not necessarily communicated the best. If you look in the console it has:
> 
> ```
> [error] main.scala:11: Invalid bit range (1023,1024) in class Playground$Example$$anonfun$4
> [error] There were 1 error(s) during hardware elaboration.
> ```
> 
> It would be nicer if that were somehow included in the stack trace but not how some errors in Chisel work. Any error that is recoverable (so that Chisel can report more than 1 error at a time) does not make it into the stack trace but instead is reported via a println. Scastie does weird things when it inserts stack traces into your source, but the [empty] stack trace has the message "Fatal errors during hardware elaboration. Look above for error list." which should be true when run in a regular command-line.

Ah. Yes, there is an error. Which is necassary to demonstrate the problem of trackinh down the error . What I need to fix it in my usecase, is the stacktrace....

jackkoenig:
> Ah. Yes, there is an error. Which is necassary to demonstrate the problem of trackinh down the error . What I need to fix it in my usecase, is the stacktrace....

Point well taking and this feature has been requested before. I have a branch https://github.com/chipsalliance/chisel3/tree/improve-error-reporting adding `--throw-on-first-error` that I intend to PR tonight or tomorrow and squeeze in to 3.5.1 that will do just what you want here.

oharboe:
> > Ah. Yes, there is an error. Which is necassary to demonstrate the problem of trackinh down the error . What I need to fix it in my usecase, is the stacktrace....
> 
> Point well taking and this feature has been requested before. I have a branch https://github.com/chipsalliance/chisel3/tree/improve-error-reporting adding `--throw-on-first-error` that I intend to PR tonight or tomorrow and squeeze in to 3.5.1 that will do just what you want here. Feel free to try it out in the meantime.

Super!

jackkoenig:
> [chipsalliance/chisel3/tree/improve-error-reporting](https://github.com/chipsalliance/chisel3/tree/improve-error-reporting) adding `--throw-on-first-error` that I intend to PR tonight or tomorrow and squeeze in to 3.5.1 that will do just what you want here. Feel

PR here: https://github.com/chipsalliance/chisel3/pull/2376

ekiwi (approved):

PR #507, opened 2022-02-21
Support Verilator backend on Windows

howardlau1999:
According to [Spinal HDL documentation](https://spinalhdl.github.io/SpinalDoc-RTD/dev/SpinalHDL/Simulation/install/Verilator.html), Verilator can be used on Windows via MSYS2. But the executable on Windows is `verilator_bin` (see [here](https://github.com/SpinalHDL/SpinalHDL/blob/8631157de28920e6e50854a62f15aaebf52adc62/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L477)), we need to check the OS to get the correct name.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,6 +84,25 @@ jobs:
       - name: Test
         run: sbt "testOnly -- -n RequiresVerilator"
 
+  verilator-windows:
+    name: verilator regression on windows
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Verilator
+        run: C:\msys64\usr\bin\bash.exe -l -c "pacman -Sy --noconfirm --needed base-devel mingw-w64-x86_64-toolchain git flex mingw-w64-x86_64-cmake mingw-w64-x86_64-verilator"
+      - run: echo 'C:\msys64\usr\bin' | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - run: echo 'C:\msys64\mingw64\bin' | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      - name: Print Verilator Version
+        run: verilator_bin --version
+      - name: Setup Scala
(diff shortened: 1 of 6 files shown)


howardlau1999:
@jackkoenig PTAL

ekiwi (changes requested):
Thanks! This looks like a good change.

Could you please run `sbt "testOnly -- -n RequiresVerilator"` on your windows machine and confirm that all tests pass?

Then I will be happy to merge.

howardlau1999:
> Thanks! This looks like a good change.
> 
> Could you please run `sbt "testOnly -- -n RequiresVerilator"` on your windows machine and confirm that all tests pass?
> 
> Then I will be happy to merge.

I have added a github action job for windows in the workflow. But it seems the tests were skipped...

ekiwi:
> I have added a github action job for windows in the workflow. But it seems the tests were skipped...

Yeah, not sure what is going on. I would recommend to try and debug on your local Windows machine.

howardlau1999:
> > I have added a github action job for windows in the workflow. But it seems the tests were skipped...
> 
> Yeah, not sure what is going on. I would recommend to try and debug on your local Windows machine.

Some tests failed. I guess that's because their names are too long that leads to error on windows, what should I do?

ekiwi:
> I guess that's because their names are too long that leads to error on windows, what should I do?

Not sure. Sorry, I do not know a lot about Windows.

If your changes improve things for you and since they do not break anything, you could just remove the CI changes and I can merge your changes without tests.

Otherwise you would have to find a way to configure Windows to accept longer filenames.

howardlau1999:
> > I guess that's because their names are too long that leads to error on windows, what should I do?
> 
> Not sure. Sorry, I do not know a lot about Windows.
> 
> If your changes improve things for you and since they do not break anything, you could just remove the CI changes and I can merge your changes without tests.
> 
> Otherwise you would have to find a way to configure Windows to accept longer filenames.

Well, I choose to remove the CI changes and not to fight with windows

ekiwi (approved):
Thanks!

Issue #398, opened 2021-09-14
make simulation ticks available to the user


Z-KN:
When I am building a testbench, I want to get the current simulation time after some step function, so what is related function in chisel-tester2? In other words, what is the counterpart in chisel-tester2 of $time, a built-in function, in verilog?

ekiwi:
Currently we do not supply an interface for this. That could be added, but since `chiseltest` has no notion of real time, the simulation time that we would be able to provide is essentially the number of lock ticks that have elapsed. Would that work for your use-case? You can always multiply the step count with your clock period to get the real time elapsed.

schoeberl:
Access to the clock tick counter would make sense for easy handling of timeouts.

ekiwi:
> Access to the clock tick counter would make sense for easy handling of timeouts.

Timeouts are already supported through the [setTimeout](https://github.com/ucb-bar/chisel-testers2/blob/a1afe0516a49f296cafb5aeab68b87413e96c2ef/src/test/scala/chiseltest/tests/TimeoutTest.scala) function.

schoeberl:
> On 23 Sep, 2021, at 17:37, Kevin Laeufer ***@***.***> wrote:
> 
> 
> Access to the clock tick counter would make sense for easy handling of timeouts.
> 
> Timeouts are already supported through the setTimeout <https://github.com/ucb-bar/chisel-testers2/blob/a1afe0516a49f296cafb5aeab68b87413e96c2ef/src/test/scala/chiseltest/tests/TimeoutTest.scala> function.
> 
> 
OK, fine. But I meant your test bench might like to do something more advanced on a timeout.

Martin

carlosedp:
You mean having something like `dut.clock.peek()` that returns the amount of elapsed ticks? I think it might be useful.

ekiwi:
> You mean having something like `dut.clock.peek()` that returns the amount of elapsed ticks? I think it might be useful.

Yes. Just that it would be called `getTicks` or something similar since `peek()` already returns whether the clock is high or low.

carlosedp:
Implemented in #510

PR #510, opened 2022-03-21
Add a getStep method for clock step "peeking"

### LINKED ISSUE #71: Access to a clock counter
For simulation, it would be convenient to have access to a clock cycle counter. Adding a counter into the design and then routing it to an output is not the best workaround.

carlosedp:
Added a (naive) solution to track the clock steps that is not dependent of the backend.

The solution could track multiple clock domains thru a mutable map of values but the test backend apparently doesn't support it (a test has been written but ignored).

Fixes #71 , #398


carlosedp:
Would you leave a feedback on this @schoeberl ?

schoeberl:
Nice feature. Maybe we should discuss the name of the function. ```getSteps``` as plural, or better ```getCnt``` as at which count we are.

ekiwi:
> `getSteps` as plural, or better `getCnt` as at which count we are.

I like `getSteps` since the function to advance the clock is called `step`. Another alternative is `getStepCount`.

ekiwi (commented):
Thanks for pushing this forward @carlosedp. I might try to change the implementation a bit, but the unit test looks good!

Let's see what name we want to settle on.

carlosedp:
We can discuss on today's dev meeting. Once settled-in, I send an update to the PR.

carlosedp:
Additionally, I'd love to have the multi-clock test case working. Is there anything we can do on some backend for this @ekiwi?

ekiwi:
> Additionally, I'd love to have the multi-clock test case working. Is there anything we can do on some backend for this @ekiwi?

Not without a lot of work. Currently multi clock just isn't supported.

ekiwi (approved):
LGTM, we just need to decide on whether `getStep` should include the implicit reset cycle or not.

carlosedp:
Thanks for the reimplementation Kevin! I understand that without any `clock.step()` calls, we are at step zero but since the reset is implicit (and happens on first clock) it advances one clock.

I'd vote for user cycles, starting at 0 since the user does not have control over this initial reset.

carlosedp:
I also prefer `getStep` instead of plural `getSteps` as I read it as "We are at step X right now" but it's personal preference :)

PR #518, opened 2022-04-17
[GenericBackend] Add `UnpeekableException` if signal not found

michael-etzkorn:
For Modules with internal signals when the `GenericBackend` goes to peek, the signal will not be found.  This results in a slightly unintuitive `java.util.NoSuchElementException`.

```scala
class dummy extends Module {
  val unseen = Reg(UInt(6.W))
  unseen := 42.U
}
```
The following test on the above Module throws java.util.NoSuchElementException: key not found exception
```scala
     test(new dummy) { c =>
        c.unseen.expect(42.U)
     }

```

The proposal is to add an `UnpeekableException` which suggests the user use `chiseltest.experimental.expose`. 

The exception message now reads:
```sh
[info]  chiseltest.UnpeekableException: Signal not found. Perhaps you're peeking a non-IO signal. 
[info]  If so, consider using the chiseltest.experimental.expose API.
```

---
As an aside:
- The repetitive code makes me think `BackendExecutive`, `package` or a parent class should handle checking for this as it's possible other Backends would still have this issue.

### DIFF
--- a/src/main/scala/chiseltest/exceptions.scala
+++ b/src/main/scala/chiseltest/exceptions.scala
@@ -5,6 +5,7 @@ package chiseltest
 class NotLiteralException(message: String) extends Exception(message)
 class LiteralTypeException(message: String) extends Exception(message)
 class UnpokeableException(message: String) extends Exception(message)
+class UnpeekableException(message: String) extends Exception(message)
 class UnsupportedOperationException(message: String) extends Exception(message)
 
 class ClockResolutionException(message: String) extends Exception(message)
--- a/src/main/scala/chiseltest/internal/GenericBackend.scala
+++ b/src/main/scala/chiseltest/internal/GenericBackend.scala
@@ -2,7 +2,14 @@
 
 package chiseltest.internal
 
-import chiseltest.{ChiselAssertionError, ClockResolutionException, Region, StopException, TimeoutException}
+import chiseltest.{
+  ChiselAssertionError,
+  ClockResolutionException,
+  Region,
+  StopException,
+  TimeoutException,
+  UnpeekableException
+}
 import chisel3._
 import chiseltest.coverage.TestCoverage
 import chiseltest.simulator.{SimulatorContext, StepInterrupted, StepOk}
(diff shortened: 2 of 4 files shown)


ekiwi (changes requested):
This looks great. Could you add a unittest please? Just some circuit where you peek and poke internal signals and you add a `assertThrows` to make sure the correct exception is thrown?

Thanks!

michael-etzkorn:
I'll add this as soon as I get a chance and request a review again at that time!

michael-etzkorn:
Added a check in `BasicTest` for this new exception. Output looks good for Treadle 1.5.2, but as I mentioned on gitter, 1.6-SNAPSHOT for Treadle seems to have created some issue with chiseltest's main. 
 
```sbt
[info] BasicTest:
[info] Testers2
[info] - should test static circuits
[info] - should fail on poking outputs
[info] - should fail on peeking internal signals
[info] - should fail on expect mismatch
[info] - should test record partial poke
[info] - should fail on record expect mismatch
[info] - should fail on partial expect mismatch
[info] - should fail with user-defined message
[info] - should test inputless sequential circuits
[info] - should test combinational circuits
[info] - should test sequential circuits
[info] - should test reset
[info] Run completed in 5 seconds, 340 milliseconds.
[info] Total number of tests run: 12
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 12, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

michael-etzkorn:
```diff
val defaultVersions = Map(
  "chisel3" -> "3.6-SNAPSHOT"
---  "treadle" -> "1.6-SNAPSHOT"
+++  "treadle" -> "1.5.2"
)
```

Following change results in CI passing, but I presume we'll have to fix the CI issue before merge. 

I'll re-request a review now anyway since the PR changes seem ok (main with recent snapshot downloads seems to be broken now too).

michael-etzkorn:
@ekiwi Should I repush to have this pass CI now that https://github.com/chipsalliance/treadle/issues/407 is fixed? The changes I made in `BasicTest` now pass.

ekiwi (approved):
Thanks! :shipit:

PR #526, opened 2022-04-27
formal: add MagicPacketTracker VIP

ekiwi:
This re-usable VIP has proven to be pretty useful for verifying various FIFOs.

### DIFF
--- a/src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
+++ b/src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
@@ -1,0 +1,109 @@
+package chiseltest.formal.vips
+
+import chisel3._
+import chisel3.experimental.IO
+import chisel3.util._
+
+/** Tracks random packets for formally verifying FIFOs
+  *
+  *  This ensures that when some data enters the FIFO, it will always be dequeued after the correct number of elements.
+  *  So essentially we are verifying data integrity. Note that this does not imply that the FIFO has no bugs
+  *  since e.g., a FIFO that never allows elements to be enqueued would easily pass our assertions.
+  *
+  *  This module was inspired by the MagicPacketTracker used in the evaluation of the following paper:
+  *  Mann, Makai, and Clark Barrett. "Partial order reduction for deep bug finding in synchronous hardware.", CAV'20.
+  */
+object MagicPacketTracker {
+  def apply[D <: Data](enq: ValidIO[D], deq: ValidIO[D], depth: Int, debugPrint: Boolean): Unit = {
+    val tracker = Module(new MagicPacketTracker(chiselTypeOf(enq.bits), depth, debugPrint))
+    tracker.enq := enq
+    tracker.deq := deq
+    val startTracking = IO(Input(Bool()))
+    tracker.startTracking := startTracking
+  }
+  def apply[D <: Data](enq: ValidIO[D], deq: ValidIO[D], depth: Int): Unit =
+    apply(enq, deq, depth, debugPrint = false)
+  def apply[D <: Data](enq: DecoupledIO[D], deq: DecoupledIO[D], depth: Int, debugPrint: Boolean): Unit =
+    apply(asValid(enq), asValid(deq), depth, debugPrint)
+  def apply[D <: Data](enq: DecoupledIO[D], deq: DecoupledIO[D], depth: Int): Unit =
+    apply(asValid(enq), asValid(deq), depth)
+
+  private def asValid[D <: Data](port: DecoupledIO[D]): ValidIO[D] = {
+    val validIO = Wire(ValidIO(chiselTypeOf(port.bits)))
+    validIO.valid := port.fire
+    validIO.bits := port.bits
+    validIO
+  }
+}
+
+class MagicPacketTracker[D <: Data] private (dataTpe: D, fifoDepth: Int, debugPrint: Boolean) extends Module {
+  require(fifoDepth > 0, "Fifo depth needs to be positive!")
+  val enq = IO(Input(ValidIO(dataTpe)))
+  val deq = IO(Input(ValidIO(dataTpe)))
+
+  // count the number of elements in the fifo
+  val elementCount = RegInit(0.U(log2Ceil(fifoDepth + 1).W))
+  val nextElementCount = Mux(
+    enq.fire && !deq.fire,
+    elementCount + 1.U,
+    Mux(!enq.fire && deq.fire, elementCount - 1.U, elementCount)
+  )
+  elementCount := nextElementCount
+
+  // track a random "magic" packet through the fifo
+  val startTracking = IO(Input(Bool()))
+  val isActive = RegInit(false.B)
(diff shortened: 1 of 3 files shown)


src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
```diff
@@ -0,0 +1,109 @@
...
+  // track a random "magic" packet through the fifo
+  val startTracking = IO(Input(Bool()))
+  val isActive = RegInit(false.B)
+  val packetValue = Reg(chiselTypeOf(enq.bits))
+  val packetCount = Reg(chiselTypeOf(elementCount))
+
+  when(!isActive && enq.fire && startTracking) {
+    when(deq.fire && elementCount === 0.U) {
+      assert(
+        enq.bits.asUInt === deq.bits.asUInt,
```
vighneshiyer, line 62: Is this the default behavior of the Chisel Queue? I thought it is fully pipelined. Maybe there should be a flag similar to 'flow' to emit/exclude this assertion.

src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
```diff
@@ -0,0 +1,109 @@
...
+        "element should pass through the fifo, but %x != %x",
+        enq.bits.asUInt,
+        deq.bits.asUInt
+      )
+    }.otherwise {
+      isActive := true.B
+      packetValue := enq.bits
+      packetCount := nextElementCount
```
vighneshiyer, line 70: nextElementCount in this context is enq.fire && !deq.fire so is it OK to store elementCount directly instead? (and change the check below to perform the assertion on packetCount === 0.U)?

src/test/scala/chiseltest/formal/MagicPacketTrackerTests.scala
```diff
@@ -0,0 +1,383 @@
...
+  it should "verify some versions of the Queue from the Chisel standard library" taggedAs FormalTag in {
+    assume(DefaultBackend != CVC4EngineAnnotation, "CVC4 is too slow with some of these tests (5h+)")
+    verify(new ChiselQueueTest(DefaultDepth, 32, false, false), defaultOptions)
+    verify(new ChiselQueueTest(DefaultDepth, 32, false, true), defaultOptions)
+    verify(new ChiselQueueTest(DefaultDepth, 32, true, false), defaultOptions)
+    verify(new ChiselQueueTest(DefaultDepth, 32, true, true), defaultOptions)
+  }
+}
+
+private class PacketTrackerCounterOverflowTest extends Module {
```
vighneshiyer, line 98: What is this the point of this test? Is it to make sure the overflow assertion fires since that shouldn't happen in the other tests?

src/test/scala/chiseltest/formal/MagicPacketTrackerTests.scala
```diff
@@ -0,0 +1,383 @@
...
+
+
+// very similar to the QueueSpec in the chisel3 tests, but a formal test
+private class ChiselQueueTest(queueDepth: Int, bitWidth: Int, useSyncReadMem: Boolean, hasFlush: Boolean) extends Module {
+  val dut = Module(new Queue(UInt(bitWidth.W), queueDepth, useSyncReadMem = useSyncReadMem, hasFlush = hasFlush))
+  val io = IO(chiselTypeOf(dut.io)) ; io <> dut.io
+  // we need to tie the flush pin to false, because the MagicPacketTracker cannot (currently)
+  // handle flushes, it will complain about a missing packet if a flush happens
+  dut.io.flush.foreach(_ := false.B)
+  MagicPacketTracker(enq = dut.io.enq, deq = dut.io.deq, depth = queueDepth)
```
vighneshiyer, line 114: I see that the MagicPacketTracker apply function creates an Input for startTracking in the context in which its called, but this is unlike other Chisel modules whose apply functions gives you a handle to the instantiated module and you have to manually wire its IO upwards... I like your API better though since it avoids that wiring boilerplate.

src/test/scala/chiseltest/formal/MagicPacketTrackerTests.scala
```diff
@@ -0,0 +1,383 @@
...
+import chisel3._
+import chisel3.util._
+import chiseltest._
+import firrtl.AnnotationSeq
+import org.scalatest.flatspec.AnyFlatSpec
+
+class MagicPacketTrackerTests extends AnyFlatSpec with ChiselScalatestTester with Formal with FormalBackendOption {
+  behavior of "MagicPacketTracker"
+
+  private def defaultOptions: AnnotationSeq = Seq(BoundedCheck(7), DefaultBackend)
```
vighneshiyer, line 12: How exactly does this work? What is DefaultBackend and how is it materialized when the test closures are run?

src/test/scala/chiseltest/formal/MagicPacketTrackerTests.scala
```diff
@@ -0,0 +1,383 @@
...
+    assume(DefaultBackend != CVC4EngineAnnotation, "CVC4 is too slow with some of these tests (5h+)")
+    verify(new QueueFormalTest(new MyQueueV4(DefaultDepthPow2,32)), defaultOptions)
+  }
+
+  it should "verify QueueV5" taggedAs FormalTag in {
+    assume(DefaultBackend != CVC4EngineAnnotation, "CVC4 is too slow with some of these tests (5h+)")
+    verify(new QueueFormalTest(new MyQueueV5(DefaultDepthPow2,32)), defaultOptions)
+  }
+
+  it should "find bug in QueueV6 w/ pipe = false" taggedAs FormalTag in {
```
vighneshiyer, line 63: This is really cool

vighneshiyer (commented):
Just a few nits and questions, but this looks great!

src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
ekiwi: If the queue signals that an element is available to be dequeued, only then will that assertion take effect. So if the queue does not allow pass-through, then it has to signal empty for that cycle and thus this assertion will never fire.

src/test/scala/chiseltest/formal/MagicPacketTrackerTests.scala
ekiwi: Yes. Essentially what happens here is that the depth parameter of the tracker is set incorrectly which means that the internal counter cannot track all the elements that the queue can safe. In all other tests, the depth of the tracker is set correctly to equal the depth of the queue.

src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
ekiwi: We could. Not sure if that would make it easier to understand though.

src/main/scala/chiseltest/formal/vips/MagicPacketTracker.scala
vighneshiyer: Yes you're right - it looks good

PR #529, opened 2022-05-10
Use circt with verilator backend

chick:
The overall goal is to be able to run chiseltest unit tests with `CIRCT`/`firtool` *instead* of `firrtl`

## This is first step. It does the following
- Creates a new `Backend`, `VerilatorCirctBackend` that uses `firtool` to lower unit tests into system verilog to be run by `verilator`
- Adds some tests that use the `VerilatorCirctBackend`
- This meets a minimum test of `CIRCT`/`firrtool` in that it checks that the system verilog generated can be used successfully on existing unit tests

## TODO:
- consider adding more flags to the `firtool` call 
- See: [firtool example uses](https://github.com/sifive/federation/blob/master/mk/common/circt.mk#L5)
- Consider primarily `--dedup`

## Missing:
- chiseltest uses requires the **Scala** `firrtl` compiler to generate the following things that `firtool` does not do
  - Build a map between DUT top level IOs and the system verilog string names 
  - Generates a list of Combinational Paths in Annotations that are used to check peek poke sequences on those ports
  - Generates a list of coverage annotations
- decide what to do in each of the above cases.
  - currently `firrtl` compiler is called to generated necessary data in addition to the `firtool`


src/test/scala/chiseltest/backends/verilator/VerilatorCirctBasicTests.scala
```diff
@@ -0,0 +1,140 @@
...
+import chisel3._
+import chiseltest._
+import chiseltest.simulator.RequiresVerilator
+import chiseltest.tests.{PassthroughModule, StaticModule}
+import firrtl.AnnotationSeq
+import org.scalatest._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class VerilatorCirctBasicTests extends AnyFlatSpec with ChiselScalatestTester with Matchers {
```
ekiwi, line 12: There should be no need to duplicate the tests here. In general, the compliance tests in simulator should be enough.

If you really want to reuse these tests, I would suggest to make an abstract base class that takes the backend annotation as a constructor argument. Then you can derive both the original Verilator and the VerilatorCirct tests from the same base class. The goal here would be to try and reduce duplicate code.

src/main/scala/chiseltest/simulator/VerilatorCirctSimulator.scala
```diff
@@ -0,0 +1,120 @@
...
+import firrtl._
+import firrtl.annotations._
+import chiseltest.simulator.jna._
+import firrtl.stage.OutputFileAnnotation
+
+case object VerilatorCirctBackendAnnotation extends SimulatorAnnotation {
+  override def getSimulator: Simulator = VerilatorCirctSimulator
+}
+
+private object VerilatorCirctSimulator extends VerilatorSimulatorTrait {
```
ekiwi, line 15: I wonder if it would be easier to add a flag to the original `VerilatorSimulator` that switches the way that Verilog is generated.

src/main/scala/chiseltest/internal/Utils.scala
```diff
@@ -0,0 +1,21 @@
+package chiseltest.internal
+
+import os.Path
+
+private[chiseltest] object Utils {
```
ekiwi, line 5: Since this seems to only be used in one place, I would prefer if you could just define it as a private function in whatever class is using it.

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
```diff
@@ -0,0 +1,51 @@
...
+/** To disable tests that require the commercial VCS simulator use the following: `sbt testOnly -- -l RequiresVerilator` */
+object RequiresVerilatorCirct extends Tag("RequiresVerilatorCirct")
+
+class VerilatorCirctBasicCompliance extends BasicCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctStepCompliance extends StepCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctPeekPokeCompliance extends PeekPokeCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctWaveformCompliance extends WaveformCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctCoverageCompliance extends CoverageCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctMemoryCompliance extends MemoryCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
+class VerilatorCirctStopAssertAssumeCompliance extends StopAssertAssumeCompliance(VerilatorCirctSimulator, RequiresVerilatorCirct)
```
ekiwi, line 19: This looks good. However, you will also have to add a task to the CI to actually run these tests.

ekiwi (changes requested):
Looks good. In general there is some extra code that could probably be reduced to make for a smaller change since essentially all that should change is the way that firrtl is turned into Verilog.

For the CI, you need to define a task that runs all the Circt Verilator specific tests.

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
```diff
@@ -0,0 +1,51 @@
...
+
+package chiseltest.simulator
+
+import chiseltest.simulator.jna.JNAUtils
+import chiseltest.utils.CaptureStdout
+import org.scalatest.Tag
+import chiseltest.utils.FlatSpecWithTargetDir
+
+/** To disable tests that require the commercial VCS simulator use the following: `sbt testOnly -- -l RequiresVerilator` */
+object RequiresVerilatorCirct extends Tag("RequiresVerilatorCirct")
```
ekiwi, line 11: Instead of a `RequiresVerilatorCirct` tag, I would prefer if you could make a `RequiresFirTool` tag and then tag all tests as `RequiresVerilator` and `RequiresFirTool` since we might later start using the `firtool` for other Verilog simulators as well.

src/main/scala/chiseltest/internal/Utils.scala
chick: placed function where used. I'd like to note this code is super useful if you are trying to see or test a giant ugly call to verilator or some other program

src/main/scala/chiseltest/simulator/VerilatorCirctSimulator.scala
chick: I actually started that way, but I think this is better to keep this more separate. I think I will probably be doing more work in this area. This keeps it clean and distinct

src/test/scala/chiseltest/backends/verilator/VerilatorCirctBasicTests.scala
chick: Good point, will delete this file

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
chick: @ekiwi I'm not sure which is the right way to do this. Would you add an extra argument to the whole `(BasicCompliance, StepCompliance, ...., StopAssertAssumeCompliance` family, in order to pass in multiple tabs?

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
chick: I'm not prepared to add these to CI at this point, installation of firtool is still being refined.

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
ekiwi: we can probably overload the constructor to accept a Seq of Tags

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
ekiwi: OK. Just note that the merge of the PR will be blocked until it has working CI. I am happy to review the draft of course.

src/main/scala/chiseltest/simulator/VerilatorCirctSimulator.scala
ekiwi: So what is the plan for the other Verilog Simulators like icarus and VCS? Should the also have an option to use `firtool` instead of the firrtl compiler?

sequencer:
I have been thinking this:
> Build a map between DUT top level IOs and the system verilog string names

The future question might can be: how to map arbitrary signals from Chisel to System Verilog instance hierarchy? 

Is possible to use the chisel3 [trace api](https://github.com/chipsalliance/chisel3/pull/2077) for IO name generation?
adding a NLA Annotation to MFC?
I think this is useful to trace io/internal signals.

src/main/scala/chiseltest/simulator/VerilatorSimulator.scala
```diff
@@ -234,7 +234,9 @@ private trait VerilatorSimulatorTrait extends Simulator {
     if (verbose) {
       // print the command and pipe the output to stdout
       println(cmd.mkString(" "))
-      println(makeScriptFromCommand(cmd, Some(cwd / s"chiseltest_command_${cmd.head}.sh")))
+      if (cwd != null) {
+        println(makeScriptFromCommand(cmd, Some(cwd / s"chiseltest_command_${cmd.head}.sh")))
```
ekiwi, line 238: For you purposes, `cwd == null` essentially means `cwd == os.pwd`. So if you replace the `null` with `os.pwd` things should continue to work and you won't have to add the check here.

src/test/scala/chiseltest/simulator/VerilatorCirct.scala
chick: @ekiwi I've looked at this a little bit and it seems do-able (i think) but difficult enough that I'd prefer to do this work under a different PR, the CIRCT stuff is undergoing a lot of work and this PR will enable me to use chiseltest+circt in more situations and discover

ekiwi:
This PR has gone stale. Feel free to re-open when you want to continue working on this.

PR #531, opened 2022-06-23
added +plusarg support for Icarus

lsteveol:
Leveraged the existing work for the Verilator `+plusarg` support to also extend to Icarus Verilog. I don't currently have a VCS license on my personal machine, so I don't have a way to validate VCS. Will see if I can add the same to VCS.

Added an Icarus Verilog test which will test the `+plusarg` functionality. This is also a clone of the `VerilogBlackBoxTest` that is used for Verilator.


lsteveol:
Thanks @ekiwi , ran `scalafmtAll` and updated

PR #533, opened 2022-06-30
changed RawTester.test to accept a testname

lsteveol:
`RawTester.test` can now accept a `testName` instead of defaulting to one based on the millisecond time stamp. This is similar to how `RawTester.verify` works.


ekiwi (changes requested):
This looks good to me, but is missing a unittest which checks that the correct directory is created. Happy to merge, once you add that test. Thanks!

ekiwi (approved):
Thanks for adding the test! I made it work with treadle which is the easiest default. Icarus does not seem to be required for us to test this functionality.

PR #539, opened 2022-07-08
Fix verilator time

jackkoenig:
Fixes https://github.com/ucb-bar/chiseltest/issues/538

h/t @terpstra for the bug report and fix

I'm not sure how this happens but you can see my comment in the test about some really bizarre behavior that made writing this test take forever. Basically I originally had:
```scala
  behavior of "Verilator backend"
```
When running the test it would correctly generate Verilog and invoke Verilator to create a simulation binary, but `c.out` always returned `0` no matter what I did and any assertions or printfs inside the design were ignored. Merely changing the String "Verilator backend" to anything else fixed the issue.

I feel like I'm taking crazy pills but maybe there's some crazy interaction going on in the ScalaTest harness.

### DIFF
--- a/src/test/scala/chiseltest/backends/verilator/VerilatorTimeTaskTest.scala
+++ b/src/test/scala/chiseltest/backends/verilator/VerilatorTimeTaskTest.scala
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+package chiseltest.backends.verilator
+
+import chisel3._
+import chisel3.util._
+import chiseltest._
+import chiseltest.simulator.RequiresVerilator
+import org.scalatest.flatspec.AnyFlatSpec
+
+class VerilatorTimeTaskTest extends AnyFlatSpec with ChiselScalatestTester {
+  behavior of "Verilator backend"
+
+  val annos = Seq(VerilatorBackendAnnotation)
+
+  it should "support BlackBoxes that invoke the Verilog $time task" taggedAs RequiresVerilator in {
+    class BlackBoxTime extends BlackBox with HasBlackBoxInline {
+      val io = IO(new Bundle {
+        val clock = Input(Clock())
+        val out = Output(UInt(64.W))
+      })
+      setInline(s"${this.name}.v",
+      s"""module ${this.name}(
+         |  input clock,
+         |  output reg [63:0] out
+         |);
+         |  always @(posedge clock) begin
+         |    out <= $$time;
+         |  end
+         |endmodule
+         |""".stripMargin)
+    }
+
+    test(new Module {
+      val out = IO(Output(UInt(64.W)))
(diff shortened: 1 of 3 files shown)


jackkoenig:
Alright so this fix does not work on order versions of Verilator...

terpstra:
The other option is to make `real_time` a global variable so you can access it from the top-level function.

michael-etzkorn:
Maybe some weird interaction with `VerilogBlackBoxTests` since that has 'behavior of "Verilator Backend"'? Have you tried moving the code there?  It might make sense to move this test there anyhow.

https://github.com/ucb-bar/chiseltest/blob/638909a00a80a94ee8a0f11c9dfeea48138db805/src/test/scala/chiseltest/backends/verilator/VerilogBlackBoxTest.scala#L62


I'm very curious how this sort of bug occurs.

jackkoenig:
@michael-etzkorn 

> Update: I was unable to reproduce using just behavior of "Verilator backend".

I too am unable to reproduce it now. @davidbiancolin can attest to the fact that I am not crazy and this was happening. I think it must be some weird incremental compile/caching bug somewhere. I did do `sbt clean` and nuke the `test_run_dir` directory when I was having this issue, but I never nuked `target/` and it's also possible the JVM does some sort of caching of JNA stuff.

TLDR probably nothing to be concerned about but it was very weird

michael-etzkorn:
> 
> 
>I think it must be some weird incremental compile/caching bug somewhere. I did do `sbt clean` and nuke the `test_run_dir` directory when I was having this issue, but I never nuked `target/` and it's also possible the JVM does some sort of caching of JNA stuff.
> 
> 
> 
> TLDR probably nothing to be concerned about but it was very weird

I 100% trust you. Very fascinating. Sounds like one of those horror-story bugs, but glad the original line of source code isn't likely to be the true culprit.

src/test/scala/chiseltest/backends/verilator/VerilatorTimeTaskTest.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import chisel3._
+import chisel3.util._
+import chiseltest._
+import chiseltest.simulator.RequiresVerilator
+import org.scalatest.flatspec.AnyFlatSpec
+
+class VerilatorTimeTaskTest extends AnyFlatSpec with ChiselScalatestTester {
+  // Note that this does NOT work for some reason if the string is just "Verilator backend"
+  behavior of "The Verilator backend"
```
michael-etzkorn, line 12:
```suggestion
  behavior of "Verilator backend"
```

The original seems fine -- likely a locally isolated bug

jackkoenig:
@ekiwi I think this is done, it also includes [the same] test for `$time` in Icarus Verilog.

ekiwi (approved):
LGTM :shipit:

PR #542, opened 2022-07-22
ChiselEnum: implement peek and better error message

ekiwi:


src/main/scala/chiseltest/package.scala
```diff
@@ -422,8 +423,14 @@ package object chiseltest {
       inner
     }
 
-    def enumToString(record: EnumType): BigInt => String = {
-      def inner(bits: BigInt): String = "[unimplemented enum decode]"
+    def enumToString(tpe: EnumType): BigInt => String = {
+      def inner(bits: BigInt): String = {
+        val fullName = chisel3.internaltest.EnumHelpers.valueToName(tpe, bits).getOrElse("???")
+        // we only want to class and value name, not the package or enclosing class
+        val noPackage = fullName.split('.').takeRight(2).mkString(".")
```
vighneshiyer, line 430: Is the full name usually package.chiselModuleName.enumObjectName.enumValue ?
Are there cases where there may be nested (dynamically created) classes inside a chisel module (or any nested set of classes) like package.chiselModuleName.factoryClass.enumObjectName.enumValue in which case we want the enclosing class?
Since this is just a debugging feature - it's not a big deal to get this precisely right in every case

src/test/scala/chiseltest/tests/ChiselEnumTest.scala
```diff
@@ -47,26 +46,40 @@ class ChiselEnumTest extends AnyFlatSpec with ChiselScalatestTester with Matcher
...
 
-  ignore should "peek enum literals" in {
+  it should "peek enum literals" in {
     test(new Module {
       val io = IO(new Bundle {
         val in = Input(UInt(2.W))
         val out = Output(EnumExample())
       })
-      io.out := io.in.asTypeOf(chiselTypeOf(io.out))
+      val (enum, _) = EnumExample.safe(io.in)
```
vighneshiyer, line 55: enum is a reserved keyword I think? this may break in Scala 3

vighneshiyer, line 55: This is a lot nicer than the type casting used before

src/main/scala/chisel3/internaltest/EnumHelpers.scala
```diff
@@ -0,0 +1,21 @@
...
+import chisel3.experimental.EnumType
+
+/** Helper functions to allow for peeks and better debugging of ChiselEnums.
+  * This needs to be in a `chisel3` package in order to access the package private
+  * `factory` field of EnumType.
+  */
+object EnumHelpers {
+  def fromBits[T <: EnumType](tpe: T, bits: BigInt): T = {
+    val all = tpe.factory.all
+    all.find(_.litValue == bits).get.asInstanceOf[T]
```
vighneshiyer, line 14: Why is the .asInstanceOf[T] cast required here? I know it's done pretty much all over the place, but I'm still not sure why

vighneshiyer, line 14: Why is the .asInstanceOf[T] cast required here? I know it's done pretty much all over the place, but I'm still not sure why

src/main/scala/chiseltest/package.scala
```diff
@@ -422,8 +423,14 @@ package object chiseltest {
...
+        val noEnclosingClass = noPackage.split('$').last
```
vighneshiyer, line 431: Could you include an example of what fullName typically is as a comment to make the transforms on lines 430-1 clearer?

src/main/scala/chisel3/internaltest/EnumHelpers.scala
vighneshiyer, line 14: Why is the .asInstanceOf[T] cast required here? I know it's done pretty much all over the place, but I'm still not sure why

vighneshiyer (commented):
Looks great! Just a few comments. I think if I hit approve now it will automerge so I want to avoid that until these nits are cleared up

src/main/scala/chiseltest/package.scala
ekiwi: There might be exceptions. I would just wait until someone complains.

src/test/scala/chiseltest/tests/ChiselEnumTest.scala
ekiwi: I will change it.

src/main/scala/chisel3/internaltest/EnumHelpers.scala
ekiwi: no idea, feel free to debug this if you are interested.

src/main/scala/chiseltest/package.scala
ekiwi: I think that would be too noisy

vighneshiyer (approved):

PR #545, opened 2022-08-03
formal: add support for bitwuzla solver

ekiwi:
This will be ready to merge once the final bug in `bitwuzla` has been fixed:  https://github.com/bitwuzla/bitwuzla/issues/35

### DIFF
--- a/src/main/scala/chiseltest/formal/backends/smt/SMTLibResponseParser.scala
+++ b/src/main/scala/chiseltest/formal/backends/smt/SMTLibResponseParser.scala
@@ -4,14 +4,15 @@
 package chiseltest.formal.backends.smt
 
 import scala.annotation.tailrec
+import scala.collection.mutable
 
 private object SMTLibResponseParser {
   def parseValue(v: String): Option[BigInt] = {
     val expr = SExprParser.parse(v)
     expr match {
       // this is the assignment, something like ((in #b00000))
-      case SExprNode(List(SExprNode(List(_, value)))) => parseValue(value)
-      case _                                          => throw new NotImplementedError(s"Unexpected response: $expr")
+      case SExprNode(Seq(SExprNode(Seq(_, value)))) => parseValue(value)
+      case _                                        => throw new NotImplementedError(s"Unexpected response: $expr")
     }
   }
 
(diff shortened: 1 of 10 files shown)


src/main/scala/chiseltest/formal/backends/smt/SMTLibResponseParser.scala
```diff
@@ -72,33 +72,56 @@ private object SMTLibResponseParser {
...
+      case '('                                   => finish(); tokens = tokens :+ "("
+      case ')'                                   => finish(); tokens = tokens :+ ")"
+      case '|' if inEscape                       => tmp += '|'; finish(); inEscape = false
+      case '|' if !inEscape                      => finish(); inEscape = true; tmp = "|"
+      case ' ' | '\t' | '\r' | '\n' if !inEscape => finish()
+      case other                                 => tmp += other
+    }
+    finish()
+    tokens
+  }
```
jackkoenig, line 123:
```suggestion
  private def tokenize(line: String): List[String] = {
    val tokens = mutable.ListBuffer.empty[String]
    var inEscape = false
    var tmp = ""
    def finish(): Unit = {
      if (tmp.nonEmpty) {
       tokens += tmp
        tmp = ""
      }
    }
    line.foreach {
      case '('                                   => finish(); tokens += "("
      case ')'                                   => finish(); tokens += ")"
      case '|' if inEscape                       => tmp += '|'; finish(); inEscape = false
      case '|' if !inEscape                      => finish(); inEscape = true; tmp = "|"
      case ' ' | '\t' | '\r' | '\n' if !inEscape => finish()
      case other                                 => tmp += other
    }
    finish()
    tokens.toList
  }
```
Appending to a `List` is going to be pretty slow for large output Lists (n^2). If you're really committed to using a `List`, you'll get much better performance this way. I'd also consider doing a similar thing with a `VectorBuilder` for more efficient memory use.

jackkoenig (approved):

src/main/scala/chiseltest/formal/backends/smt/SMTLibResponseParser.scala
ekiwi: This is a great suggestion! I will change it.

PR #546, opened 2022-08-08
Verilator make with fixed number of jobs

pineapplehunter:
This will change varilator make to run on a fixed number of jobs. Specificly the number of cores in the computer.

From the `make` manual,
> If the -j option is given without an argument, make  will  not  limit the number of jobs that can run simultaneously.

This behavior was causing my computer to swap a lot.


ekiwi (approved):
Thanks! :shipit:

Issue #35, opened 2019-04-30
Peeking zero-width I/Os


edwardcwang:
While I know that zero-width I/Os don't really exist, it might be nice as syntactic sugar to be able to peek (or even dummy poke?) zero-width wires and have it return 0. This is especially useful in generators where parameters might be set in a way such that some wires end up with zero-width.

This might also be a [treadle](https://github.com/freechipsproject/treadle) issue... @chick ?

Not super high priority since it can be worked around with some if statements.

Currently it fails with a slightly cryptic message:

```
java.lang.AssertionError: assertion failed: Error: getValue("zero_width_io_name") : argument is not an element of this circuit
  at scala.Predef$.assert(Predef.scala:219)
  at treadle.executable.ExecutionEngine.getValue(ExecutionEngine.scala:165)
  at treadle.TreadleTester.peek(TreadleTester.scala:245)
  at chisel3.tester.TreadleBackend.peekBits(TreadleBackend.scala:90)
  at chisel3.tester.TreadleBackend.expectBits(TreadleBackend.scala:99)
  at chisel3.tester.package$testableData.expectWithStale(package.scala:72)
  at chisel3.tester.package$testableData.expect(package.scala:91)
```

chick:
Hmm. This would be easy to do, perhaps as a mode where you can poke or peek anything you want without erroring. But treadle only has symbols for things that have survived all the way to low firrtl. so it would not otherwise know that the signal is zero length, it just simply isn't there, right?

edwardcwang:
Yeah, that's the tricky part methinks... I guess this is the inverse of #2.

ducky64:
Perhaps zero-width signals shouldn't disappear in low-FIRRTL? Even in Chisel frontend we kind of have a distinction between zero-width signals and optional Bundle fields, and zero width signals could be considered to be a form of design intent?

edwardcwang:
Well, zero-width wires are also a Chisel/Chirrtl thing and don't actually exist in hardware...

chick:
Annotations

edwardcwang:
![anno](https://user-images.githubusercontent.com/10752510/57000511-ec7e9400-6b68-11e9-87af-3bd5204851ba.png)

ekiwi:
This should be fixed by: https://github.com/ucb-bar/chiseltest/commit/7d98ab58459832607361d045a90bdbccfe265297

PR #552, opened 2022-08-16
Support for Decoupled/Valid with zero-width data

kammoh:
As discussed in #35, `chiseltest` does not currently support zero-width data/ports.
This PR provides a solution (well, maybe a workaround) specific to `Decoupled` and `Valid`
interfaces. Decoupled interfaces with zero-width data can also be used
as a control/synchronization handshake where no data payload exists.
Here we take an explicit approach: the designer knows the
width of data at instantiation and can accordingly utilize appropriate
`Option[Data]` values.


kammoh:
CI failures seem to be completely unrelated. (One is probably a bug in Verilator caching?)

ekiwi:
Thanks for the PR. I think there might be a more general solution for 0-bit wires which is why I would rather not merge this yet.

ekiwi:
@kammoh Could you have a look at my changes and see if this addresses your problem? I did remove your option to pass `Option[Data]` to enqueue. Instead you can now just pass `0.U` for any 0-bit signal. Bundles with 0-bit signals now don't need those to be set, even when not using `pokePartial`.

PR #558, opened 2022-10-04
fix: S-interpolator for assert, assume and printf

SingularityKChen:
To fix the errors in https://github.com/chipsalliance/chisel3/pull/2751.

### DIFF
--- a/src/test/scala/chiseltest/backends/icarus/IcarusBlackBoxTest.scala
+++ b/src/test/scala/chiseltest/backends/icarus/IcarusBlackBoxTest.scala
@@ -55,7 +55,8 @@ class PlusArgReader extends BlackBox with HasBlackBoxInline {
 
 class PlusArgReaderWrapper(expected: Int) extends Module {
   val reader = Module(new PlusArgReader)
-  chisel3.assert(reader.io.out === expected.U, "Expected $expected, got %x.\n", reader.io.out)
+  val msg = s"Expected $expected, got %x.\n" // this works around the fact that s".." is forbidden in the assert
+  assert(reader.io.out === expected.U, msg, reader.io.out)
 }
 
 class IcarusBlackBoxTests extends AnyFlatSpec with ChiselScalatestTester {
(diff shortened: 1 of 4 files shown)


ekiwi:
Unfortunately this changes the functionality of the code since it will now print out `$expected` instead of the string representation of `expected`. @jackkoenig: what is the recommended way to fix this?

src/test/scala/chiseltest/backends/icarus/IcarusBlackBoxTest.scala
```diff
@@ -55,7 +55,8 @@ class PlusArgReader extends BlackBox with HasBlackBoxInline {
 
 class PlusArgReaderWrapper(expected: Int) extends Module {
   val reader = Module(new PlusArgReader)
-  assert(reader.io.out === expected.U, s"Expected $expected, got %x.\n", reader.io.out)
+  val msg = s"Expected $expected, got %x.\n" // this works around the fact that s".." is forbidden in the assert
+  assert(reader.io.out === expected.U, msg, reader.io.out)
```
jackkoenig, line 59:
```suggestion
  assert(reader.io.out === expected.U, cf"Expected ${expected.toString}, got ${reader.io.out}%x.\n")
```

jackkoenig:
> @jackkoenig: what is the recommended way to fix this?

You can either split out the `String` as you have done, or use `cf` but explicitly `.toString` as in the "suggestion" comment I left.

ekiwi:
> You can either split out the `String` as you have done, or use `cf` but explicitly `.toString` as in the "suggestion" comment I left.

I would prefer not to call `toString` explicitly. What exactly does `cf` do? Can you link me to the documentation for that (hard to google)

ekiwi:
> I would prefer not to call `toString` explicitly. What exactly does `cf` do? Can you link me to the documentation for that (hard to google)

Never mind, I found it: https://www.chisel-lang.org/chisel3/docs/explanations/printing.html
Somehow I thought the Chisel string interpolator was `p`.

jackkoenig:
Until fairly recently, `p` was the Chisel String interpolator, but in order to add format specifiers, we added `cf`. Since `cf` can do anything `p` can, and it's name makes more sense ("Chisel format string") rather than ("printable"), I figure we should just use `cf` instead.

SingularityKChen:
> > @jackkoenig: what is the recommended way to fix this?
> 
> You can either split out the `String` as you have done, or use `cf` but explicitly `.toString` as in the "suggestion" comment I left.

Hi Jack, thanks for your help. But there is one thing that I'm quite confused about. Why do we need to explicitly `.toString` to the Int type in `cf`? Is `cf` cannot convert Scala types?

In that case, is the following code correct?
```scala
val intA: Int = 10
val uIntB: UInt = intA.U
printf(s"intA=$intA and " + cf"uIntB=$uIntB")
print(cf"intA=${intA.toString} and uIntB=$uIntB")
```

PR #580, opened 2022-12-07
make scala 2.13 the default version

ekiwi:
We still support 2.12, but chiseltest developers
will work with 2.13 by default.


jackkoenig (approved):
I've been debating doing this for a while on the chisel3 repo. The only reason I haven't is that it's pretty easy to accidentally rely on Scala 2.13-only features and it's a bit annoying if they're not caught until CI. On the flip side, I'd rather eliminate all warnings on Scala 2.13 and not worry if any remain in 2.12.

Regardless, a reasonable move.

PR #586, opened 2022-12-15
Fix scala 2.13 warnings and update sbt

ekiwi:


ekiwi:
No backport since this changes the function signature for `join`

src/main/scala/chiseltest/internal/BackendInterface.scala
```diff
@@ -11,13 +11,13 @@ import scala.collection.mutable
 trait AbstractTesterThread
 
 class TesterThreadList(protected val elts: Seq[AbstractTesterThread]) {
-  def toSeq(): Seq[AbstractTesterThread] = elts
+  def toSeq: Seq[AbstractTesterThread] = elts
 
-  def join() {
+  def join: Unit = {
     Context().backend.doJoin(elts, None)
   }
```
jackkoenig, line 18: I actually think `join` should keep the `()`. The idea is that if a 0-arity function has side-effects, it should have the `()`. `Iterator.hasNext` vs. `Iterator.next()` is a good example: https://www.scala-lang.org/api/2.13.8/scala/collection/Iterator.html

ekiwi: I think you are right. I will change it.

ekiwi: changed

jackkoenig (approved):

PR #589, opened 2022-12-16
partialPoke/Expect: add more tests, improve implementation

ekiwi:
This makes partial pokes / expects slightly less broken by implementing them in terms of the normal poke / expect function. One use case that has new tests added and that was definitely somewhat broken is bundles inside of vecs.

This also removes four warnings, which - in this case and afaik - were indicative of a real bug.


jackkoenig (approved):
Neat!

PR #591, opened 2022-12-22
Delay first rising clock edge by 1 half-period in VPI Harness

jackkoenig:
This prevents a race condition between the rising clock edge and randomization when RANDOMIZE_DELAY=1. Note that the traditional Chisel randomization delay of #0.002 is based on VCS's default timescale of 1ns/1ps and does not properly work with other timescales. It is much safer to use a full timescale delay, and thus prudent to delay the first rising clock edge till the 2nd time step.

Note that this does change waveforms slightly such that you have a full clock period with no clock before it starts rather than the first rising edge being after merely half a clock-period. See `old` and `new` (except of course there is no zero-time falling edge, I can't figure out how to get rid of it in Wavedrom)

Screen Shot 2022-12-22 at 2 37 45 PM

### DIFF
--- a/src/main/scala/chiseltest/simulator/ipc/VpiVerilogHarnessGenerator.scala
+++ b/src/main/scala/chiseltest/simulator/ipc/VpiVerilogHarnessGenerator.scala
@@ -32,7 +32,12 @@ private[chiseltest] object VpiVerilogHarnessGenerator {
       codeBuffer.append(s"  wire[${width - 1}:0] $name;\n")
     }
 
-    codeBuffer.append(s"  always #`CLOCK_PERIOD $clockName = ~$clockName;\n")
+    codeBuffer.append(s"  initial begin\n")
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

ekiwi (approved):

ekiwi (changes requested):
I forgot this yesterday, but this PR is missing a test that fails without your fix. Could you please add Icarus Verilog based test that fails without the delay? Otherwise we might at some point wonder why the delay is there, remove it, see that no tests are failing and thus believe that it isn't actually necessary.

jackkoenig:
> Could you please add an Icarus Verilog based test that fails without the delay? 

Yep, good point and suggestion. I have not quite gotten a test working yet, but pushed a WIP branch that I'll look at again next week or in the new year.

aswaterman:
Isn't the failing test going to be nondeterministic (at least with respect to Verilog codegen changes), since the resolution of the race depends on the event queue ordering?

ekiwi:
Let's just get it in, even without a test.

ekiwi (approved):

PR #593, opened 2023-01-06
do not warn about firrtl imports

ekiwi:
We know what we are doing :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -57,7 +57,8 @@ scalacOptions ++= Seq(
   "-language:reflectiveCalls",
   "-deprecation",
   "-feature",
-  "-Xcheckinit"
+  "-Xcheckinit",
+  "-Wconf:msg=firrtl:s", // do not warn about firrtl imports
 ) ++ {
   CrossVersion.partialVersion(scalaVersion.value) match {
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -57,7 +57,8 @@ scalacOptions ++= Seq(
   "-language:reflectiveCalls",
   "-deprecation",
   "-feature",
-  "-Xcheckinit"
+  "-Xcheckinit",
+  "-Wconf:msg=firrtl:s", // do not warn about firrtl imports
```
jackkoenig, line 61:
```suggestion
  "-Wconf:msg=Importing from firrtl:s", // do not warn about firrtl imports
```
I now realize this is what we're using in Chisel itself, but IMO this message is a little too broad. I would suggest a slightly more specific message, but it's up to you!

jackkoenig (approved):
I suggest a more specific message, but I think it's reasonable either way.

build.sbt
ekiwi: I would prefer to stick to Adam's example.

ekiwi: I think you were actually right ... sorry. I am going to add this to the other warnings PR. Now that I understand better how this works, seems like it would be better to be more specific.

PR #594, opened 2023-01-06
remove support for deprecated Interval data type

ekiwi:
Chisel 3.6 deprecated Interval support.


jackkoenig:
Intervals are deprecated in 3.6, but not removed, I think it would be better for chiseltest to support them for the 3.6.x series. You can suppress the deprecation warnings with `-Wconf` stuff

ekiwi:
> You can suppress the deprecation warnings with `-Wconf` stuff

How exactly would I suppress the Interval warnings while still seeing the Scala 2.13 deprecated warnings?

jackkoenig:
> > You can suppress the deprecation warnings with `-Wconf` stuff
> 
> How exactly would I suppress the Interval warnings while still seeing the Scala 2.13 deprecated warnings?

Here's one way of doing it: https://github.com/ucb-bar/chiseltest/pull/597

ekiwi:
Thanks for showing me how to ignore the warnings @jackkoenig !

PR #596, opened 2023-01-07
Add Verilator 5.004 to CI

jackkoenig:


ekiwi:
> Verilator 5 added support for SystemVerilog timing constructs, so my understanding is that it's a massive change to the underlying threading implementation.

I thought that by default the new timing feature would be off, but could not find any confirmation on that. You could try passing ` --no-timing`.

The number of threads should also be 1 be default: https://verilator.org/guide/latest/exe_verilator.html#cmdoption-threads

PR #604, opened 2023-01-10
Fix vpd generation

terpstra:
Presumably this was never tested, because it failed trivially.


ekiwi (approved):
Thanks! You are right, there is no VCS in CI, so things are pretty untested.

terpstra:
(I targeted 0.5.x as that's what we are using, but I note that the bug is also present on master)

ekiwi:
> (I targeted 0.5.x as that's what we are using, but I note that the bug is also present on master)

Oh, that is why it isn't merging. Could you please re-open this PR against `main`? Our `mergify` bot will then backport the fix to the `0.5.x` branch. This is essentially the same flow as Chisel uses.

PR #607, opened 2023-01-11
Remove Icarus debug messages

pineapplehunter:
When using Icarus backend there was many debug messages that was not needed. I changed it to not output anything normally but show debug messages when `SimulatorDebugAnnotation` is set, like the Verilator backend.

By the way I found that some tests in `PeekPokeCompliance` had not properly finished the dut. it was causing the simulator to keep running after the test finished.

### DIFF
--- a/src/test/scala/chiseltest/simulator/PeekPokeCompliance.scala
+++ b/src/test/scala/chiseltest/simulator/PeekPokeCompliance.scala
@@ -91,6 +91,7 @@ abstract class PeekPokeCompliance(sim: Simulator, tag: Tag = DefaultTag) extends
       if(res != values) { dut.finish() }
       assert(res == values, s"width = $w")
     }
+    dut.finish()
   }
 
   it should "propagate SInt values of variable widths through the circuit" taggedAs(tag) in {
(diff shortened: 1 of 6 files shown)


pineapplehunter:
It seems like the CI is having some network issues.
I think it would pass if run again.

ekiwi:
> It seems like the CI is having some network issues.
> I think it would pass if run again.

What happened is that we now disallow warnings and Chisel just added a new deprecation warning. I will make a PR to fix that shortly, then you can rebase on `main` and it should work.

src/test/scala/chiseltest/simulator/PeekPokeCompliance.scala
```diff
@@ -91,6 +91,7 @@ abstract class PeekPokeCompliance(sim: Simulator, tag: Tag = DefaultTag) extends
       if(res != values) { dut.finish() }
       assert(res == values, s"width = $w")
     }
+    dut.finish()
```
ekiwi, line 94: This is a great catch!

ekiwi (approved):
LGTM. I will make a PR to deal with the deprecation warning soon.

ekiwi:
OK, so we are waiting for: https://github.com/ucb-bar/chiseltest/pull/608 to be merged, then you can rebase.

ekiwi:
@pineapplehunter Once you rebase on `main`, we can get this in.

pineapplehunter:
Rebased!

ekiwi:
Awesome. Thank you!

PR #609, opened 2023-01-11
[test] 0-bit and UInt<?> support for different simulators

ekiwi:


src/test/scala/chiseltest/backends/verilator/VerilatorZeroWidthIntsTest.scala
```diff
@@ -0,0 +1,30 @@
...
+import chiseltest.tests.{PassthroughModule, ZeroWidthTestBundle, ZeroWidthTestBundleSigned}
+import org.scalatest.freespec.AnyFreeSpec
+
+class VerilatorZeroWidthIntsTest extends AnyFreeSpec with ChiselScalatestTester {
+  private val annos = Seq(VerilatorBackendAnnotation)
+
+  "peek, poke and expect zero width UInt" taggedAs RequiresVerilator  in {
+    test(new PassthroughModule(UInt(0.W))).withAnnotations(annos) { c =>
+      c.in.poke(0.U)
+      c.in.expect(0.U)
```
jackkoenig, line 16: These are not zero width literals, `0.U(0.W)`, is zero-width. Attempting to change the width behavior of `0.U` would be sketchy so we've punted on it for now.

ekiwi: I know. I am relying on the fact that chiseltest does not care about the width as long as the value fits into the I/O port. When I originally created these tests `0.U(0.W)` was not supported by upstream Chisel.

How about we get this PR in and I will promise to make a followup PR to fix this as there are some tests from a prior PR (https://github.com/ucb-bar/chiseltest/pull/552/files) which have the same problem?

jackkoenig (approved):

PR #611, opened 2023-01-11
chiseltest: eliminate dump file annotations

terpstra:
I'm not 100% sure this is the correct fix on main.
In our private branch, WriteVpdAnnotation was missing which caused circt to fail complaining about the annotation.
In main, I don't see ANY of the annotations being filtered, though.

So maybe this is no longer necessary?


ekiwi:
Sorry, but I do not understand what bug this is supposed to address. What is the problem you are trying to solve here?

terpstra:
When you run chiseltest with generation of a vpd file, it ends up leaving the WriteVpdAnnotation on the file sent to firrtl. The circt firrtl compiler errors out if there are any annotations which it does not understand. It does not understand WriteVpdAnnotation.

terpstra:
@GeorgeLyon wrote the actual fix. This was my (apparently failed) attempt to upstream it.

ekiwi:
OK. I will close this for now. We probably want to filter out more annotations, but if we want to fix that we should do it right and make a base class for all user annotations in `chiseltest`, something like `ChiselTestOption` and then just filter everything out that extends that.

PR #625, opened 2023-03-04
FIX: remove treadleModule in build.sc

SingularityKChen:
As treadle source code is imported in https://github.com/ucb-bar/chiseltest/commit/a6b1091f66bf58f37855093b314d3d73dae12f23, `treadleModule` should be removed in `build.sc`.


ekiwi (approved):
Awesome. Thanks!

PR #631, opened 2023-05-04
Chiseltest JNI Backend

oliveryu11:
TODOs:
- Add a switch to go between JNI and JNA, via some annotation
- test on other computers and validate via CI
- Remove duplication between JNIUtils and JNAUtils
- Do a final performance benchmark with the GCD test


build.sbt
```diff
@@ -84,3 +84,5 @@ libraryDependencies ++= Seq(
       )
   }
 }
+
+fork := true
```
ekiwi, line 88: Before we merge this PR, please make sure that it works without this change.

ekiwi: That is important, because otherwise it will make `chiseltest` harder to use.

src/main/scala/chiseltest/simulator/jni/JniAPI.scala
```diff
@@ -0,0 +1,76 @@
...
+  } else {
+    System.getProperty("os.name") match {
+      case "Mac OS X" => "darwin"
+      case "Linux"    => "linux"
+      case s: String => s
+    }
+  }
+
+  val tmpDir = os.Path(System.getProperty("java.io.tmpdir"))
+  val jniSoDir = tmpDir / "chiseltest_jni_bridge"
```
ekiwi, line 25: Ideally you would compile everything in the local test directory instead of a temporary directory. That will make it easier for debugging.

ekiwi:
This looks very promising. I left some initial comments. Please ping me when this is ready for a more thorough review.

PR #639, opened 2023-06-09
remove mill build file

ekiwi:
It is no longer up to date and it looks like @sequencer removed chiseltest from https://github.com/chipsalliance/playground anyways.


edwardcwang:
Is it hard to bring it up to date? I actually use mill and this would be a bit sad :(

ekiwi:
> Is it hard to bring it up to date? I actually use mill and this would be a bit sad :(

It is always a pain to keep those two files up to date. Previously @sequencer would at least use the mill script and keep it up to date. Since that seems to be gone, there is no real justification to do the work, unless you @edwardcwang  want to test and update the mill script.

sequencer (approved):
LGTM

ekiwi:
> LGTM, Chisel is not depending on SFC any more, so not possible to support chiseltest :(

Chiseltest does actually now work with Chisel 5 and I am going to try to make it work with 6. Just because the default compiler is now CIRCT, doesn't mean that you cannot use an alternative firrtl compiler to make treadle work.

Issue #661, opened 2023-08-01
chiseltest gets the signal name wrong when trying to peek, poke, or expect an OpaqueType


konda-x1:
When a `Record` extends `OpaqueType`, it can only have one element with an empty string as its name (provided that its `opaqueType` method returns `true`). In that case, during elaboration, chisel doesn't append any underscores to a signal with the `Record` type. However, chiseltest still tries to append an underscore, thus getting the wrong signal name.

Example:
```scala
import chisel3.experimental.BundleLiterals.AddBundleLiteralConstructor
import chisel3.experimental.OpaqueType
import chisel3.{fromDoubleToLiteral => _, fromIntToBinaryPoint => _, _}
import chiseltest._
import org.scalatest.flatspec.AnyFlatSpec

import scala.collection.immutable.SeqMap

class OpaqueTypeSpec extends AnyFlatSpec with ChiselScalatestTester {

  class OpaqueRecord[T <: Data](val data: T) extends Record with OpaqueType {
    val elements: SeqMap[String, T] = SeqMap("" -> data)
  }

  //noinspection TypeAnnotation
  class OpaquePassthrough[T <: Data](data: T) extends Module {
    val in = IO(Input(new OpaqueRecord(data)))
    val out = IO(Output(new OpaqueRecord(data)))
    out := in
  }

  behavior of "OpaqueType"

  it should "poke and expect successfully" in {
    test(new OpaquePassthrough(SInt(8.W))) { dut =>
      def sIntVal = -4.S(8.W)
      def sIntRec = new OpaqueRecord(SInt(8.W)).Lit(_.data -> sIntVal)
      dut.in.poke(sIntRec)
      dut.out.expect(sIntRec)
    }
  }

}

```

yields the following error message:

```
[info] OpaqueTypeSpec:
[info] OpaqueType
[info] - should poke and expect successfully *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: Error: getValue("in_") : argument is not an element of this circuit
[info]   at scala.Predef$.assert(Predef.scala:279)
[info]   at treadle.executable.ExecutionEngine.getValue(ExecutionEngine.scala:238)
[info]   at treadle.TreadleTester.peek(TreadleTester.scala:303)
[info]   at chiseltest.simulator.TreadleContext.peek(TreadleSimulator.scala:85)
[info]   at chiseltest.internal.GenericBackend.pokeBits(GenericBackend.scala:70)
[...]

PR #662, opened 2023-08-01
Replace getDataNames with DataMirror.fullModulePorts

konda-x1:
Addresses #661.

* Remove `getDataNames`
* Add test with `Record` extending `OpaqueType`


ekiwi:
Looks like the automatic backport did not work. If you are feeling adventurous, then you could open another PR against the `0.6.x` branch for the backport. Unfortunately I won't be able to do a release for another ~2 weeks though since I need to finish writing my ASPLOS paper submission.

PR #702, opened 2024-01-14
Add pono backend to chisel formal for bmc

Gallagator:
The code is fairly straightforward as much of the existing btor code can be reused. The only real difference between the two btor2 backends is their command line invocation.

----

From here, it should be trivial to add k-induction as one will only need to change the engine from `bmc` to `ind` when running `pono`.

---

# Testing

I've run the [Keepmax](https://github.com/ucb-bar/chiseltest/blob/main/src/test/scala/chiseltest/formal/examples/KeepMax.scala) tests on the `pono` engine as well as `btormc` and they both succeed. I have also verified that upon editing `KeepMax` to fail, a counterexample is generated by both engines.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,7 +128,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        backend: [z3, cvc4, btormc, bitwuzla]
+        backend: [z3, cvc4, btormc, bitwuzla, pono]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/src/main/scala/chiseltest/formal/backends/Maltese.scala
+++ b/src/main/scala/chiseltest/formal/backends/Maltese.scala
@@ -2,7 +2,7 @@
 
 package chiseltest.formal.backends
 
-import chiseltest.formal.backends.btor.BtormcModelChecker
+import chiseltest.formal.backends.btor._
 import chiseltest.formal.backends.smt._
 import chiseltest.formal.{DoNotModelUndef, DoNotOptimizeFormal, FailedBoundedCheckException}
 import firrtl2._
@@ -55,6 +55,8 @@ private case object BoolectorEngineAnnotation extends FormalEngineAnnotation
   *   bitwuzla often performs better than Z3 or CVC4.
   */
 case object BitwuzlaEngineAnnotation extends FormalEngineAnnotation
+
+case object PonoEngineAnnotation extends FormalEngineAnnotation
 
 /** Formal Verification based on the firrtl compiler's SMT backend and the maltese SMT libraries solver bindings. */
 private[chiseltest] object Maltese {
@@ -166,10 +168,11 @@ private[chiseltest] object Maltese {
     engines.map {
       case CVC4EngineAnnotation      => new SMTModelChecker(CVC4SMTLib)
       case Z3EngineAnnotation        => new SMTModelChecker(Z3SMTLib)
-      case BtormcEngineAnnotation    => new BtormcModelChecker(targetDir)
+      case BtormcEngineAnnotation    => new Btor2ModelChecker(BtormcBtor2, targetDir)
       case Yices2EngineAnnotation    => new SMTModelChecker(Yices2SMTLib)
       case BoolectorEngineAnnotation => new SMTModelChecker(BoolectorSMTLib)
       case BitwuzlaEngineAnnotation  => new SMTModelChecker(BitwuzlaSMTLib)
+      case PonoEngineAnnotation      => new Btor2ModelChecker(PonoBtor2, targetDir)
     }
   }
 
--- a/src/main/scala/chiseltest/formal/backends/btor/Btor2ModelChecker.scala
+++ b/src/main/scala/chiseltest/formal/backends/btor/Btor2ModelChecker.scala
@@ -6,27 +6,33 @@ import chiseltest.formal.backends._
 import firrtl2.backends.experimental.smt._
 
 private[chiseltest] trait Solver {
+  def name(): String
   def cmd(btorFile: String, kMax: Int): Seq[String]
+  def valid_return_codes(): Set[Int]
 }
 
 object BtormcBtor2 extends Solver {
+  override def name(): String = "btormc"
   override def cmd(btorFile: String, kMax: Int): Seq[String] = {
     val kmaxOpt = if (kMax > 0) Seq("--kmax", kMax.toString) else Seq()
     Seq("btormc") ++ kmaxOpt ++ Seq(btorFile)
   }
+  def valid_return_codes(): Set[Int] = Set(0)
 }
 
 object PonoBtor2 extends Solver {
+  override def name(): String = "pono"
   override def cmd(btorFile: String, kMax: Int): Seq[String] = {
     val kmaxOpt = if (kMax > 0) Seq("-k", kMax.toString) else Seq()
-    Seq("pono") ++ Seq("-e", "bmc") ++ kmaxOpt ++ Seq("--vcd", "ignore.vcd") ++ Seq(btorFile)
+    Seq("pono") ++ Seq("-e", "bmc") ++ kmaxOpt ++ Seq("--witness") ++ Seq(btorFile)
   }
+  def valid_return_codes(): Set[Int] = Set(0, 255)
 }
 
 class Btor2ModelChecker(solver: Solver, targetDir: os.Path) extends IsModelChecker {
   override val fileExtension = ".btor2"
-  override val name:   String = "btormc"
-  override val prefix: String = "btormc"
+  override val name:   String = solver.name()
+  override val prefix: String = solver.name()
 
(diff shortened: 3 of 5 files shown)


ekiwi (changes requested):
This look really good!

The last remaining step is to add the `pono` backend to the CI tests. There are two files you need to modify:
1. https://github.com/ucb-bar/chiseltest/blob/b2d8908f6c8e2cc46a42ec61f9aa031805a4267c/src/test/scala/chiseltest/formal/FormalBackendOption.scala#L23
2. https://github.com/ucb-bar/chiseltest/blob/b2d8908f6c8e2cc46a42ec61f9aa031805a4267c/.github/workflows/test.yml#L131

In both places, please add `pono` to the list of backends.

ekiwi:
Looks like now a couple of tests are failing. You can run the tests locally by launching `sbt` and then running the following command in the `sbt` shell:
```
testOnly chiseltest.** -- -n Formal -Dformal_engine=pono
```

Gallagator:
I've noticed the tests are failing - pono seems to be acting very weirdly... I've rerun one of the tests on the .btor file for `KeepMax` and it's returning 255 because the output is unknown. Here is the output when runing:

``` sh
$ pono -e bmc -k 3 -v10 --vcd proof.vcd  KeepMax.btor
Parsing BTOR2 file: KeepMax.btor
Solving property: (bvnot (bvand (ite (bvult max out_1) #b1 #b0) (bvand (bvnot reset) _cycles)))
INIT:
(bvand reset (bvnot _resetCount))
TRANS:
(bvand (bvand (bvand (bvand (bvand (bvnot (bvand (bvnot reset) (bvnot _resetCount))) (bvnot (bvand (bvnot _resetCount.next) (bvnot reset.next)))) (ite (= _cycles.next (bvand (bvnot reset) (bvnot (bvand (bvnot _cycles) (bvnot ((_ extract 0 0) (bvadd (concat #b0 _cycles) #b01))))))) #b1 #b0)) (ite (= max.next (ite (= #b1 reset) #b00000000 (ite (bvult max in) in max))) #b1 #b0)) (ite (= max out_1.next) #b1 #b0)) (ite (= _resetCount.next (bvnot (bvand (bvnot _resetCount) (bvnot ((_ extract 0 0) (bvadd #b01 (concat #b0 _resetCount))))))) #b1 #b0))
BMC adding init constraint for step 0
BMC bound_start_: 0
BMC bound_step_: 1

BMC checking at bound: 0
  BMC adding bad state constraint for j = 0
  BMC check at bound 0 unsatisfiable

BMC checking at bound: 1
  BMC reached_k = 0, i = 1
  BMC adding transition for j-1 = 0
  BMC adding bad state constraint for j = 1
  BMC check at bound 1 unsatisfiable

BMC checking at bound: 2
  BMC reached_k = 1, i = 2
  BMC adding transition for j-1 = 1
  BMC adding bad state constraint for j = 2
  BMC check at bound 2 unsatisfiable

BMC checking at bound: 3
  BMC reached_k = 2, i = 3
  BMC adding transition for j-1 = 2
  BMC adding bad state constraint for j = 3
  BMC check at bound 3 unsatisfiable
unknown
b0

$echo $?
255
```

It's found that the the property is unsatisfiable for each step but somehow returns unknown. This seems like a bug in pono (though I'm a bit wary of saying that). When running with the induction engine `-e ind` pono returns `unsat` as we expect.

ekiwi:
> It's found that the the property is unsatisfiable for each step but somehow returns unknown.

It might be worth investigating why it returns that. It is possible that they return unknown because checking that the property holds for `k` checks does not proof that it always holds. It could also be that something is going wrong with `pono`.

ekiwi:
Another thing to be aware of: Last time I checked, which is now about 3 years ago, `pono` did not support checking more than one property at once. I thus had to combine all properties into one before dispatching to `cosa2` (which is what `pono` used to be called): https://github.com/ucb-bar/maltese-smt/blob/497f7ba057609bf637de6e4fa2de2a80826c0a0f/src/maltese/mc/Btor2ModelChecker.scala#L68

At least that was the idea. Looking at the code it seems like I never actually implemented that part of the system: https://github.com/ucb-bar/maltese-smt/blob/497f7ba057609bf637de6e4fa2de2a80826c0a0f/src/maltese/mc/TransitionSystem.scala#L88

Gallagator:
Your latter suggestion seems to be the case - it returns UNKNOWN for bmc: https://github.com/stanford-centaur/pono/blob/master/engines/bmc.cpp#L53-L77

I verified that this is the offending function by adding a print before the return.

I think I'll need to add a function for the `Solver` trait which will process the return code and turn the unknown to a pass for bmc. Might not get this done till tommorow, it's getting late over here.

ekiwi:
> I think I'll need to add a function for the `Solver` trait which will process the return code and turn the unknown to a pass for bmc. Might not get this done till tommorow, it's getting late over here.

No worries. Take your time please and let me know if there are any other things I can help with. Thanks!

ekiwi:
> I think I'll need to add a function for the `Solver` trait which will process the return code and turn the unknown to a pass for bmc.

Another thought: If the difference between `btormc` and `pono` gets too big, it might also make sense to just duplicate the code in [Btor2ModelChecker](https://github.com/ucb-bar/chiseltest/blob/74365bbfa421afcac2734506d2d97d571b284b53/src/main/scala/chiseltest/formal/backends/btor/Btor2ModelChecker.scala#L26), i.e., split it into a `BtormcModelChecker` and a `PonoModelChecker` class and get rid of the whole `solver: Solver` parameter.

Gallagator:
I've managed to fix most of the tests by allowing pono to return 255 but some tests are failing assertions because they don't fail at the right step: https://github.com/ucb-bar/chiseltest/blob/main/src/test/scala/chiseltest/formal/UndefinedValuesTests.scala#L21.
I'm not sure much can be done to make pono fail specific steps. Is it not enough for the engine to simply fail?

ekiwi:
> I'm not sure much can be done to make pono fail specific steps. Is it not enough for the engine to simply fail?

BMC should always fail at the exact same step, no matter which engine we are using. If pono fails at a different step than btormc, then that could mean that there is a bug in pono.

It might be a good idea to get the `.btor` file for one of the tests where pono fails in the wrong cycle (try to find the simplest one). Then run that file manually through `btormc` and `pono` to make sure that you get results of different lengths. If that is the case then you could file an issue with pono, presenting your findings.

Gallagator:
Thanks for the explanation! I've made an issue on pono here: https://github.com/stanford-centaur/pono/issues/320 - unfortunately, it does seem to be producing witnesses of incorrect size (usually one - offs) but it does seem to get to the bad state in the correct step (I think).

Looking at the repo, there has been no activity for at least 2 years so I'm doubtful this will be fixed. What are the next steps? I could take a look into pono to see why it isn't producing a witness but I don't fancy my chances.

ekiwi:
> What are the next steps? I could take a look into pono to see why it isn't producing a witness with an additional step but I don't fancy my chances.

Unfortunately that might be the next step. You will definitely learn a lot and I assume that the needed change is going to be fairly small. Maybe if you make a PR with that change we can get someone at Stanford to merge it.

Gallagator:
I've pushed my current changes though not all the tests pass as I've mentioned. I've looked into `pono` but I haven't yet been able to figure out the cause of the issues. Further, there is a problem I didn't notice before where pono is unable to give an undefined value when the access to memory is out of bounds. Just leaving this comment here in case I make no more progress...

ekiwi:
> I've pushed my current changes though not all the tests pass as I've mentioned.

Awesome. Thanks!

> Further, there is a problem I didn't notice before where pono is unable to give an undefined value when the access to memory is out of bounds.

Could you elaborate on that? Which test exactly fails and in what manner?

Gallagator:
Apologies for the slow response and also being a little vague.

> Could you elaborate on that? Which test exactly fails and in what manner?

This is the test I was refering to: https://github.com/ucb-bar/chiseltest/actions/runs/7602647013/job/20703458234?pr=702#step:5:570.
Pono produces an `unsat` output. I believe this is the test file: https://github.com/ucb-bar/chiseltest/blob/b2d8908f6c8e2cc46a42ec61f9aa031805a4267c/src/test/scala/chiseltest/formal/MemoryTests.scala#L200

There's also a problem when using shiftregs instead of `past` where the engine should also fail:
https://github.com/ucb-bar/chiseltest/actions/runs/7602647013/job/20703458234?pr=702#step:5:510

Gallagator:
Yes, unfortunately this seemingly easy task has blown up right in my face. I did indeed only want k induction! I think I'll try to add k-induction support to btormc as that seems to be the easiest way forward. After which I may give custom k-induction a go but it's probably best to go one step at a time.

Thanks for all your guidance and support thus far! I'll see what I can do for chiseltest in the coming week(s).

ekiwi:
> Thanks for all your guidance and support thus far! I'll see what I can do for chiseltest in the coming week(s).

Sounds great! Please feel free to ask if there is anything blocking you or you get stuck. Feel free to email me as well.

Gallagator:
Closing as we've reached a bit of a dead end - may be worth visiting in future if pono sees any future activity/

PR #705, opened 2024-01-31
Bump to chisel 6 release

kammoh:
Hi,
I'm trying to help with expediating a chiseltest 6.0.0 release as 6.0-SNAPSHOT does not work with the chisel-6.0.0 release.
I also noticed the Verilator coverage support was broken with Verilator versions >= 5.012 (and probably >=5).
Tested only on macOS. All tests (excluding VCS backend, but *including* Verilator, Icarus, and formal) pass. Tests were run as:
`sbt "testOnly chiseltest.** -- -l RequiresVcs"`

Please let me know if any changes are required or otherwise free to make any modifications.

Summary of changes:
- Bumped chisel -> 6.0.0
 - Fixed missing (moved) 'Width' class
- Bumped scala -> 2.13.12
 - Fixed compilation issues due to 'scala.io' namespace clash
- bump sbt -> 1.9.8
- fix Verilator 5.x coverage
- fix Verilator 5.x tests

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.9.8
+sbt.version=1.9.7
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.8.2
+sbt.version=1.9.8
--- a/src/main/scala/chiseltest/simulator/ChiselBridge.scala
+++ b/src/main/scala/chiseltest/simulator/ChiselBridge.scala
@@ -154,8 +154,12 @@ private object ChiselBridge {
     // ignoreDecodeTableAnnotation since it is not needed by the firrtl compiler
     case _: DecodeTableAnnotation => None
     //
-    case _ => Some(UnsupportedAnnotation(anno.getClass.getSimpleName, anno.toString))
-    // case _ => throw new NotImplementedError(s"TODO: convert ${anno}")
+    case _ =>
+      println(
+        s"[WARNING] Unsupported annotation: ${anno.getClass.getSimpleName}\n" +
+          s" Please report this issue at https://github.com/ucb-bar/chiseltest/issues"
+      )
+      Some(UnsupportedAnnotation(anno.getClass.getSimpleName, anno.toString))
   }
 
   private def convert(c: Circuit): firrtl2.ir.Circuit =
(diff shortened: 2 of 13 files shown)


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.8.2
```
ekiwi, line 1: Could you set this to `1.9.7` please in order to work around this problem: https://github.com/sbt/sbt/issues/7463

src/main/scala/chiseltest/simulator/ChiselBridge.scala
```diff
@@ -154,7 +154,8 @@ private object ChiselBridge {
     // ignoreDecodeTableAnnotation since it is not needed by the firrtl compiler
     case _: DecodeTableAnnotation => None
     //
-    case _ => throw new NotImplementedError(s"TODO: convert ${anno}")
+    case _ => Some(UnsupportedAnnotation(anno.getClass.getSimpleName, anno.toString))
```
ekiwi, line 157: I understand that it is nice if we can just continue. However, the fact that we throw this `NotImplementedError` has gotten me a good number of issues and PRs which allowed me to check whether we can ignore the annotation or not.

Maybe a good compromise would be to generate the `UnsupportedAnnotation` as you do here, but then also add a `println` that tells users about the unsupported annotation and urges them to file an issue at https://github.com/ucb-bar/chiseltest/issues.

ekiwi (changes requested):
Two small things that need to be addressed. Otherwise, this is awesome! Thanks a lot!

ekiwi (approved):
This looks great! Will merge once all the tests pass.

I will try to make a `6.0.0` release of chiseltest soon.

ekiwi:
The failures in the `formal` tests are because of a new `ConstType`: https://github.com/chipsalliance/chisel/commit/b55bfd0be11044fb8d713573aa25995fe9e6910a

I think we could probably just drop that info for now.

The `sbt test` on mac that is failing might be sporadic. I will restart it soon

kammoh:
The link to the issue on the last commit comment should be -> https://github.com/verilator/verilator/pull/2505

ekiwi:
> It's ok with me if you'd rather hold back for the 6.0.0 release to include that fix, but in that case would it be possible to have a `6.0-SNAPSHOT` release with these changes sooner?

Yes. Once you fix the failure of the `formal` testsuite by dealing with the `ConstType` I will try to find a workaround.

kammoh:
Thanks!

kammoh:
I couldn't find much documentation on `ConstType` but it seems it should be safe to just take and convert the inner type.

ekiwi:
> I couldn't find much documentation on `ConstType` but it seems it should be safe to just take and convert the inner type

Thanks! I think so, too.

ekiwi:
Could you move the Verilator 5 CI tests to a separate PR. Seems like they are timing out right now and there is no reason they should keep this PR from getting merged. Thanks!

kammoh:
Typo in comment (should be "remove") but I'm leaving it for now. It would be great if you could fix it on merge.

ekiwi:
Thanks!

PR #711, opened 2024-02-09
[CI] Increase timeout for verilator tests

kammoh:
Increase the timeout to prevent fails. At some point though, the number of supported/tested verilator versions should probably be reduced.


ekiwi:
I think something else must be going wrong here. After building verilator, the actual tests should only take around 5min.

